Issue #80, opened 2015-02-18
Contents recreate_toc() and switch_image3 can have race condition


pllim:
I noticed this when I had 20 multi-extension FITS loaded at the same time. If `Contents` attempts to perform image switching operation while table of contents is being populated, I get this:
```
  File ".../lib/python2.7/site-packages/ginga-2.2.20150212221817-py2.7.egg/ginga/qtw/plugins/Contents.py", line 86, in switch_image3
    self.switch_image2(items[0], 0)
IndexError: list index out of range
```

This is most likely the cause:
```
    def recreate_toc(self):
        self.logger.debug("Recreating table of contents...")
        toclist = list(self.nameDict.keys())
        toclist.sort()

        self.treeview.clear()

# Somewhere here, switch_image3() was invoked by a callback

        for key in toclist:
            ...
```

I am thinking to fix it like this:
```
    def switch_image3(self):
        items = list(self.treeview.selectedItems())
        if len(items) == 0:
            return
        self.switch_image2(items[0], 0)
```

ejeschke:
@pllim, I think we'd better fix it with a threading reentrant lock.  I'll take a look at it.

ejeschke:
Could you try commit c4410d72c1244b4507374574ea78da8dc3671ceb ?  This has a lock added to protect against this race condition.

pllim:
Hmm. I am getting a Python traceback when I switch image via `Thumbs`
```
Traceback (most recent call last):
  File ".../lib/python2.7/site-packages/ginga-2.2.20150212221817-py2.7.egg/ginga/qtw/plugins/Thumbs.py", line 36, in mousePressEvent
    self.thumbs_cb()
  File ".../lib/python2.7/site-packages/ginga-2.2.20150212221817-py2.7.egg/ginga/qtw/plugins/Thumbs.py", line 101, in <lambda>
    image_future)
  File ".../lib/python2.7/site-packages/ginga-2.2.20150212221817-py2.7.egg/ginga/misc/plugins/ThumbsBase.py", line 170, in load_file
    image_future=bnch.image_future)
TypeError: add_preload() got an unexpected keyword argument 'image_future'
Traceback (most recent call last):
  File ".../lib/python2.7/site-packages/ginga-2.2.20150212221817-py2.7.egg/ginga/qtw/plugins/Thumbs.py", line 36, in mousePressEvent
    self.thumbs_cb()
[...]

PR #83, opened 2015-02-24
Fix TypeError when Control calls add_preload() with image_future

pllim:
Fix `TypeError` when `Control.py` calls `add_preload()` with `image_future` keyword, as mentioned in #80.

The real fix is in lines 630-633. The rest is just my Emacs automatically removing trailing whitespace.

### DIFF
--- a/ginga/Control.py
+++ b/ginga/Control.py
@@ -44,7 +44,7 @@ class ControlError(Exception):
     pass
 
 class GingaControl(Callback.Callbacks):
-     
+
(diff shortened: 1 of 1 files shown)


ejeschke:
Thanks @pllim, I saw this after I already fixed from reading an earlier email.  Issue should be resolved in 066ebe3dd723a8a9c8b9ba3b78b029d375718555.

pllim:
Okay, no problem.

PR #84, opened 2015-02-25
Fix race condition in Contents (Qt)

pllim:
Fix ejeschke/ginga#80

### DIFF
--- a/ginga/qtw/plugins/Contents.py
+++ b/ginga/qtw/plugins/Contents.py
@@ -1,6 +1,6 @@
 #
 # Contents.py -- Table of Contents plugin for fits viewer
-# 
+#
 # Eric Jeschke (eric@naoj.org)
 #
 # Copyright (c)  Eric R. Jeschke.  All rights reserved.
@@ -51,7 +51,7 @@ class Contents(GingaPlugin.GlobalPlugin):
         #treeview.itemDoubleClicked.connect(self.switch_image2)
         treeview.itemSelectionChanged.connect(self.switch_image3)
         self.treeview = treeview
-        
+
         # create the column headers
         col = 0
         l = []
@@ -88,8 +88,9 @@ class Contents(GingaPlugin.GlobalPlugin):
     def switch_image3(self):
         with self.toc_lock:
             items = list(self.treeview.selectedItems())
-            self.switch_image2(items[0], 0)
-        
+            if len(items) > 0:
+                self.switch_image2(items[0], 0)
+
     def get_info(self, chname, name, image):
         path = image.get('path', None)
         future = image.get('image_future', None)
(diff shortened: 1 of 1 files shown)


ejeschke:
@pllim, in your testing, does this fix issue #80 ?

pllim:
As far as I know, I haven't seen the traceback messages with this fix.

ejeschke:
I think I'll review whether to back out the lock code then.  There is only one GUI thread and so it should be the only thread clearing the table.  A reentrant lock doesn't do anything for this and a regular lock would freeze the GUI.  It sounds like more of a case where the Qt selection for the table is not always in sync with the table itself.  I don't track this and sort of assumed that Qt would do so.

pllim:
There's only one thread? I see multiple "thread being created" messages when I start Ginga. There's also an option to specify how many threads to use, the default being 30.

ejeschke:
There is a thread pool, but only one thread ever runs the GUI code.  The GUI toolkits are notoriously bad for reentrant code.  The other threads are used to offload tasks like opening files, network requests for catalogs, etc. That way the GUI also stays more responsive during those activities.

pllim:
While we're on this topic, is it possible for two different threads to access the same `Datasrc` simultaneously? Since my plugin actually modifies the image stored in `Datasrc`, this might be a problem.

ejeschke:
Yes, it's possible, but there is a lock to mitigate the access. Normally, other threads are only adding new images to a datasrc, they are not doing anything with the data fields of the images that are in the datasrc. Once you have a handle to the image you can modify it.

ejeschke:
@pllim, could you test commit 8cf7dff63bb8b617cd1b94c6620528e84b1f1f5c with the conditions that caused the error message in Contents?  I've backed out the thread lock and kept your fix.  If this seems to solve things I'd like to close this issue.

pllim:
@ejeschke , yes, it works. Just FYI, now that I have your logger also prints real time info to screen, I think this problem stemmed from the fact that every time I click to switch to any of the 20 images, Ginga does a "large image update" that takes about 0.35 seconds per update.

ejeschke:
The "large" in that message simply refers to the main widget associated with the channel, not that the image is actually large.  BTW, the time looks reasonable, depending on your hardware and configuration, and whether the image is in memory or not already.  On my 3 year old Dell desktop, running Ubuntu 14.04, loading files over NFS an 18MB image is about 0.15 sec for "large image update" if already in memory.

pllim:
Okay, not that then. Let's see.... The problem happened at the step where I modify the image, set the new data to the image object, and redo ToC; So somehow setting new data and redoing ToC in the next line caused the race condition. Regardless, it is fixed now.

ejeschke:
Great, Tnx.

Issue #93, opened 2015-03-31
Primary header should be inherited


pllim:
Currently, in an exposure with `[PRIMARY, SCI, ERR, DQ]` extensions, I have to explicitly use `MultiDim` to access the primary header. A more intuitive behaviour for this format is to have `Header` of `[SCI, ERR, DQ]` to automatically inherit `PRIMARY` header, and not even display `PRIMARY` as an option in `MultiDim`. This format is widely used for HST (and possibly JWST).

I have an idea on how to do this by inheriting your `Header` class and then load the `PRIMARY` with brute force. Is there a more elegant way to do this?

ejeschke:
Hi @pllim, my thinking is that the best place to do this currently is by modifying `MultiDim`.  Because it is simply creating new `AstroImage` objects and inserting them into the viewer, it can simply grab the main header instead of the HDU header (take a look at the [code](https://github.com/ejeschke/ginga/blob/master/ginga/misc/plugins/MultiDim.py) in the `set_hdu` method).

This then becomes a simple setting (preference) that we would create for `MultiDim`, similar to the preferences that there are for other plugins.  I think this is a few line mod at most.  Take a look at `MultiDim` and tell me what you think.  If you don't want to submit a PR I can probably get to this in the next few days.

pllim:
Not really. I think `io_fits.py` is a better place for this.

pllim:
Anyways, I'll submit a PR later today and we can discuss. :)

ejeschke:
I'm testing this PR now.  Is it merging the primary header with the individual HDU headers?  The header seems to be changing as I scroll through the HDUs in `MultiDim`.  So it is not just displaying the primary header, but merging it with the HDU header, with the HDU header keywords having preference?

pllim:
Yes, but let me explain. My idea is to display the header such that the original header is displayed first, followed by inherited primary header. If there are conflicting keywords, the ones in primary are ignored. As you can see, I merge them in `AstroImage.get_header()` method without changing `AstroImage.metadata['header']`. So theoretically, I can still extract only the original header by calling `image.metadata['header']`, right? My idea is also to make `AstroImage.inherit_primary_header` configurable using your built-in config system, but I am not very sure how to do that elegantly.

ejeschke:
I see that some HST images have a FITS keyword `INHERIT=T` in the HDU headers.  Should we make this behavior only happen if that keyword exists and its value is true?

pllim:
[...]

PR #97, opened 2015-04-02
Inherit primary header

pllim:
PR for #93

@ejeschke , this works for me. I marked where you could change this behavior into a user preference but did not implement it. Sorry for the auto trailing whitespace removal. Please let me know what you think. Thanks.


ejeschke:
This has been modified as per issue #93 and merged.

Issue #95, opened 2015-04-02
Mosaic crashes in create_blank_image()


pllim:
Hi again. I dragged an image file from your file browser plugin onto the display area to make a mosaic, I get the following traceback:
```
2015-04-02 13:41:40,111 | I | Mosaic.py:360 (drop_cb) | files dropped: ['file:///mypath/image.fits']
2015-04-02 13:41:40,122 | D | Task.py:929 (execute) | now executing task '<ginga.main.Ginga object at 0x7f058f163c10>.task1'
2015-04-02 13:41:40,123 | D | Task.py:369 (execute) | Running error_wrap((<bound method Mosaic.mosaic of <Mosaic.Mosaic object at 0x7f050c279690>>, ['file:///mypath/images.fits']), {'new_mosaic': True})
2015-04-02 13:41:40,134 | D | Control.py:471 (guess_filetype) | MIME type is image/fits
2015-04-02 13:41:40,135 | D | Control.py:497 (load_image) | assuming file type: image/fits'
2015-04-02 13:41:40,135 | I | Control.py:507 (load_image) | Loading image from file:///mypath/image.fits kwdargs={'numhdu': None}
2015-04-02 13:41:40,135 | D | AstroImage.py:99 (load_file) | Loading file 'file:///mypath/image.fits' ...
2015-04-02 13:41:40,136 | D | io_fits.py:129 (load_file) | Loading file '/mypath/image.fits' ...
2015-04-02 13:41:40,239 | D | Control.py:523 (load_image) | Successfully loaded file into image object.
2015-04-02 13:41:40,239 | D | ImageView.py:896 (get_data_xy) | before adjustment, win_x=240 win_y=232
2015-04-02 13:41:40,245 | D | ImageView.py:912 (get_data_xy) | data_x=-128 data_y=128
2015-04-02 13:41:40,246 | D | Bindings.py:1649 (window_motion) | Event map for (None, 'nobtn')
2015-04-02 13:41:40,342 | D | Mosaic.py:227 (prepare_mosaic) | image0 rot=0.000000 cdelt1=0.000009 cdelt2=0.000009
2015-04-02 13:41:40,342 | D | Mosaic.py:234 (prepare_mosaic) | creating blank image to hold mosaic
2015-04-02 13:41:40,343 | D | ImageView.py:753 (get_rgb_object) | times: total=0.0000
[...]

PR #101, opened 2015-04-14
Brainstorming for #95

pllim:
As discussed in #95, here are some ideas for the `Mosaic` plugin. Again, please ignore my Emacs automatically removing trailing whitespaces. I could not effectively catch the `MemoryError`, but if we calculate FOV dynamically (or even just set the default smaller), we can avoid the error from happening in the first place.

* I added error class name to be displayed in `Control.py` (Line 420). I find this useful.
* I unchecked "Drop New" by default. Intuitively for me, when I drop new file, I want to add them to the existing mosaic, not overwrite it.
* I added `_calc_fov()` method in `Mosaic.py` to roughly estimate the FOV needed to contain all the images. However, this does not work if user created the mosaic from one image, and then drop in a different image separately. It works if user CNTRL+click on all the images and drop them in at once. Like you said, there can be a separate method that stores all the RA and DEC info and recalculate FOV when user drops in new file separately, and use that to create bigger mosaic and paste the existing mosaic in, but I can't implement that without butchering the existing codes. Maybe you know better where to stick that kind of logic in?
* I disabled creation of new mosaic if `dist > fov_deg` because it is not a behavior I want. But I guess we don't have to disable this if we can calculate `fov_deg` such that it is always bigger than `dist`.
* You can ignore the changes in `dp.py`. I inserted memory check there but it didn't stop `MemoryError` from happening.

### DIFF
--- a/ginga/Control.py
+++ b/ginga/Control.py
@@ -51,7 +51,7 @@ class ControlError(Exception):
     pass
 
 class GingaControl(Callback.Callbacks):
-     
+
     def __init__(self, logger, threadPool, module_manager, preferences,
                  ev_quit=None):
         Callback.Callbacks.__init__(self)
@@ -82,7 +82,7 @@ class GingaControl(Callback.Callbacks):
         self.timer_factory = Timer.TimerFactory(ev_quit=self.ev_quit)
         task = Task.FuncTask2(self.timer_factory.mainloop)
         task.init_and_start(self)
-        
+
         self.lock = threading.RLock()
         self.channel = {}
         self.channelNames = []
@@ -118,7 +118,7 @@ class GingaControl(Callback.Callbacks):
 
         self.fn_keys = ('f1', 'f2', 'f3', 'f4', 'f5', 'f6', 'f7', 'f8',
                         'f9', 'f10', 'f11', 'f12')
-        
+
         self.global_plugins = {}
         self.local_plugins = {}
         self.operations  = []
@@ -147,7 +147,7 @@ class GingaControl(Callback.Callbacks):
     ####################################################
     # CALLBACKS
     ####################################################
-    
+
     def showxy(self, fitsimage, data_x, data_y):
         try:
             image = fitsimage.get_image()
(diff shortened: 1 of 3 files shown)


ejeschke:
Hi @pllim, can you check out commit bc0f9600ffacd228fc1c16c66effb3620cdd6e1f before we go further with this PR?  This adds support for dynamically expanding the mosaic when new tiles are dropped.  See `examples/configs/plugin_Mosaic.cfg` for all the possible settings.  Some of these need to be exposed so they can be changed on the fly in the GUI--this is TBD.

With this behavior, I'm wondering if we need to do an initial pass to calculate the fov.

One other issue with your PR as it stands is that it assumes the reference pixel for the WCS is in the center of the image tiles.  This is true for your sample images, but not required in practice, I think (in fact the reference pixel need not be contained in the data part of the file).  So some typical mosaics that I tried with Subaru instrument files failed.  If we keep the fov calculation (which we maybe could as an option), we'll need to account for this.

pllim:
@ejeschke , I like your changes. It does not look like we need to pre-calculate FOV anymore. So, you can probably ignore this PR, because I opened another PR for the one change that I still wish to see for the traceback.

Regarding your changes, I notice that when I set my preference for `Mosaic` to create new thumbnails, I see the new entries in `Contents` but not `Thumbs`. In addition, when I have "label images" turned on, a new label overwrites all the previous ones in all the mosaic entries I see in `Contents`, which does not make sense visually.

ejeschke:
I see what you are saying about the labels.  It only works if the image is not resized. That's going to require some more drastic workaround, like clearing and redrawing all the labels if the image changes.  Until that happens, I've added a test to only draw the labels if allow_expand==False.

Regarding the creation of the Thumb for a mosaic, I think the problem is that there is no path associated with the mosaic.  Thumbs uses a file's path as part of a key to identify the thumb (since there can be different files with the same name and also because if a thumbnail is cached).  I think Contents doesn't actually check for a path, since it sorts files by channel.

Then there is also the problem that if the data drops out of the `Datasrc` then clicking on the thumbnail should reconstruct the mosaic from the tiles.  We actually have this capability in the subclassed plugin we use here, but it only creates Contents entries, not Thumbs entries.  The new ability to keep dropping new tiles on the mosaic could also complicate this.

In short, this is a bit sticky problem and requires some consideration.

pllim:
It is fine for me if `Thumbs` does not display mosaic thumbnails, as long as it is documented clearly. I like that I can still access it from `Contents`.

Currently, this entry in `plugin_Mosaic.cfg` is very misleading:
```
# Allow mosaic images to create thumbnail entries
make_thumbs = False
```

pllim:
In addition, the mosaic names in `Contents` are not very descriptive. Perhaps a new field in `Mosaic` plugin to allow user to overwrite mosaic name (or at least, the `dp` prefix) when user starts a new one?

PR #102, opened 2015-04-15
Expose exception class name in Errors plugin

pllim:
When I see traceback on `Errors` plugin, I want to know the type of exception associated with it. This adds the exception class name to the traceback shown.
 
The actual change is on Line 420. The rest is just my Emacs auto-removing trailing whitespace.

This was part of #101, but should be a PR in its own right.

### DIFF
--- a/ginga/Control.py
+++ b/ginga/Control.py
@@ -51,7 +51,7 @@ class ControlError(Exception):
     pass
 
 class GingaControl(Callback.Callbacks):
-     
+
     def __init__(self, logger, threadPool, module_manager, preferences,
                  ev_quit=None):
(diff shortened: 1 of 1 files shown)


ejeschke:
@pllim, what is your emacs setting to automatically remove trailing white space?  As an emacs user, it seems that I ought to be using that to avoid the kind of noise this generates in PRs.

pllim:
@ejeschke , you might find http://docs.astropy.org/en/stable/development/codeguide_emacs.html useful. Personally, this is what I do in my `.emacs` file:
```
;; Automatically remove trailing whitespace when file is saved.
(add-hook 'before-save-hook 'delete-trailing-whitespace)
```

PR #104, opened 2015-04-21
use astropy frames

Cadair:


ejeschke:
Merged manually in commit e6ce60fc71e4942b2221f70b35a26ebf757968e6.  Still need to fix the bad RA check for sunpy, so leaving this PR open until we fix that.

PR #108, opened 2015-05-06
Add a new Astropy 1.0+ WCS Class

Cadair:
This uses the new Astropy FITS WCS -> Coordinate infrastructure to detect the coordinate type from the header. Also It uses a few dirty tricks to speed things up by reducing the amount of times a Frame is initialized to once per file.


Cadair:
@ejeschke I have tested this on a couple of images, but I don't know if I have got everything sorted.

ejeschke:
@Cadair, this looks fine.  Wondering though, about the long term viability of supporting both 'astropy' and 'astropy2' as wcs schemes.  Think they can be merged if the appropriate PRs get merged upstream in astropy?  Also, does the speed improvement depend on PR 3737?  I see a `hasattr()` test for this method in wcsmod.  Presumably the else clause does it by brute force?

Cadair:
So the main reason for keeping them separate was it made the code cleaner, less version catching stuff.

The astropy2 class is 1.0+ only, so at the point you want to drop support for pre astropy 1.0 you could remove the old one.

The has attr hacks in there at the moment are to provide 1.0 compatibility and to astropy/astropy#3757.  We could also add a check for the copy=False argument to realize_frame being suggested in that PR.

Issue #79, opened 2015-02-18
First cut levels not applied when typed in Info plugin, first time only


ejeschke:
Affects: reference viewer

When Ginga is first started and an image is loaded, typing a value for one of the cut levels in the Info plugin box and pressing enter does not apply the value.  Pressing a second time applies the value, and it appears that any subsequent values entered or pressed are applied correctly.

From looking at the log, it appears that for this initial case, the value is applied, but then reset.

pllim:
Maybe this is related: When I have `autocuts='off'` set in `channel_Image.cfg` and I display a bunch of images at once using `ginga *.fits`, the "Cut Low" and "Cut High" values are both 0. I was expecting that the initial cut levels will be based on one of the images (first or last; doesn't matter to me), but then are fixed at those values until user explicitly change them. Having default 0/0 limits is not very useful.

ejeschke:
Fixed.

PR #122, opened 2015-06-17
Attempt to guess cut levels when autocuts is off

pllim:
When `autocuts` is off, set initial cut levels to `autolevels` values from the first image. This is my attempt to fix the problem *I* reported in #79 (not the original issue).

### DIFF
--- a/ginga/ImageView.py
+++ b/ginga/ImageView.py
@@ -503,6 +503,10 @@ class ImageViewBase(Callback.Callbacks):
 
             if self.t_['autocuts'] != 'off':
                 self.auto_levels(redraw=False)
+            else:
+                cut_lo, cut_hi = self.get_cut_levels()
+                if cut_lo == cut_hi == 0:
+                    self.auto_levels(redraw=False)
 
         except Exception as e:
             self.logger.error("Failed to initialize image: %s" % (str(e)))


ejeschke:
Hi @pllim, I understand what you are trying to do here, but I wonder about whether this is the right way to go about it.  I suppose the default cut levels are lo==hi==0 when nothing has been saved to the preferences, but after the first save to channel_Image.cfg the cut levels will (may) have different values (most likely different from 0) and they could be floating pt values, both cases will not be caught by this check.

To test this, can you set your cut levels to (say) 0 and 1000 and then save your preferences for the channel you are testing with (e.g. "Image").  Then see whether this PR works--my guess is no.

Setting autocuts to "off" essentially means do not attempt to perform an autocuts.  Would it not make more sense to set it to "override" in your case?  Then it will do a cut levels until you explicitly change it.

pllim:
I tried "override" but it is not what I needed. I don't want the cut levels to change when I switch image, but I also don't want to see a (0, 0) cut when I first load the images.

> after the first save to channel_Image.cfg the cut levels will (may) have different values (most likely
> different from 0) and they could be floating pt values, both cases will not be caught by this check.

If user already has non-zero cut levels saved and `autocuts=off`, there is no reason to overwrite the user setting, right?

> Setting autocuts to "off" essentially means do not attempt to perform an autocuts. 

I agree. But software default of (0, 0) does not quite make sense to me.

I am open to suggestions.

ejeschke:
Hi @pllim, I think I might have a solution for you.  Could you try commit 79d512527296f7d7b5b0d4c5fc1954acfccef163 (which incidentally fixes an outstanding cut levels issue #79)?

Set your `autocuts` option to "once".  Incidentally, this once setting can also be applied to autocenter and autozoom, and can be selected from Preferences.  Let me know if this works for you.

pllim:
Yes, "once" works for me. Thanks!

PR #123, opened 2015-06-18
--Do Not Merge-- MultiDim fixes

rupak0577:
A few fixes for MultiDim -

1) Fixes the activation problem of the slider

2) Fixes incorrect number of HDUs shown

3) Loads MultiDim if file opened has NAXIS >= 3

4) Fixes Slice display issue

### DIFF
--- a/ginga/misc/plugins/MultiDim.py
+++ b/ginga/misc/plugins/MultiDim.py
@@ -490,6 +490,7 @@ class MultiDim(GingaPlugin.LocalPlugin):
         if idx is not None:
             # set the HDU in the drop down if known
             self.w.hdu.set_index(idx)
+            self.set_hdu(idx)
 
         self.w.hdu.set_enabled(len(self.fits_f) > 0)
 
--- a/ginga/misc/plugins/MultiDim.py
+++ b/ginga/misc/plugins/MultiDim.py
@@ -485,8 +485,8 @@ class MultiDim(GingaPlugin.LocalPlugin):
         info = self.fits_f.info(output=False)
         self.prep_hdu_menu(self.w.hdu, info)
         self.num_hdu = upper
-        self.logger.debug("there are %d hdus" % (upper))
-        self.w.numhdu.set_text("%d" % (upper))
+        self.logger.debug("there are %d hdus" % (upper+1))
+        self.w.numhdu.set_text("%d" % (upper+1))
         if idx is not None:
             # set the HDU in the drop down if known
             self.w.hdu.set_index(idx)


ginga/qtw/GingaQt.py
```diff
@@ -633,7 +633,11 @@ def new_ws_cb(self, w, rsp, b, names):
     def gui_load_file(self, initialdir=None):
         if self.filesel.exec_():
             fileNames = list(map(str, list(self.filesel.selectedFiles())))
-            self.load_file(fileNames[0])
+            lf = self.load_file(fileNames[0])
+
+        # Load MultiDim plugin if file is NDIM
```
ejeschke: I think this needs to be indented to fall under the if.  Otherwise variable `lf` is not defined.  Suggest variable `image` instead of `lf`, as this would be consistent with other usage.

Cadair:
:+1:

ginga/misc/plugins/MultiDim.py
```diff
@@ -414,6 +414,7 @@ def prep_hdu_menu(self, w, info):
             idx = 0
         if idx >= len(self.hdu_info):
             idx = len(self.hdu_info) - 1
+        self.set_hdu(0)
```
ejeschke: How can you be sure that HDU0 has a valid data segment?

ginga/qtw/GingaQt.py
```diff
@@ -633,7 +633,11 @@ def new_ws_cb(self, w, rsp, b, names):
     def gui_load_file(self, initialdir=None):
         if self.filesel.exec_():
             fileNames = list(map(str, list(self.filesel.selectedFiles())))
-            self.load_file(fileNames[0])
+            image = self.load_file(fileNames[0])
```
ejeschke: @rupak0577, because of the location of the detection code here, it will only open the MultiDim plugin if the file is opened via the File menu.

ejeschke:
@rupak0577, if you check commit 1bafee7c2be7027e70b3774f3c21ab4f0f7c8cd0, I have incorporated part of your PR there.  Specifically, it registers a callback in the plugin for new images appearing in the channel, if the user has set a preference `auto_start_naxis=True` and a newly loaded image has `NAXIS>=3` then it automatically opens the plugin.  That way it works no matter how the image is loaded (and also is not specific to Qt).

ejeschke:
These other issues are not necessarily addressed, so maybe revert the change to `ginga/qtw/GingaQt.py` and rebase against master?

rupak0577:
@ejeschke That is a better solution. Thanks!

nabobalis:
This can be merged. I do not see an issue here. @ejeschke

ejeschke:
Great, merging.

PR #124, opened 2015-06-18
--Do Not Merge-- Cuts fixes

rupak0577:
All fixes related to Cuts

### DIFF
--- a/ginga/misc/plugins/Cuts.py
+++ b/ginga/misc/plugins/Cuts.py
@@ -505,8 +505,6 @@ Keyboard shortcuts: press 'h' for a full horizontal cut and 'j' for a full verti
         self.canvas.add(cut, tag=tag)
         select_flag = self.settings.get('select_new_cut', True)
         self.add_cuts_tag(tag, select=select_flag)
-        self.w.delete_cut.set_enabled(True)
-        self.w.delete_all.set_enabled(True)
 
         self.logger.debug("redoing cut plots")
         return self.redo()
@@ -528,6 +526,8 @@ Keyboard shortcuts: press 'h' for a full horizontal cut and 'j' for a full verti
         self.canvas.add(cut, tag=tag)
         select_flag = self.settings.get('select_new_cut', True)
         self.add_cuts_tag(tag, select=select_flag)
+        self.w.delete_cut.set_enabled(True)
+        self.w.delete_all.set_enabled(True)
 
         self.logger.debug("redoing cut plots")
         return self.redo()
--- a/ginga/misc/plugins/Cuts.py
+++ b/ginga/misc/plugins/Cuts.py
@@ -141,6 +141,12 @@ class Cuts(GingaPlugin.LocalPlugin):
                     )
         w, b = Widgets.build_info(captions, orientation=orientation)
         self.w.update(b)
+        if len(self.tags) > 1:
+            b.delete_cut.set_enabled(True)
+            b.delete_all.set_enabled(True)
+        else:
+            b.delete_cut.set_enabled(False)
+            b.delete_all.set_enabled(False)
 
         # control for selecting a cut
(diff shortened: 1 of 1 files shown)


ejeschke:
Can you check the current version of Cuts in master?  It's been overhauled a lot.  Maybe submit a new PR against that for any improvements.

ejeschke:
@rupak0577, are you going to add anything else to this PR?  Otherwise I could merge it--looks good.

rupak0577:
@ejeschke There has been a discussion about adding a Bezier cut but I am not sure how to go about it. Could you give me a few pointers? Also, a way to redraw the plot so that it works with MultiDim - the 'Take Tab' solution does not work.

ejeschke:
@rupak0577 , in commit 52f5313375ebd19fef863e9a7e341d1a9695324d I added support for drawing Bezier curves.  This could possibly be the basis for a Bezier cut.  Try it out via the Drawing plugin--you select beziercurve and then draw as though you were creating a polygon-type path (use 'v' to add vertexes)--right now you can only add two vertexes (plus the starting and ending points).  After the third point is established the curve is shown--release the cursor button to complete the curve.

You can then edit the curve using the edit mode I described earlier.  After entering edit mode, click on one of the ends of the curve to select it and then you can manipulate the points.

Assuming that the BezierCurve is fully functional it would then be fairly simple to add it as a cuts type to the cuts plugin. *BUT*--it needs a valid `contains_arr()` method to work.  If you have time, you can check [here](http://polymathprogrammer.com/2012/04/03/does-point-lie-on-bezier-curve/) for information on how to formulate an algorithm for `contains_arr()`.

ejeschke:
When you say "redraw the plot so that it works with MultiDim"--does this mean, so that both plugins can be seen on either side of the main window?

ejeschke:
I just tested using "Take Tab" to move the MultiDim plugin to the left side.  It works.  You have to right click (or on the mac, control+click) on the tabs on the left side to get the pop-up menu, choose IMAGE: MultiDim and it will move over.

rupak0577:
@ejeschke I did that but it doesn't update the plot until I re-click on the Cuts plugin. So I was thinking if we could redraw the plot whenever the viewer changes.

rupak0577:
Rebased the commit.

ejeschke:
@rupak0577 , sorry I am still not understanding completely.  When I choose a different HDU in MultiDim, even when it is on the left side--the plot updates, even if Cuts does not have the focus.

Cuts needs to have the focus (green color) to draw new cuts or reposition a cut--this is because when you change the focus to a new plugin most plugins will disable focus in the `pause()` method--this is to prevent input conflicts--imagine both Pick and Cuts responding to the mouse, for example.

Or is this about making the Cuts update the plot when a new slice is selected through the NAXIS controls?  That we could fix pretty easily, I think.

rupak0577:
@ejeschke Yeah, that's what I meant - to update the plot when a new slice is selected.

rupak0577:
@ejeschke How should the contains_arr() method work? I looked at some of the methods but I am confused as to how it works.

ejeschke:
@rupak0577 , I have moved the discussion on the Bezier cut over to a new issue #136 , to keep this PR from getting stalled on Bezier cuts implementation.  See comment over there.

ejeschke:
@rupak0577, commit 759a84d77730fd9c0ef991ad1417f80ce64a380a adds support for updating the cut when a new slice is selected.

ejeschke:
@rupak0577, could you rebase this PR against master one last time?  I will then merge it.  I believe this will then address all points in Issue #131 except for the Bezier cut, which is now in issue #136.  If you want to queue up some more cuts improvements lets do a new PR.

ejeschke:
Cleaned up slightly and merged.

PR #125, opened 2015-06-19
Save support

rupak0577:
Added save support to-

1) Cuts - Save plot and data
2) MultiDim - Save support for images and movies(uses mencoder; still has some issues)

### DIFF
--- a/ginga/misc/plugins/Cuts.py
+++ b/ginga/misc/plugins/Cuts.py
@@ -615,12 +615,14 @@ Keyboard shortcuts: press 'h' for a full horizontal cut and 'j' for a full verti
             return
 
         target = SaveDialog(title='Save plot', extfilter='*.png').get_path()
-        fig.savefig(target, dpi=100)
+        if target:
+            fig.savefig(target, dpi=100)
 
         target = SaveDialog(title='Save data', extfilter='*.npz').get_path()
-        target_file = open(target, 'w')
-        np.savez_compressed(target_file, x=xarr, y=yarr)
-        target_file.close()
+        if target:
+            target_file = open(target, 'w')
+            np.savez_compressed(target_file, x=xarr, y=yarr)
+            target_file.close()
 
     def __str__(self):
         return 'cuts'
--- a/ginga/misc/plugins/Cuts.py
+++ b/ginga/misc/plugins/Cuts.py
@@ -12,7 +12,6 @@ import numpy
 from ginga.misc import Widgets, Plot
 from ginga import GingaPlugin, colors
 from ginga.util.six.moves import map, zip
-from ginga.qtw.QtHelp import SaveDialog
 
 # default cut colors
 cut_colors = ['magenta', 'skyblue2', 'chartreuse2', 'cyan', 'pink',
@@ -274,9 +273,9 @@ Keyboard shortcuts: press 'h' for a full horizontal cut and 'j' for a full verti
         idx = len(self.tags) - 1
         tag = self.tags[idx]
         self.select_cut(tag)
-        # plot cleared in redo() if no more cuts
         if tag == self._new_cut:
             self.save_btn.set_enabled(False)
+        # plot cleared in redo() if no more cuts
         self.redo()
 
     def delete_all_cb(self, w):
@@ -286,8 +285,8 @@ Keyboard shortcuts: press 'h' for a full horizontal cut and 'j' for a full verti
         self.cutstag = self._new_cut
         self.w.cuts.append_text(self._new_cut)
         self.select_cut(self._new_cut)
-        # plot cleared in redo() if no more cuts
         self.save_btn.set_enabled(False)
+        # plot cleared in redo() if no more cuts
         self.redo()
 
     def add_cuts_tag(self, tag):
@@ -395,6 +394,10 @@ Keyboard shortcuts: press 'h' for a full horizontal cut and 'j' for a full verti
                 text.color = colors[0]
             #text.color = color
             self._redo(lines, colors)
+            if n:
+                self.w.delete_cut.set_enabled(True)
+                self.w.delete_all.set_enabled(True)
+                self.save_btn.set_enabled(True)
 
         # force mpl redraw
         self.plot.fig.canvas.draw()
@@ -579,8 +582,6 @@ Keyboard shortcuts: press 'h' for a full horizontal cut and 'j' for a full verti
         self.canvas.add(cut, tag=tag)
         self.add_cuts_tag(tag)
 
-        self.save_btn.set_enabled(True)
-
         self.logger.debug("redoing cut plots")
         return self.redo()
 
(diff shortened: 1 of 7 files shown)


ejeschke:
Is `OpenCv` an option to create videos?

rupak0577:
Unfortunately, it seems OpenCv has some problems but I will try.

ejeschke:
@rupak0577, can you rebase against master?  @Cadair, do you or Nabil have any comments on this PR?

Cadair:
@nabobalis

nabobalis:
Yes?

Cadair:
See above.

rupak0577:
@ejeschke @nabobalis There is a small problem with the cut-offs for the movie that I can't figure out -
img

I would like to know if it can be improved in any other way and review before merging.

nabobalis:
The cut save, does not seem to save a plot at all. The same goes for the save slice from MultiDim. The dialog box comes up, I enter a name and hit save. Nothing is there.

The save movie button seems to do nothing.

rupak0577:
@nabobalis The problem is with MultiDim. You have to use the slider before you can save an image. But the save movie should work anyway. Simply enter a slice range and click save.

@ejeschke It seems the plot does not clear the data in `redo()` for Cuts which brings up the save window even if the plot is cleared.

nabobalis:
@rupak0577 Anyway to change that? It should just save the image regardless of having used the slider.

rupak0577:
@nabobalis It is because MultiDim does not load the first slice.

@ejeschke Anyway to fix that? Also, it does not show the Slice and Value data which could be useful here.

rupak0577:
@ejeschke @nabobalis I have fixed all bugs that I could except the first slice issue. Needs a review.

Cadair:
@rupak0577 if this is ready for review, I suggest you edit the title and remove the don't merge marker.

ginga/qtw/QtHelp.py
```diff
@@ -414,6 +414,16 @@ def cb():
     def get_content_area(self):
         return self.content
 
+class SaveDialog(QtGui.QFileDialog):
```
Cadair: Do we need to define one of these for GTK if we want this to work under GTK?

rupak0577: Hmm, I had forgotten about that. Will add it.

ginga/misc/plugins/Cuts.py
```diff
@@ -12,6 +12,7 @@
 from ginga.misc import Widgets, Plot
 from ginga import GingaPlugin, colors
 from ginga.util.six.moves import map, zip
+from ginga.qtw.QtHelp import SaveDialog
```
ejeschke: Because this plugin is in `.../ginga/misc/plugins`, it is backend-independent.  This is nice because it avoids having to maintain two separate versions of the plugin.  In order to maintain that, we need to have a generic `SaveDialog` available in the `Widgets`. Just make sure there is one in `ginga.gtkw.Widgets` that works the same.

ginga/misc/plugins/Cuts.py
```diff
@@ -590,6 +602,19 @@ def set_mode_cb(self, mode):
...
+        fig, xarr, yarr = self.plot.get_data()
+
+        target = SaveDialog(title='Save plot', extfilter='*.png').get_path()
+        if target:
+            fig.savefig(target, dpi=100)
+
+        target = SaveDialog(title='Save data', extfilter='*.npz').get_path()
+        if target:
+            target_file = open(target, 'w')
+            numpy.savez_compressed(target_file, x=xarr, y=yarr)
```
ejeschke: Is `savez_compressed` part of numpy 1.7 and up?

ejeschke: Recommend here using a context manager:
```python
with open(target, 'w') as target_file:
    numpy.savez_compressed(target_file, x=xarr, y=yarr)
```

ginga/misc/plugins/MultiDim.py
```diff
@@ -38,6 +48,7 @@ def __init__(self, fv, fitsimage):
         self.imgname = 'NONAME'
         self.image = None
         self.orientation = 'vertical'
+        self.curr_slice = []
```
ejeschke:
```python
self.curr_slice = None
```

ginga/misc/plugins/MultiDim.py
```diff
@@ -493,6 +532,46 @@ def redo(self):
 
         self.w.hdu.set_enabled(len(self.fits_f) > 0)
 
+    def save_slice_cb(self):
+        target = SaveDialog(title='Save slice', extfilter='PNG image (*.png)').get_path()
+        if target:
+            hival = self.fitsimage.get_cut_levels()[1]
+            plt.imsave(target, self.curr_slice, vmax=hival, cmap=plt.get_cmap('gray'), origin='lower')
+            self.fv.showStatus("Successfully saved slice")
```
ejeschke: Suggest:
```python
image = self.fitsimage.get_image()
curr_slice = image.get_data()
```
Creating instance variables that hold on to potentially large arrays makes me nervous...

ginga/misc/plugins/MultiDim.py
```diff
@@ -493,6 +532,46 @@ def redo(self):
...
+        data = np.array(hdu.data, copy=True).clip(loval, hival)
+
+        data_rescaled = ((data - loval) * 255 / (hival - loval)).astype(np.uint8, copy=False)
+        H, W = data.shape[1:]
+
+        video = VideoSink((H, W), target)
+        for i in xrange(start, end):
+            video.run(np.flipud(data_rescaled[i]))
+        video.close()
+
```
ejeschke: Wondering if we can make the `VideoSink` class act like a context manager, so you could do:
```python
with VideoSink((H, W), target) as video:
    for i in xrange(start, end):
        video.run(np.flipud(data_rescaled[i]))
```
The nice thing about context managers is that they automatically clean up and don't leave open files or processes around if the encoding craps out in the middle due to some error...

ejeschke:
@rupak0577 , I've left comments on the key items above, most of which are minor.  Plugin seems to work for me after installing mencoder, and gave me a reasonable behavior when it wasn't installed.

Fix up what you can and I'll take a look at fixing the problem of loading the first slice.  Think we can merge this soon if the others are happy.

rupak0577:
@ejeschke Thanks! Will update it as per your points.

rupak0577:
@ejeschke Updated.

ejeschke:
Merged.  @rupak0577, I made some minor tweaks, just to improve the appearance (esp under gtk).  Would appreciate testing by @nabobalis, @Cadair and @rupak0577 

Found an issue--"save slice" doesn't add a file extension if omitted.  The dialog leads you to think that it might.  The "save movie" seemed to add one however.  I think this should behave the same for both cases and should add a default extension if the user does not provide one.

PR #126, opened 2015-06-24
Contents treeview and nameDict access

pllim:
* Fixed a bug where `Contents` `treeview` in Qt does not expand after being updated.
* Make `fileDict` use a consistent API to access `nameDict` (Qt only).

### DIFF
--- a/ginga/qtw/plugins/Contents.py
+++ b/ginga/qtw/plugins/Contents.py
@@ -66,7 +66,7 @@ class Contents(GingaPlugin.GlobalPlugin):
 
 
     def switch_image(self, chname, imname):
-        fileDict = self.nameDict[chname]
+        fileDict = self.get_contents_by_channel(chname)
         key = imname.lower()
         bnch = fileDict[key]
         path = bnch.path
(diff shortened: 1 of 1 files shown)


pllim:
I think Travis failed because you require Numpy 1.7 or higher but you have a test for Numpy 1.6.

ejeschke:
@pllim, thanks for catching the numpy issue--I've updated the travis build

I think this PR is not strictly necessary, the reason being that nameDict went from being a regular dictionary to being a caseless dictionary.  But maybe this makes the code a little easier to read.  I'll add the change to gtk version and merge it.

pllim:
It is fine if you don't want to change how `fileDict` is set, but I would like `self.treeview.expandAll()` to be incorporated if you are okay with it.

ejeschke:
I made the auto expand a Contents preference, because I think some users will have a workflow that involves a lot of images and will not necessarily want the tree to be auto expanding when new images are added or deleted.  I left the default to True for now.

pllim:
Okay, thanks!

PR #129, opened 2015-06-28
Slit support

rupak0577:
Adds Slit support to Cuts

### DIFF
--- a/ginga/canvas/types/basic.py
+++ b/ginga/canvas/types/basic.py
@@ -585,11 +585,15 @@ class BezierCurve(Path):
         return self.select_contains_points(viewer, points, data_x, data_y)
 
     # TODO: this probably belongs somewhere else
-    def get_pixels_on_curve(self, image):
+    def get_pixels_on_curve(self, image, getvalues=True):
         data = image.get_data()
         wd, ht = image.get_size()
-        res = [ data[y, x] if 0 <= x < wd and 0 <= y < ht else numpy.NaN
-                for x, y in self.get_points_on_curve(image) ]
+        if getvalues:
+            res = [ data[y, x] if 0 <= x < wd and 0 <= y < ht else numpy.NaN
+                    for x, y in self.get_points_on_curve(image) ]
+        else:
+            res = [ [x, y] if 0 <= x < wd and 0 <= y < ht else numpy.NaN
+                    for x, y in self.get_points_on_curve(image) ]
         return res
 
     def draw(self, viewer):
--- a/ginga/misc/plugins/Cuts.py
+++ b/ginga/misc/plugins/Cuts.py
@@ -342,6 +342,8 @@ Keyboard shortcuts: press 'h' for a full horizontal cut and 'j' for a full verti
             self.w.btn_move.set_enabled(False)
             self.w.btn_edit.set_enabled(False)
             self.set_mode('draw')
+
+            self.plot2.clear()
         else:
             self.w.delete_cut.set_enabled(True)
             self.w.delete_all.set_enabled(True)
@@ -351,6 +353,13 @@ Keyboard shortcuts: press 'h' for a full horizontal cut and 'j' for a full verti
 
             if self.w.btn_edit.get_state():
                 self.edit_select_cuts()
+
+            # Redraw slit plot
+            obj = self.canvas.getObjectByTag(tag)
+            line = self._getlines(obj)
+            self._plot_slit(line[0])
+
+        self.plot2.fig.canvas.draw()
 
     def cut_select_cb(self, w, index):
         tag = self.tags[index]
@@ -561,7 +570,16 @@ Keyboard shortcuts: press 'h' for a full horizontal cut and 'j' for a full verti
                 points.extend(coords[:-1])
                 x1, y1 = x2, y2
         elif obj.kind == 'beziercurve':
-            #points = obj.get_pixels_on_curve(image)
+            coords = obj.get_pixels_on_curve(image, getvalues=False)
+            # Exlude NaNs
+            coords = [c for c in coords if not numpy.any(numpy.isnan(c))]
+
+        shape = image.shape
+        # Exclude points outside boundaries
+        coords = [(coord[0], coord[1]) for coord in coords if 0 <= coord[0] < shape[1]
+                  and 0 <= coord[1] < shape[0]]
+        if not coords:
+            self.plot2.clear()
             return
 
         coords = numpy.array(coords)
--- a/ginga/misc/plugins/Cuts.py
+++ b/ginga/misc/plugins/Cuts.py
@@ -100,8 +100,8 @@ class Cuts(GingaPlugin.LocalPlugin):
         self.cuttypes = ['line', 'path', 'freepath', 'beziercurve']
         self.cuttype = 'line'
         self.save_enabled = False
-        self.axes_states = []
         self.transpose_enabled = False
+        self.selected_axis = None
 
         # For collecting data orthogonal to the cut
         self.widthtypes = ['none', 'x', 'y', 'perpendicular']
@@ -133,6 +133,9 @@ class Cuts(GingaPlugin.LocalPlugin):
         #canvas.set_callback('key-press', self.keydown)
         canvas.setSurface(self.fitsimage)
         self.canvas = canvas
+
+        # register for new image notification in this channel
+        fitsimage.set_callback('image-set', self.new_image_cb)
 
         self.gui_up = False
 
@@ -324,17 +327,13 @@ class Cuts(GingaPlugin.LocalPlugin):
 
     def build_axes(self):
         self.hbox_axes.remove_all()
+        self.selected_axis = None
         image = self.fitsimage.get_image()
         if image is not None:
-            self.axes_states = []
-            # For easier mapping of indices with the axes
-            self.axes_states.append(None)
-
             # Add Checkbox widgets
-            for i in xrange(1, len(image.get_mddata().shape)+1):
-                name = 'NAXIS%d' % i
-                chkbox = Widgets.CheckBox(name)
-                self.axes_states.append(False)
+            # `image.naxispath` returns only mdim axes
+            for i in xrange(1, len(image.naxispath)+3):
+                chkbox = Widgets.CheckBox('NAXIS%d' % i)
                 self.hbox_axes.add_widget(chkbox)
 
                 # Disable axes 1,2
@@ -344,6 +343,8 @@ class Cuts(GingaPlugin.LocalPlugin):
 
                 # Add callback
                 self.axes_callback_handler(chkbox, i)
+
+            self.redraw_slit('clear')
 
     def axes_callback_handler(self, chkbox, pos):
         chkbox.add_callback('activated', lambda w, tf: self.axis_toggle_cb(w, tf, pos))
@@ -359,7 +360,10 @@ Keyboard shortcuts: press 'h' for a full horizontal cut and 'j' for a full verti
 
         if (tag == self._new_cut) or len(self.tags) < 2:
             self.w.delete_cut.set_enabled(False)
-            self.w.delete_all.set_enabled(False)
+            if len(self.tags) > 1:
+                self.w.delete_all.set_enabled(True)
+            else:
+                self.w.delete_all.set_enabled(False)
 
             self.w.btn_move.set_enabled(False)
             self.w.btn_edit.set_enabled(False)
@@ -378,9 +382,11 @@ Keyboard shortcuts: press 'h' for a full horizontal cut and 'j' for a full verti
             if self.w.btn_edit.get_state():
                 self.edit_select_cuts()
 
-            self.save_slit.set_enabled(True)
+            self.save_cuts.set_enabled(True)
+            if self.selected_axis:
+                self.save_slit.set_enabled(True)
 
-            self.redraw_slit('redraw')
+            self._plot_slit()
 
     def cut_select_cb(self, w, index):
         tag = self.tags[index]
@@ -424,9 +430,6 @@ Keyboard shortcuts: press 'h' for a full horizontal cut and 'j' for a full verti
         if not tag in self.tags:
             self.tags.append(tag)
             self.w.cuts.append_text(tag)
-
-        self.w.delete_cut.set_enabled(True)
-        self.w.delete_all.set_enabled(True)
 
         select_flag = self.settings.get('select_new_cut', True)
         if select_flag:
@@ -480,16 +483,16 @@ Keyboard shortcuts: press 'h' for a full horizontal cut and 'j' for a full verti
         """This is called when a new image arrives or the data in the
         existing image changes.
         """
-        curr_axis = self.axes_states
-        self.build_axes()
-
-        # Restore axis state
-        children = self.hbox_axes.get_children()
-        for p, val in enumerate(curr_axis):
-            if val is True:
-                children[p-1].set_state(True)
-
         self.replot_all()
+
+    def new_image_cb(self, fitsimage, image):
+        if fitsimage != self.fitsimage:
+            # Focus is not our channel-->not an event for us
+            return False
+        elif image is None:
+            return False
+        else:
+            self.build_axes()
 
     def _get_perpendicular_points(self, obj, x, y, r):
         dx = float(obj.x1 - obj.x2)
@@ -576,18 +579,11 @@ Keyboard shortcuts: press 'h' for a full horizontal cut and 'j' for a full verti
                             shadow=True, fancybox=True,
                             prop={'size': 8}, labelspacing=0.2)
 
-    def _plot_slit(self, obj):
+    def get_coords(self, obj):
         image = self.fitsimage.get_image()
-
-        if self.cutstag == self._new_cut:
-            return
 
         # Check whether multidimensional
         if len(image.naxispath) <= 0:
-            return
-
-        enabled_axes = [pos for pos, val in enumerate(self.axes_states) if val is True]
-        if not enabled_axes:
             return
 
         # Get points on the line
@@ -613,27 +609,19 @@ Keyboard shortcuts: press 'h' for a full horizontal cut and 'j' for a full verti
         coords = [(coord[0], coord[1]) for coord in coords if 0 <= coord[0] < shape[1]
                   and 0 <= coord[1] < shape[0]]
         if not coords:
-            self.plot2.clear()
+            self.redraw_slit('clear')
             return
 
-        coords = numpy.array(coords)
-        self.slit_data = self.get_slit_data(coords, enabled_axes)
+        return numpy.array(coords)
 
-        if self.transpose_enabled:
-            self.slit_data = self.slit_data.T
-
-        self.plot2.ax.imshow(self.slit_data,  interpolation='nearest',
-                             origin='lower', aspect='auto').set_cmap('gray')
-        self.set_labels()
-        self.plot2.fig.tight_layout(pad=0.3)
-
-    def get_slit_data(self, coords, enabled_axes):
+    def get_slit_data(self, coords):
         image = self.fitsimage.get_image()
         data = image.get_mddata()
         naxes = data.ndim
 
-        # Find correct indices of axes
-        enabled_axes = [abs(x - naxes) for x in enabled_axes]
+        # Small correction
+        selected_axis = abs(self.selected_axis - naxes)
+
         spatial_axes = [naxes-1, naxes-2]
 
         # Build N-dim slice
@@ -642,23 +630,37 @@ Keyboard shortcuts: press 'h' for a full horizontal cut and 'j' for a full verti
         # Slice data according to axis
         for i, sa in enumerate(spatial_axes):
             axes_slice[sa] = coords[:, i]
-        for ea in enabled_axes:
-            axes_slice[ea] = slice(None, None, None)
+        axes_slice[selected_axis] = slice(None, None, None)
 
-        return data[axes_slice]
+        self.slit_data = data[axes_slice]
+
+    def _plot_slit(self):
+        if not self.selected_axis:
+                return
+
+        obj = self.canvas.getObjectByTag(self.cutstag)
+        line = self._getlines(obj)
+
+        self.get_slit_data(self.get_coords(line[0]))
+        if self.transpose_enabled:
+            self.redraw_slit('transpose')
+        else:
+            self.plot2.ax.imshow(self.slit_data,  interpolation='nearest',
+                                 origin='lower', aspect='auto').set_cmap('gray')
+            self.set_labels()
+            self.plot2.fig.tight_layout(pad=0.3)
+            self.plot2.fig.canvas.draw()
 
     def _replot(self, lines, colors):
         for idx in range(len(lines)):
             line, color = lines[idx], colors[idx]
             line.color = color
             self._plotpoints(line, color)
-            self._plot_slit(line)
 
         return True
 
     def replot_all(self):
         self.plot.clear()
-        self.plot2.clear()
         idx = 0
         for cutstag in self.tags:
             if cutstag == self._new_cut:
@@ -677,18 +679,13 @@ Keyboard shortcuts: press 'h' for a full horizontal cut and 'j' for a full verti
                 text.color = colors[0]
             #text.color = color
             self._replot(lines, colors)
-            if n:
-                self.w.delete_cut.set_enabled(True)
-                self.w.delete_all.set_enabled(True)
-                self.save_cuts.set_enabled(True)
-                self.save_slit.set_enabled(True)
 
         # Redraw slit image for selected cut
-        self.select_cut(self.cutstag)
+        if self.cutstag != self._new_cut:
+            self._plot_slit()
 
         # force mpl redraw
         self.plot.fig.canvas.draw()
-        self.plot2.fig.canvas.draw()
 
         self.canvas.redraw(whence=3)
         self.fv.showStatus("Click or drag left mouse button to reposition cuts")
@@ -981,10 +978,6 @@ Keyboard shortcuts: press 'h' for a full horizontal cut and 'j' for a full verti
                 fig.savefig(target, dpi=100)
                 numpy.savez_compressed(target, x=xarr, y=yarr)
         elif mode == 'slit':
-            enabled_axes = [pos for pos, val in enumerate(self.axes_states) if val is True]
-            if not enabled_axes:
-                return
-
             target = Widgets.SaveDialog(title='Save slit data').get_path()
             if target:
                 fig2, xarr2, yarr2 = self.plot2.get_data()
@@ -992,31 +985,27 @@ Keyboard shortcuts: press 'h' for a full horizontal cut and 'j' for a full verti
                 numpy.savez_compressed(target, x=xarr2, y=yarr2)
 
     def axis_toggle_cb(self, w, tf, pos):
-        # Deactivate other checkboxes
         children = self.hbox_axes.get_children()
-        for p, val in enumerate(self.axes_states):
-            if val is None:
-                continue
-            elif val is True:
-                self.axes_states[p] = False
-                children[p-1].set_state(False)
 
-        self.axes_states[pos] = tf
-        self.logger.info('Axes states : %s ' % self.axes_states)
+        # Deactivate previously selected axis
+        if self.selected_axis is not None:
+            children[self.selected_axis-1].set_state(False)
 
-        # Clear plot if no axis is enabled
-        if True not in self.axes_states:
+        # Check if the old axis is clicked
+        if pos == self.selected_axis:
+            self.selected_axis = None
+            self.save_slit.set_enabled(False)
             self.redraw_slit('clear')
         else:
-            self.redraw_slit('redraw')
+            self.selected_axis = pos
+            children[pos-1].set_state(tf)
+            if self.cutstag != self._new_cut:
+                self.save_slit.set_enabled(True)
+            self._plot_slit()
 
     def redraw_slit(self, mode):
         if mode == 'clear':
             self.plot2.clear()
-        elif mode == 'redraw':
-            obj = self.canvas.getObjectByTag(self.cutstag)
-            line = self._getlines(obj)
-            self._plot_slit(line[0])
         elif mode == 'transpose':
             self.slit_data = self.slit_data.T
             self.plot2.ax.imshow(self.slit_data,  interpolation='nearest',
@@ -1030,8 +1019,8 @@ Keyboard shortcuts: press 'h' for a full horizontal cut and 'j' for a full verti
         old_val = self.transpose_enabled
         self.transpose_enabled = tf
 
-        if old_val != tf and self.cutstag != self._new_cut and True in self.axes_states:
-            self.redraw_slit(mode='transpose')
+        if old_val != tf and self.cutstag != self._new_cut and self.selected_axis:
+            self.redraw_slit('transpose')
 
     def set_labels(self):
         image = self.fitsimage.get_image()
--- a/ginga/misc/plugins/Cuts.py
+++ b/ginga/misc/plugins/Cuts.py
@@ -100,7 +100,8 @@ class Cuts(GingaPlugin.LocalPlugin):
         self.cuttypes = ['line', 'path', 'freepath', 'beziercurve']
         self.cuttype = 'line'
         self.save_enabled = False
-        self.axes_chkboxes = []
+        self.axis3_enabled = False
+        self.mdfile = False
 
         # For collecting data orthogonal to the cut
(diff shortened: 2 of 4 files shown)


ginga/misc/plugins/Cuts.py
```diff
@@ -319,23 +330,54 @@ def stop(self):
...
-                x1, y1 = x2, y2
-
-        points = numpy.array(points)
-        self.plot.cuts(points, xtitle="Line Index", ytitle="Pixel Value",
-                       color=color)
+
+        if self.cut_axis_enabled:
+            path = image.get('path', None)
+            hdus = pyfits.open(path, 'readonly')
+            phd = hdus[0].data
```
rupak0577: Is there any other way of accessing the data?

ejeschke: Can we assume the data is always in HDU 0?  If we are going to access the HDU directly in this plugin then maybe we need a GUI control for that.

I'm wondering if instead this data access should be abstracted behind `AstroImage` (or maybe a subclass) which keeps a reference to the full pyfits object.  Then `get_pixels_on_line()` could take an `hdu` and `axis` keyword parameters.

ginga/misc/plugins/Cuts.py
```diff
@@ -319,23 +330,54 @@ def stop(self):
...
+                        coords = image.get_pixels_on_line(int(x1), int(y1),
+                                                          int(x2), int(y2), getvalues=False)
+                        # don't repeat last point when adding next segment
+                        coords.extend(coords[:-1])
+                        x1, y1 = x2, y2
+
+                coords = np.array(coords)
+                slit = phd[:, coords[:, 1], coords[:, 0]]
+
+                plt.imshow(slit, cmap=plt.get_cmap('gray'), interpolation='nearest', origin='lower')
```
ejeschke: Is this supposed to pop up a new plot or modify the existing plot?  If the latter, the matplotlib OO interface should be used on self.plot.ax ...

rupak0577: Unfortunately, self.plot.ax.imshow(...) does not update the plot. I was wondering if the plot area could be split into two but that would scale down the plot.

ginga/misc/plugins/Cuts.py
rupak0577: Hmm, I didn't think about that. So should it be something like MultiDim's HDU list? But that would mean opening the file directly which can then be used to access the data.

ginga/misc/plugins/Cuts.py
ejeschke: I think you need do something like `self.plot.fig.canvas.draw()` to refresh the plot after the `self.plot.ax.imshow()`.  You can do a `self.plot.ax.cla()` (it's something like that) to clear the axes prior to the imshow if necessary.

rupak0577: That works, thanks! But it acts weirdly and squeezes the Cuts plot.

Cadair: if you are updating an existing plot you can do `im.set_array` (or `im.set_data` I can't remember). Which will not re-draw the whole figure, and therefore be faster. (where `im` is the `AxesImage` object returned from `imshow()`.

ginga/misc/plugins/Cuts.py
Cadair: @ejeschke It seems that at the moment the `AstroImage` object doesn't hold a pointer to the whole array in the file, just the current 'image' slice. If there was a reference to the whole array that could be used for this it would be a lot easier.

In the absence of that, I think we need to open the correct HDU using the ginga file tools.

pllim: This is a bit hacky, but I use it to open a given HDU automatically from within a plugin. This is a code snippet from a method that belongs to the plugin:
```python
class MyLocalPlugin(...):
    def redo(self):
        # stuff

        #----- Extract ERR extension -----
        # errsrc is ERR version of image
        imfile = image.metadata['path']
        imname = image.metadata['name'].split('[')[0]
        err_extnum = 2  # I have my own function to find it, but let's say it is always 2 here for simplicity
        if err_extnum < 0:
            errsrc = None
        else:
            chinfo = self.fv.get_channelInfo(chname)
            errname = '{0}[{1}]'.format(imname, err_extnum)
            if errname in chinfo.datasrc:  # ERR already loaded
                errsrc = chinfo.datasrc[errname]
            else:  # Force load ERR data
                errsrc = self.fv.load_image(imfile, idx=err_extnum)
                chinfo.datasrc[errname] = errsrc
                self.fv.make_callback('add-image', chinfo.name, errsrc)

        # more stuff
```

Cadair: What does `self.fv.load_image` return?

pllim: `AstroImage`

Cadair: @pllim That won't work here because that only holds a 2D array, we need the full 3D or 4D array that is in the file.

pllim: Is it possible to treat 3D as multiple 2D?

Cadair: possible, yes, but not practical, as we need to slice the numpy array to efficiently access the slit data.

pllim: Ah, okay.

ejeschke:
@rupak0577, @Cadair, @pllim 
Now that Cuts responds to selecting slices in MultiDim (see commit 759a84d77730fd9c0ef991ad1417f80ce64a380a), do we need the cut axis bits here?  I'm thinking that the slice control should be orthogonal and not really baked into this plugin.

We could fairly easily relocate `MultiDim` to the left side panel--I'm thinking this is beneficial for other plugins that also `redo()` when the HDU or slice changes.

ejeschke:
I think ultimately we may want to look at simply making the image object hold the entire array (as suggested earlier in this thread) and select a slice via a `set_slice()` method or some such.  I'm worried a bit about large ramp frames, etc., but maybe there can simply be an option or preference.  Then plugins that want to change the slice can simply call `set_slice()` or similar on the image.  Might simplify `MultiDim` as well.

Cadair:
@ejeschke I think having the image object hold the whole array is the best way to go from a perspective of multi dimensional data. Generally you are going to want to be able to do operations on the whole array.

ejeschke:
@rupak0577, @Cadair 
Could you give commit 85c3c767689ce166328392da7e93b8372dbd9a5c a whirl?  This opens the MultiDim plugin in the lower left workspace, which means you can navigate through the slices quite easily  and watch your cut change, you can even watch the cut change using the animation controls (although I would recommend increasing the interval slightly to 0.4 or so, matplotlib plotting seems to be the bottleneck).

I'm wondering how much of the NAXIS navigation you need from this PR if that is available?

pllim:
@Cadair , I need clarification of "holding the entire array". I work with a format like the one below and I certainly do not want `ERR` and `DQ` loaded all the time.

EXT | EXTNAME |  Description
------|---------------|-----------------
0      | PRIMARY  |  Header
1      | SCI            |  2D float array
2      | ERR          |  2D float array
3      | DQ            |  2D int array

Cadair:
@pllim I mean the whole array for the current HDU. i.e. if EXT 0 is a 4D array there should be a reference available to the full array. SO you can slice, or find maxima or averages etc.

pllim:
@Cadair , thanks, that makes sense.

ejeschke:
@nabobalis, @Cadair, can you try the latest master (see comment above)?  Try using `MultiDim` and `Cuts` together--does it do more or less what you want?  I'm open to making MultiDim open by default when NAXIS>=3.

Regarding maintaining a handle to the full cube in the image--I'm going to look at that in a separate issue.

rupak0577:
@ejeschke Cuts now responds to MultiDim so that fixes the issue.

nabobalis:
@ejeschke  Cuts now responds to multidim. 

But we want to be able to handle the full cube. We love space-time diagrams and the like in solar physics too much. This is what hopefully, this PR can address in some form.

ejeschke:
@Cadair, @nabobalis, @rupak0577 
Commit 7b8c22c9a7bc7b85d450a66bdb5a804a6711685c adds support for maintaining a handle to a multidimensional array.  See the changes to the [`set_naxis()`](https://github.com/ejeschke/ginga/blob/master/ginga/misc/plugins/MultiDim.py#L357) method in `MultiDim` for how to change the slice.  The full array can be obtained with the call to `get_mddata()` on the `AstroImage` object.

ejeschke:
Suggest this PR be reworked to reflect the changes in `MultiDim` and `Cuts`.  In particular, it strikes me that it might be better to subclass `Cuts` to add the slit view image and to better let you customize this for solar.  BTW, this is not uncommon.  We subclass a number of the plugins here and @pllim has done the same.

nabobalis:
So I take it that this isn't a common thing in astrophysics?

nabobalis:
By which I mean slits for astronomy data.

ejeschke:
@nabobalis, I'm not familiar with processing datacubes.  We have only one instrument that I know of that produces them regularly and I think there has been no demand for such a feature from them.  However, if this is useful more widely then I'm happy to have it added to `Cuts`.  Maybe @pllim can comment?

Is the slit image just a different way of slicing the cube than you get with the naxis controls in `MultiDim`?

nabobalis:
Just about, we want to slice the array such that we get a distance time graph. Basically, we take cut and extend it into time.

ejeschke:
Ok, let's just rework for the changes and see where we are.  Unless the new graphs/images are done as popups the `Cuts` gui might get a little crowded.  Another option is to use a tabbed view like in `Pick` or a scrolled view, etc.

I'm reviewing your Crispex screenshot.  Do you want to reproduce the top right-hand plot as well?

nabobalis:
Yes to that question. we have partial support for it. Isn't a spectrum view nice for astrophysics? I feel kind of bad adding solar specific features!

ejeschke:
@nabobalis, I suspect that you are right that it is very useful.  It's just that I'm not so familiar with this kind of data processing.  However, I'm game to add any features that are useful.  Just want to be able to collapse GUIs that are only sometimes used.  We use `Cuts` here mostly as a quick look for judging optical/IR spectrograph images.

ejeschke:
Non-cubes, in the usual case.

ejeschke:
Probably we could add a tabbed view for the current plot, which would allow the "spectral intensity vs. position" graph to occupy the same space.  If you are ok with it then we could easily add the slit image as yet another tab.  This would push the cuts controls down to the bottom of the panel but allow the slit image to be tall.  Or we could make the slit image a pop out that you get by pressing a button.

nabobalis:
Yeah either a tab or a pop out window once you involve a slit would be nice. As long as its got enough space.

pllim:
I currently do not use Ginga with data cubes, so I cannot comment much on this. But development wise, it makes more sense to sub-class first and get it working for your particular dataset. And then if there is enough demand from general Ginga users, incorporate that into core Ginga distribution.

pllim:
Just as in `astropy`, they have the core distribution, and they have affiliated packages for specialized functionalities.

Cadair:
I know @astrofrog and @keflavich have been doing lots of work on astro radio cubes, I wonder if they have anything to say about the general use of viewing them in ginga.

keflavich:
@Cadair - what do you want to know?  Slicing across cubes (radio or optical - MUSE and KMOS produce them in optical/NIR, ALMA/EVLA/single dish in the radio) is a very common operation, but examining a single spectrum (or an average over some region) is crucial.  Swapping the displayed axes is very helpful and supported by few tools.

Cadair:
so we are suggesting adding an extension to the cuts functionality of ginga, which currently plots the intensity of the image along a drawn line, to being able to plot an image of the values of the cut along the third axis of the array. so for a x,y,t cube, it would be a distance-time diagram, or for a x,y,lambda cube it would be a distance-wavelength graph.

It sounds like this is something that would be useful for general astro cubes? 

When you say single spectrum you mean for a single pixel?

And when you say swapping the displayed axes do you mean making the main image say x,lambda? for a given y value?

keflavich:
> It sounds like this is something that would be useful for general astro cubes?

Yes!

> When you say single spectrum you mean for a single pixel?

Yes

> And when you say swapping the displayed axes do you mean making the main image say x,lambda? for a given y value?

And, yes.

Taking a slice at off-axis angles is also useful, but computationally expensive and therefore slow.  Presumably, the `slice` interface will come with some option for moving the slice (the long-slit) and then re-extracting?  It would be nice if that was configurable, either to update on-the-fly or on some keystroke.

Cadair:
So this PR doesn't do the axes swapping, although that is something that could probably be done quite easily with the recent changes to ginga. I will create a new issue for that.

This PR does do the other two things though, single spectrum for an arbitrary line, and all spectra for an arbitrary line (as an image). I think the current `Cuts` interface to ginga allows you to move the cut and then the spectra would update once at the end of the move operation (is this correct @rupak0577 @nabobalis ?).

nabobalis:
@cadair, I think so. But it had to be checked. Also it updating with the mouse needs to be added.

Cadair:
What updating with the mouse?

nabobalis:
The spectra at a pixel. Unless it click to update on pixel?

Cadair:
I thought we were putting that in the zoom plugin?

nabobalis:
I've lost track of where that is going.

Cadair:
well we can work it out, but we want to do mouse over somehow.

ginga/misc/plugins/Cuts.py
```diff
@@ -221,6 +235,29 @@ def build_gui(self, container):
...
+            for i in xrange(0, len(image.get_mddata().shape)+1):
+                name = 'NAXIS%d' % i
+                chkbox = Widgets.CheckBox(name)
+                self.axes_chkboxes.append(Bunch.Bunch(widget=chkbox, pos=i, state=False))
+                hbox.add_widget(chkbox)
+
+            # Add callbacks
+            for chkbox in self.axes_chkboxes:
+                cbox = chkbox.get('widget')
+                cbox.add_callback('activated', lambda w, tf: self.axis_toggle_cb(w, tf, chkbox))
```
rupak0577: @ejeschke Somehow, this does not work - only the last checkbox works. Is there any other way to add the callbacks?

ejeschke: @rupak0577, are you trying to add radio button like behavior with exclusivity?  If so, please use the `Widgets.RadioButton`.  For second and remaining buttons you specify the "group" using the group parameter which refers to the first widget in the group.  Also, you *may* get a callback with `False` when the widget has toggled "off" so you want to check for True before you actually do something.  You can see examples in the `Cuts` plugin radio button processing.

rupak0577: @ejeschke I am trying to add check boxes according to the number of axes so that multiple axes can be used for slicing. What I can't figure out is how to add a callback for each checkbox.

ejeschke: If they are independent checkboxes you certainly should be able to assign separate callbacks for them.  If they don't fire when toggled then there may be a bug.  Do you get the same behavior with gtk?

rupak0577: @ejeschke The `state` of the checkboxes does not change except the last one. I think that is due to the loop that adds the callback to each checkbox. There is no problem with the GUI.

ejeschke:
@rupak0577, this PR causes the ordinary Cuts on a 2D image to fail, because it now attempts to access a 3D image in `_plotpoints()` (for the slit).

ejeschke:
Similarly, I tried Cuts on a 4D image and that also fails in the same spot.  I think the code handling slit needs to check explicitly for a 3D object before trying to slice it in 3D.

rupak0577:
@ejeschke Fixed the slicing issue. 

I was wondering if the plot could be made bigger by replacing the instructions textbox with a collapsible widget.

ejeschke:
@rupak0577, if you know of a good collapsible widget I would like to hear of it.  I think the Qt QDockWidget has some capabilities like that, but it only works in certain places, and may not work in the general case for plugins.  So I think it may be necessary to construct a custom container widget.  In any case, I have long wanted something like that for the plugins.  We would need a gtk solution as well.

ginga/misc/plugins/Cuts.py
```diff
@@ -204,6 +217,15 @@ def build_gui(self, container):
...
         vbox.add_widget(vbox2, stretch=0)
+
+        # Create axes checkboxes only if file is multidimensional
+        image = self.fitsimage.get_image()
+        if not hasattr(image, 'naxispath'):
+            pass
+        elif len(image.naxispath) > 0:
+            self.mdfile = True
+            self.build_axes(vbox)
+
```
ejeschke: `redo()` is called when a new image is loaded into the viewer, so this test needs to be performed there as well, *before* the calls to replot.

ginga/misc/plugins/Cuts.py
```diff
@@ -227,6 +249,22 @@ def build_gui(self, container):
         self.select_cut(self.cutstag)
         self.gui_up = True
 
+    def build_axes(self, vbox):
+        fr = Widgets.Frame("Axes")
+        hbox = Widgets.HBox()
+        hbox.set_border_width(4)
+        hbox.set_spacing(1)
+
```
ejeschke: Since the axes checkboxes are only used for the slit, and only useful if there is a 3D cube, I recommend moving the checkbox array into the bottom of the slit tab instead of the main window.  I suggest building the hbox holder in `build_gui` and just keeping a handle to it, so that you can just clear it and repopulate it with new checkboxes when a new image arrives.

ejeschke:
Another possibility for making more room for the slit image is to make the tabs encompass more controls.  So, e.g. the Cuts tab could include all the cuts controls.  The `Close` button always needs to be visible, but this would give ample vertical room for the slit plot to expand inside the Slit tab.

rupak0577:
@ejeschke Updated. 

GTK has the Expander Widget but Qt does need a custom one. I would like to work on it but probably after the coding period if that's okay.

Regarding clearing the axes in `redo()` - since `draw_cb()` also calls `redo()`, the checkboxes get cleared after a cut. So how do I check if a new image has been loaded?

ejeschke:
@rupak0577, this is looking better.  I still get an error in the `_plot_slit()` code when the cuts line is moved so that it falls off the edge of the data.  I think you need to basically throw out any coords that lie outside the data boundaries of the slit slice (this is what happens for plotting the regular cut--you can see that it still works if the line goes off the edge).

BTW, I think `_plot_slit()` should be called *not* from `_plotpoints()`, but as a follow-on call in `_redo()`.

Regarding `redo()`: that is the official method that is called when a "new image" enters the viewer or the image changes.  So I think maybe `draw_cb()` and `redo()` need to be restructured so that you can do the right kind of check for the data axes changing in `redo`.  I'll take a look at the master branch and maybe ask you rebase against a change to allow that.

ejeschke:
@rupak0577, commit fb27fcd3ad82364c22fbb926d4f1dc171a6522d3 refactored so that the replotting is separated from the `redo()` method.  This should allow you to rebuild the axes checkboxes when the data changes.  Can you rebase against that please.

ejeschke:
Also, I notice that you have changed the code so that you are plotting the cuts directly through the matplotlib axes instead of through the `Plot` object.  This is ok, but we've now lost the "tight_layout" call and so the axes labels are slightly occluded now.  It would be nice to fix that if possible--maybe by adding some padding into the matplotlib axes?

Cadair:
You can call `tight_layout ` on the `Figure` object if you have access to that.

ejeschke:
Yeah, figure is accessible.  `tight_layout` sometimes gives a little bit too much padding, as we've seen with the gtk version.

Cadair:
Fair enough, I think you can adjust the passing with arts to `tight_layout `

ejeschke:
@rupak0577, check out commit be1a006c2178f1f7da906ca9ee6908e1dfad3ad3.  This adds support for the Expander widget.  The Qt solution is kind of hacky, so if you want to improve it later, please do.  I've changed the Cuts plugin to use it as an example.

rupak0577:
@ejeschke Thanks! It works perfectly.

I have reverted the plotting (don't know why it wasn't working before) .

Regarding the error in `_plot_slit()`, I am not getting an error although it's true that the plotting should not work for points outside the boundaries.

rupak0577:
@ejeschke Added check for points outside image and support for bezier cut.

ginga/canvas/types/basic.py
```diff
@@ -573,11 +573,15 @@ def select_contains(self, viewer, data_x, data_y):
...
         data = image.get_data()
         wd, ht = image.get_size()
-        res = [ data[y, x] if 0 <= x < wd and 0 <= y < ht else numpy.NaN
-                for x, y in self.get_points_on_curve(image) ]
+        if getvalues:
+            res = [ data[y, x] if 0 <= x < wd and 0 <= y < ht else numpy.NaN
+                    for x, y in self.get_points_on_curve(image) ]
+        else:
+            res = [ [x, y] if 0 <= x < wd and 0 <= y < ht else numpy.NaN
+                    for x, y in self.get_points_on_curve(image) ]
```
ejeschke: Hmm... not sure if you want `NaN`s mixed in with coordinate pairs.  This would cause basic numpy array transposition to fail.  Maybe we need a "magic" bad coordinate value.  At the very least this needs to return a pair instead of `NaN`.

rupak0577: The NaNs are removed before plotting in the bezier part of `_plot_slit()`

ejeschke:
@rupak0577, I think we are pretty close on this one. @nabobalis, @Cadair, can you guys have a look and give your feedback?

I'm thinking that maybe at least the Move/Draw/Edit buttons ought to fall below the tabs (not embedded in the first tab) so that even if you are mostly using the slit tab you can easily move between moving, drawing and editing the slit line.

rupak0577:
Yeah, that would be nice. Probably the cut selection box too.

This still needs to be tested with a 4D file.

ejeschke:
If you are going to move the Cut selection box down then may as well move all the Cut controls.  If the typical use case is that a single slit line will be created and then redrawn/moved/edited it will not be necessary to access that so frequently.

Cadair:
@rupak0577 

I have just tested this with a dummy 4D file. A few comments are below.

It seems that the slit is not re-drawn when you change from NAXIS3 to NAXIS4. Also you should not be able to select both NAXIS3 and NAXIS4, if you start with NAXIS3 selected when you select NAXIS4 it should de-activate 3 automatically.

Also, the yaxis should not say "Time" unless we are going to start determining the coordinate system and type for the 3rd or 4th dimensions it should probably say nothing, unless someone has a better idea.

It would also be nice if you could transpose the plot with a checkbox, personally I would think it would be more logical to have the slit position up the y axis but that could just be me.

Cadair:
Also @rupak0577 The whole cuts control should be where the draw move and edit buttons now are, because you might want to make a new slit just like you make a new cut.

Cadair:
Also, it seems that slit does not work with the freepath mode, I just seem to be getting one pixel being repeated.

rupak0577:
@Cadair Updated.

Cadair:
@rupak0577 Nice! Thanks.

One thing, you should keep the axis label that is "position along slit" so that you know what axes is the position and which is the dependant variable.

Otherwise I think I am :+1:  on this.

rupak0577:
@Cadair Adding the axis label seems difficult since it transposes the plot by transposing the array. So the axis must be manually flipped. Also the "Slit position" axis is different for NAXIS3 and NAXIS4 (for 3 it's along x-axis, for 4 it's along y).

Cadair:
hmmm, ok.

The way I see to be able to do it is to work out which axis of the array is the same length as the number of slit coordinates and then decide on the label that way.

i.e.

```python
if len(coords) == slice.shape[0]:
    ax.set_ylabel('Position along slit.")
else:
    ax.set_xlabel("Position along slit.")
```

rupak0577:
@Cadair Thanks! It's somewhat hacky but works (only for axis 3,4 though).

This looks complete now.

ejeschke:
@nabobalis, any comment?  Otherwise I think we could merge this.

nabobalis:
@ejeschke We had a few small bugs with it the last time we used it. I'll check to see if they are still there.

rupak0577:
@nabobalis @ejeschke One remaining issue is making MultiDim work with Slit. The checkboxes are cleared and whenever the image changes. Finding out whether a new image has been loaded or a new slice seems difficult.

nabobalis:
@rupak0577 Using your latest branch for this. When you create a slit, a cut is also created. Thus when you save out the data, it saves out both sets of data. Not sure if there is a way to suppress the cut. 

With regards to the Multidim slider, I guess if the NAXIS button does not clear when you move the slider would be a quick solution to that issue. 

@ejeschke Are you aware of a way to extract a fits header for a cut? It would be nice to save out the slit data as a fits file.

ejeschke:
@rupak0577, can you rebase against master.  I think we are close to merging this.  I'll take a look at the checkboxes being cleared issue.

rupak0577:
@ejeschke Updated.

The plugin seems a little crowded now. It would also be nice to have the plot remain constant in size. So I was wondering if it would be better to have the plugins as separate floating windows?

nabobalis:
Maybe have the option of having the new plots be popped out into a new window. But that would be in the preferences. 

I assume the issue of crowding is due to peoples screen size/resolutions.

rupak0577:
@nabobalis Save and checkbox clearing issues fixed.

ejeschke:
@rupak0577, I think there are a couple of minor issues still.  

1) if I have slit enabled by choosing an axis in the Slit tab and then press "Play" in MultiDim to step through the cube rapidly at the default update rate (0.10) the displays do not seem to keep up and the viewer becomes unresponsive.  This doesn't happen in Cuts master so I think it is just due to the extra work being done each step to extract and update the slit image (many things are being updated at each step).  I think at the least we need to set the default rate to something lower.  On my oldish machine 0.3 or 0.4 rate it seems to keep up ok.   Not sure what the best way to handle this because performance of algorithms always depends on the users hardware to some degree, but I guess we don't want to provide a poor experience for average current hardware out of the box.

2) I think we need to figure out a way to keep the axis controls from being rebuilt on each step through the same cube.  After all they are not changing unless we change HDUs or load a new file.  I'll take a look at this problem.

rupak0577:
@ejeschke Strangely, it happens with me for Cut master. It works fine if there is no cut but if a cut is present, it works for a while before the viewer hangs. So I guess the problem lies with the viewer. Anyways, I will try to look for a solution.

ejeschke:
@rupak0577, sorry, what happens with Cut master?  Is it possible to consistently reproduce the viewer hang?  If so, could you open an issue and include instructions on how to produce the hang?  Thanks!

rupak0577:
@ejeschke Yeah, the viewer freezes but the cuts plot continues to update itself. Buttons are unresponsive as well.
It happens at the default rate 0.10 and the size is 928x548. Somewhat inconsistent though - it either hangs early or it takes a while.

ejeschke:
@rupak0577, what about if the rate is 0.4 or so?

rupak0577:
@ejeschke Yeah, that works fine

rupak0577:
@ejeschke @nabobalis Fixed the performance issues with MultiDim

nabobalis:
If I create a slit and move the NDim Slider, ginga freezes. It will crash for me with this
```
RuntimeError: maximum recursion depth exceeded while calling a Python object
```

My maximum recursion depth is set quite high as well.

ejeschke:
@rupak0577, I get the same behavior as @nabobalis.  If I don't create a slit (just a cut), and drag the slider (or play) things are ok.  If I enable a slit the viewer becomes very slow or non-responsive.

rupak0577:
@ejeschke Added a few fixes - basically removed rebuilding of axes checkboxes and added a check so that it draws only if MultiDim is not playing. Should work smoothly now.

@nabobalis Sorry, I can't seem to reproduce your error - the slider and the buttons work perfectly for me with no crash or freeze. I have added a few more fixes so check out the latest update and see if it works.

nabobalis:
This now works without the issues I had previously. WIth regards to speed, without benchmarks, its hard to prove. 

@rupak0577 Anything else you wanted to do this? Or you happy for it to be merged?

rupak0577:
@nabobalis @ejeschke If nothing else is required, we can merge it

nabobalis:
@ejeschke Are you happy with this currently? I think it is ready to be accepted.

ejeschke:
I'm getting the following traceback in the log:
2015-10-15 16:48:18,444 | E | Callback.py:88 (make_callback) | Error making callback 'draw-event': 'Cuts' object has no attribute 'mdim_ref'
2015-10-15 16:48:18,444 | E | Callback.py:89 (make_callback) | Traceback:
  File "/home/eric/Git/Ginga/ginga/misc/Callback.py", line 72, in make_callback

  File "/home/eric/Git/Ginga/ginga/misc/plugins/Cuts.py", line 908, in draw_cb
    return self.replot_all()

  File "/home/eric/Git/Ginga/ginga/misc/plugins/Cuts.py", line 682, in replot_all
    if self.cutstag != self._new_cut and not self.mdim_ref.is_playing():

ginga/misc/plugins/Cuts.py
```diff
@@ -399,8 +483,22 @@ def redo(self):
         """This is called when a new image arrives or the data in the
         existing image changes.
         """
+        # Get reference to MultiDim
+        chname = self.fv.get_channelName(self.fitsimage)
+        chinfo = self.fv.get_channelInfo(chname)
+        self.mdim_ref = chinfo.opmon.getPlugin('MultiDim')
+
```
ejeschke: I don't think `mdim_ref` is getting set if you open `Cuts` *after* loading an image.  Then it raises an exception (I've marked it lower down) because it is not defined.

In general, I think the approach of trying to check what another plugin is doing and work around that is probably not the correct approach.

ginga/misc/plugins/Cuts.py
```diff
@@ -514,10 +677,10 @@ def replot_all(self):
...
             #text.color = color
             self._replot(lines, colors)
-            if n:
-                self.w.delete_cut.set_enabled(True)
-                self.w.delete_all.set_enabled(True)
-                self.save_btn.set_enabled(True)
+
+        # Redraw slit image for selected cut
+        if self.cutstag != self._new_cut and not self.mdim_ref.is_playing():
+            self.redraw_slit('redraw')
```
ejeschke: Here `mdim_ref` is not defined in the case I described above (open Cuts plugin after opening the file), so I get an exception raised and the Slit is not redrawn.

ejeschke:
I'm thinking that for playback, it may be better to support this via a fix to the viewer *not* to make the `image-set` callback which triggers all of the plugins, especially ones that plot things with matplotlib, because those tend to slow everything down.  Then playback would support a pretty high rate (I have a test program that basically can watch a HD video in a ginga widget at 30FPS).  You could still use the buttons to step through with the callback applied.

rupak0577:
@ejeschke  Would that be a fix to `ImageView`? I think `imshow` is the bottleneck and I have been trying to use `set_data` instead but it does not seem to work.

rupak0577:
@ejeschke @nabobalis Removed Multidim check. Ready to be merged(ignoring speed issues).

ejeschke:
@rupak0577, how about using a Ginga widget for the slit image?  It will certainly be faster than matplotlib.  Take a look at `Pick` if you need an example of how to create a small ginga widget in a plugin.

ejeschke:
Let me know if you want help.

rupak0577:
@ejeschke How can it be used to set the image data(instead of `imshow`)? Also, how can I add the axis labels?

ejeschke:
Hmm, I see your point.  Plus we'd have to make sure the "save" bits are all supported.  Perhaps a future upgrade of the plugin if we get some basic plot axes for ginga.

rupak0577:
@ejeschke It isn't the perfect solution but it would work if we plot depending upon the tab the user is in - if the current tab selected is Cuts then plot otherwise don't. But the best way is to change the data instead of calling `imshow` each time.

ejeschke:
I've fixed up some problems with this PR in commit 376bf76ebeabb3235b41bb0d83602b11f26fe573.  I'm still seeing plenty of problems interoperating with 2D images, so I've made the Slit functionality a preference (default disabled) for now.  Just add `enable_slit = True` to your `$HOME/.ginga/plugin_Cuts.cfg` file.  I'll start an issue with the problems I'm seeing.

Cadair:
@ejeschke did you open those issues?

ejeschke:
Sorry, not yet.  I'll make a note to do it.  You can see what kinds of errors you get by enabling the preference and just loading a 2D image.  Ideally, the slit features should just silently bypass if they don't work for 2D, not generate error messages or worse, cause the Cuts plot to fail.

PR #145, opened 2015-07-31
Intensity Scaling support

rupak0577:
Adds support for MultiDim to scale w.r.t first slice.

### DIFF
--- a/examples/gw/shared_canvas.py
+++ b/examples/gw/shared_canvas.py
@@ -1,10 +1,5 @@
 #! /usr/bin/env python
 #
-# example2.py -- Simple, configurable FITS viewer.
-#
-# Eric Jeschke (eric@naoj.org)
-#
-# Copyright (c)  Eric R. Jeschke.  All rights reserved.
 # This is open-source software licensed under a BSD license.
 # Please see the file LICENSE.txt for details.
 #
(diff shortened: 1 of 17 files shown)


ejeschke:
@rupak0577, this looks like it has conflicts against current master.  Can you rebase it?  We should try to polish off the PRs this week.

rupak0577:
@ejeschke Rebased. 

I would like to know if the scaling is correct.

ginga/misc/plugins/MultiDim.py
```diff
@@ -410,7 +418,15 @@ def set_naxis(self, idx, n):
             self.naxispath[m] = idx
             self.logger.debug("m=%d naxispath=%s" % (m, str(self.naxispath)))
 
-            image.set_naxispath(self.naxispath)
+            if self.scaling_enabled:
+                curr_max = np.amax(image.get_mddata()[idx])
+                curr_min = np.amin(image.get_mddata()[idx])
+
```
ejeschke: `amax` and `amin` will propagate NaNs.  I think it may be better to use `nanmax` and `nanmin`.

ejeschke:
@rupak0577, I think the scaling looks correct, but I'm not sure if this is the best approach.

It seems like you could get the same effect by simply setting the cut levels appropriately when a slice is loaded--since the cut levels do pretty much what you are doing here anyway.  That could be done by calling `cut_levels` on the viewer after loading a slice. BUT...

Another way this might work that could be more elegant is as a new entry in `ginga.AutoCuts`.  You simply need to have a reference to the first slice and it would calculate the new cuts based on the differences from the reference slice.  So then this is not a hack on `MultiDim` but simply a new kind of autocuts that you select in Preferences.

rupak0577:
@ejeschke I have created a new autocut which returns the min and max of the first slice from its `calc_cut_levels()` but it doesn't seem to work.

ejeschke:
@rupak0577, do you have a PR against `ginga.AutoCuts`?

rupak0577:
@ejeschke I have updated it with the new autocut.

ginga/AutoCuts.py
```diff
@@ -628,6 +628,20 @@ def fitting(x, slope, intercept):
...
+    def __init__(self, logger):
+        super(Scaling, self).__init__(logger)
+        self.kind = 'scaling'
+
+    def calc_cut_levels(self, image):
+        # Get max and min of first slice
+        first = image.get_mddata()[0]
+        loval = numpy.nanmax(first)
+        hival = numpy.nanmin(first)
+
```
ejeschke: `nanmin` and `nanmax` are reversed here.

But I don't think this is the correct implementation.  What this does is to apply the min/max values from the first slice to all slices.  Is that actually what is requested?

ejeschke:
I'm trying to fully grasp the use case for scaling the actual data vs. just adjusting the cut levels so that the image intensity is *visually* scaled to the same range as the first slice.  If the purpose is the visual adjustment only then it is better to pursue a solution based on setting cut levels.  The alternative, scaling the data itself, means that some of the data analysis functions (such as `Cuts`, `Info`, etc.) will not be showing the correct values for the actual data in the file (except for the first slice), because the data itself was scaled (things indexing the new `mddata` will be ok, of course).

@nabobalis, @Cadair, @rupak0577  can you try commit d0e6b9680f17d0c7ae2709b8b09323ddb3c3926b in the main repo?  This adds support for autocuts when moving through slices with `MultiDim`. In the `Preferences`, make sure that "Cut New" (down in `New Images`) is set to "on".  As you scroll through the slices, try changing the preference for "Auto Method" (in the `Auto Cuts` section--start with "minmax").  Does this give you what you want or do you want to actually scale the data itself, with its effect on the Cuts, Info, etc. ?

rupak0577:
@ejeschke Yeah, the purpose is to visually scale to the first slice.

nabobalis:
@ejeschke The idea here was that you could normalize the intensity images from a chosen image. So you can see changes in intensity from a reference frame. Ideally, that frame does not have to be the first one.

ejeschke:
I see.  In that case I think we definitely do not want to be munging the actual data, but simply using the cut levels in way that makes sense.  For example, if you want to see intensity changes from a reference frame, why not (in `Preferences`) turn `Cut New` to "off", and `Auto Method` to "minmax".  Then go to your reference frame, press the "Auto Levels" button in `Info`--this will scale the visual image to the min and max.  Then simply step through the slices with the slider.  Since cut levels are fixed at the first frame min and max values, intensity changes should be easily seen.

Note: I fixed a redraw bug in commit 0d9f8b10c1c5cd59d90be0e82884bf1754c6fb4f where the canvas was not redrawn properly if the naxis was changed and auto cut levels was turned off.  Please test with this commit or later.

rupak0577:
@ejeschke That works as required, I think. @nabobalis Can you please check it?

If it is correct, probably there should be a button (in `MultiDim` or `Synopsis`) for quick setup.

ejeschke:
I should note that if you save the preferences for the "Image" channel they will be inherited for any other channels you create, and will be loaded when you start Ginga. So then it simply becomes a matter of pressing "Auto Levels" for the reference image and stepping through the rest--pretty simple.

nabobalis:
@ejeschke That works. I had no idea that was possible. With that case, this can be closed. It doesn't need to be added.

Issue #92, opened 2015-03-25
Ginga plugins as a namespace package?


pllim:
In my experience so far in developing custom plugins for my users, I found 3 ways I could distribute them, all of which have their own pitfalls:

1. Have my program always copy the plugins over to the user's `$HOME/.ginga/plugins` directory every time it runs.
2. Keep reminding my users to manually run a special script that does the copying as needed before starting my program.
3. Forcefully copy my plugins into `ginga/misc/plugins` installation directory via my `setup.py`.

I asked my colleagues for a more elegant solution. @embray mentioned the following, which I thought you might be interested to consider:

> What Eric might consider doing is making the "ginga" package into a namespace package.
> Or in particular he might consider making a "ginga.plugins" namespace package.
> This is basically the Official Pythonic way of installing multiple optional subpackages into the
> same namespace.
>
> If for whatever reason he doesn't want to make the "ginga" package itself into a
> namespace package, he could still establish an official "ginga_plugins"
> namespace which *would* be a namespace package, and that all Ginga plugins could
> be required to belong to.
>
> Another option which is used by a few projects and isn't *too* bad would be to
> use setuptools entry points
> (https://pythonhosted.org/setuptools/setuptools.html#dynamic-discovery-of-services-and-plugins)
> to discover installed plugins.

What do you think on this matter? Thank you.

ejeschke:
I understand the issue.  I'll take a look at the namespace feature.

ejeschke:
I think the namespace approach sounds feasible.  I suppose ginga.misc.plugins, ginga.qtw.plugins and ginga.gtkw.plugins all should be exported as namespaces.  Any quick pointers on how to implement that?

pllim:
@embray, perhaps you are the best person to answer this?

pllim:
@embray wrote this on an internal webpage, so I copy-and-pasted it here for reference. What do you think, @ejeschke ? Is this feasible?

**Python Namespace Packages**

[...]

PR #146, opened 2015-07-31
Line Profile plugin

rupak0577:
Adds support for spectral analysis.

@ejeschke Is there a way to access MultiDim for the current header and data from this plugin?

### DIFF
--- a/ginga/misc/plugins/LineProfile.py
+++ b/ginga/misc/plugins/LineProfile.py
@@ -15,14 +15,12 @@ class LineProfile(GingaPlugin.LocalPlugin):
         self.pan2mark = False
         self.wd = None
         self.ht = None
+        self.selected_axis = None
 
         self.dc = self.fv.getDrawClasses()
         canvas = self.dc.DrawingCanvas()
-        canvas.enable_draw(True)
-        canvas.add_draw_mode('move', move=self.drag, up=self.update, down=self.btndown_cb)
-        canvas.register_for_cursor_drawing(self.fitsimage)
-        canvas.set_draw_mode('move')
-        canvas.add_callback('motion', self.motion_cb)
+        canvas.set_callback('cursor-down', self.btndown_cb)
+        canvas.set_callback('cursor-up', self.update)
         canvas.setSurface(self.fitsimage)
         self.canvas = canvas
 
@@ -37,6 +35,9 @@ class LineProfile(GingaPlugin.LocalPlugin):
         self.tw = None
         self.mark_data_x = [None]
         self.mark_data_y = [None]
+
+        # register for new image notification in this channel
+        fitsimage.set_callback('image-set', self.new_image_cb)
 
         self.gui_up = False
 
@@ -140,19 +141,26 @@ class LineProfile(GingaPlugin.LocalPlugin):
         container.add_widget(top, stretch=1)
         self.gui_up = True
 
+    def new_image_cb(self, fitsimage, image):
+        if fitsimage != self.fitsimage:
+            # Focus is not our channel-->not an event for us
+            return False
+        elif image is None:
+            return False
+        else:
+            self.build_axes()
+
     def build_axes(self):
         self.hbox_axes.remove_all()
+        self.selected_axis = None
+        self.clear_plot()
+
         image = self.fitsimage.get_image()
         if image is not None:
-            self.axes_states = []
-            # For easier mapping of indices with the axes
-            self.axes_states.append(None)
-
             # Add Checkbox widgets
-            for i in xrange(1, len(image.get_mddata().shape)+1):
-                name = 'NAXIS%d' % i
-                chkbox = Widgets.CheckBox(name)
-                self.axes_states.append(False)
+            # `image.naxispath` returns only mdim axes
+            for i in xrange(1, len(image.naxispath)+3):
+                chkbox = Widgets.CheckBox('NAXIS%d' % i)
                 self.hbox_axes.add_widget(chkbox)
 
                 # Add callback
@@ -162,22 +170,19 @@ class LineProfile(GingaPlugin.LocalPlugin):
         chkbox.add_callback('activated', lambda w, tf: self.axis_toggle_cb(w, tf, pos))
 
     def axis_toggle_cb(self, w, tf, pos):
-        # Deactivate other checkboxes
         children = self.hbox_axes.get_children()
-        for p, val in enumerate(self.axes_states):
-            if val is None:
-                continue
-            elif val is True:
-                self.axes_states[p] = False
-                children[p-1].set_state(False)
 
-        self.axes_states[pos] = tf
-        self.logger.info('Axes states : %s ' % self.axes_states)
+        # Deactivate previously selected axis
+        if self.selected_axis is not None:
+            children[self.selected_axis-1].set_state(False)
 
-        # Clear plot if no axis is enabled
-        if True not in self.axes_states:
+        # Check if the old axis has been clicked
+        if pos == self.selected_axis:
+            self.selected_axis = None
             self.clear_plot()
         else:
+            self.selected_axis = pos
+            children[pos-1].set_state(tf)
             self.redraw_mark()
 
     def instructions(self):
@@ -224,69 +229,37 @@ Use MultiDim to change step values of axes.""")
             return
         self.wd, self.ht = self.image.get_size()
 
-        try:
-            curr_axis = self.axes_states
-            self.build_axes()
+        self.redraw_mark()
 
-            # Restore axis state
-            children = self.hbox_axes.get_children()
-            for p, val in enumerate(curr_axis):
-                if val is True:
-                    children[p-1].set_state(True)
-
-            if len(self.marks) > 1:
-                self.del_all_btn.set_enabled(True)
-        except AttributeError:
-            self.build_axes()
-
-    def motion_cb(self, canvas, button, data_x, data_y):
-        if self.mark_selected is None:
-            if not 0 <= data_x < self.wd or not 0 <= data_y < self.ht:
-                self.clear_plot()
-            else:
-                self._plot(data_x, data_y, mark=None)
-        return False
-
-    def _plot(self, data_x, data_y, mark=None):
-        # Exclude points outside boundaries
-        if not 0 <= data_x < self.wd or not 0 <= data_y < self.ht:
-            self.clear_plot()
-            return
-
+    def _plot(self, mark=None):
         # Transpose array for easier slicing
         mddata = self.image.get_mddata().T
         naxes = mddata.ndim
 
-        self.enabled_axes = [pos for pos, val in enumerate(self.axes_states) if val is True]
-
-        if self.enabled_axes:
-            axis_data = self.get_axis(self.enabled_axes[0])
-            axes_slice = self._slice(naxes, data_x, data_y, mk=mark)
+        if self.selected_axis:
+            axis_data = self.get_axis(self.selected_axis)
+            slice_obj = self._slice(naxes, mk=mark)
 
             self.clear_plot()
-            self.plot.plot(axis_data, mddata[axes_slice])
+            self.plot.plot(axis_data, mddata[slice_obj])
 
-    def _slice(self, naxes, data_x, data_y, mk):
+    def _slice(self, naxes, mk):
         # Build N-dim slice
-        axes_slice = [0] * naxes
+        slice_obj = [0] * naxes
 
         # For axes 1 and 2
         if mk is not None:
-            axes_slice[0] = self.mark_data_x[mk]
-            axes_slice[1] = self.mark_data_y[mk]
-        else:
-            axes_slice[0] = data_x
-            axes_slice[1] = data_y
+            slice_obj[0] = self.mark_data_x[mk]
+            slice_obj[1] = self.mark_data_y[mk]
 
         # For axis > 3
         for i in xrange(2, naxes):
-            axes_slice[i] = self.image.revnaxis[i-2] + 1
+            slice_obj[i] = self.image.revnaxis[i-2] + 1
 
-        # Slice enabled axis
-        for ea in self.enabled_axes:
-            axes_slice[ea-1] = slice(None, None, None)
+        # Slice selected axis
+        slice_obj[self.selected_axis-1] = slice(None, None, None)
 
-        return axes_slice
+        return slice_obj
 
     def get_axis(self, i):
         try:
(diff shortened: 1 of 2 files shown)


Cadair:
So this at the moment is limited to line graphs?

Have you put any thought into how this would work with > 3D cubes. In crispex, you can have the same functionality as this, but for 4D cubes, so the line profile becomes an image.

My thoughts on this was an interface along the lines of a set of check boxes for all the axes not plotted in the main image, so that you can select one or two of them to create a line plot or a image plot.

rupak0577:
@Cadair If I am correct, the line profile for 4D cubes is along time so I think it would be somewhat like the slit though I am not sure how it works.

Regarding the interface, adding the check boxes can be done. It probably needs a pop-out window for the 4D image since it might get crowded.

I am not sure whether this should be a separate plugin or should be added to Zoom.

ejeschke:
> Is there a way to access MultiDim for the current header and data from this plugin?

@rupak0577, you can access another plugin via the following sort of code:
```python
# Get handle to MultiDim plugin for this channel
chinfo = self.fv.get_channelInfo(chname)
pl_obj = chinfo.opmon.getPlugin('MultiDim')

# Get image being shown
if not pl_obj.gui_up:
    self.fv.show_error("MultiDim has not yet been started--no data available")
    return   #??
image = pl_obj.image
data = image.get_data()
header = image.get_header()
```

ginga/misc/plugins/LineProfile.py
```diff
@@ -0,0 +1,109 @@
...
+        fits_f = pyfits.open(path, 'readonly')
+
+        self.lp_plot(fits_f)
+
+    def lp_plot(self, fits_f):
+        wav = self.get_axis(fits_f[7].header)
+        self.plot.plot(wav, fits_f[7].data[156,153,:])
+        title = fits_f[0].header.get('TDESC7') + ' ' + str(fits_f[0].header.get('TWMIN7')) + '-' + str(fits_f[0].header.get('TWMAX7')) + ' Ang'
+        self.plot.set_titles(xtitle='Wavelength', title=title)
+
```
ejeschke: @rupak0577, This plugin looks very tightly tied to a particular kind of FITS file.  Nothing wrong with that, but I think in the current form it may be better as a custom plugin.  I think this may be some additional impetus to solve issue #92, so that such plugins can be distributed separately and installed.

Maybe what we could do in the mean time is add support for auto-discovering all the plugins in a user's $HOME/.ginga/plugins folder and loading them.  So that you don't need to use the `--plugins` parameter to load custom plugins if they are in that folder.

nabobalis: Yeah, the line profile position has also got to be selectable.

Cadair: I think there is still a fair bit to do here, it should be made to be general to any >2D file though.

rupak0577:
@ejeschke @Cadair @nabobalis I am rewriting this again so haven't pushed it. Will do it soon.

nabobalis:
Try not to PR items until you think they are nearly ready.

rupak0577:
@nabobalis @ejeschke I have re-written it with mouse-over support and axis selection.

ginga/misc/plugins/LineProfile.py
```diff
@@ -0,0 +1,186 @@
...
+from ginga.misc import Widgets, Plot
+
+import numpy as np
+
+class LineProfile(GingaPlugin.LocalPlugin):
+
+    def __init__(self, fv, fitsimage):
+        super(LineProfile, self).__init__(fv, fitsimage)
+
+        self.layertag = 'lineprofile-canvas'
```
ejeschke: You don't need a this kind of variable unless you are creating a canvas that gets inserted and removed over the channel widget.

ginga/misc/plugins/LineProfile.py
```diff
@@ -0,0 +1,186 @@
...
+               np.arange(0, self.header.get('NAXIS%d' % self.wave_index), 1) * \
+               self.header.get('CDELT%d' % self.wave_index)
+        return axis
+
+    def axis_toggle_cb(self, w, tf, pos):
+        self.axes_states[pos] = tf
+        self.logger.info('Axes states : %s ' % self.axes_states)
+
+    def __str__(self):
+        return 'lineprofile'
```
ejeschke: Needs a newline at the end

ginga/misc/plugins/LineProfile.py
```diff
@@ -0,0 +1,186 @@
...
+        chname = self.fv.get_channelName(self.fitsimage)
+        chinfo = self.fv.get_channelInfo(chname)
+        self.mddim = chinfo.opmon.getPlugin('MultiDim')
+
+        if not self.mddim.gui_up:
+            self.fv.start_local_plugin(chname, str('multidim'), None)
+
+        # Get image being shown
+        self.image = self.mddim.image
+
```
ejeschke: Not quite understanding why the MultiDim plugin needs to be started to use this one.  Can't you just get the image using:
```python
image = self.fitsimage.get_image()
```
?

ejeschke:
@rupak0577, this looks better!  I think it may be worth adding a canvas over the channel image and registering for motion on that.  When you start and stop the plugin the canvas is inserted or deleted.  What this does is prevent you from getting motion event callbacks when your GUI is not up--and since the point of the plugin is to plot the area under the mouse it doesn't make sense to call the motion callback when the GUI is not up.  I'm sending you an edited version of the plugin to show you how to do it.  It's a common idiom in most of the plugins.

ginga/misc/plugins/LineProfile.py
rupak0577: Hmm, I thought it would be required for the retrieving the current header but it does seem unnecessary.

ejeschke:
Could someone send me a sample FITS file that this can be used with?  I like to have examples of everything for regression testing.

ginga/misc/plugins/LineProfile.py
ejeschke: Yeah, not necessary for accessing the header.  If it's not needed I would remove all that.

Cadair:
@ejeschke I think @nabobalis will try and make a 4D science image to test with, but I used the following to make a test image.

```python
import numpy as np
import matplotlib.cbook
from astropy.io import fits
from astropy.modeling import models

array = plt.imread(matplotlib.cbook.get_sample_data('grace_hopper.jpg'))[...,0].T
lin = models.Linear1D()
gau = models.Gaussian1D()
ll = lin(np.linspace(1,10,30))
gg = gau(np.linspace(-1,1,15))
a3d = array[None, ...] * gg[:, None,None]
a4d = a3d[None, ...] * ll[:, None, None, None]

fits.writeto('test_4d.fits',a4d, clobber=True)
```

Cadair:
@rupak0577 This does not seem to be working for me, I pick and axis and nothing happens when I mouse over or click... Am I doing something wrong?

ejeschke:
@Cadair, it's checking the header for a `CTYPE1` (2, 3, 4, etc.) that has a value of "WAVE"--if it doesn't see that it doesn't plot anything.  You'll have to advise if this is correct behavior or not...

Cadair:
I would have thought the best way to do it would be to plot a profile along a user specified axis. Maybe defaulting to WAVE if present.

rupak0577:
@Cadair Updated so that it plots the user selected axis.

ginga/misc/plugins/LineProfile.py
```diff
@@ -0,0 +1,220 @@
...
+        # Get image being shown
+        self.image = self.fitsimage.get_image()
+
+    def motion_cb(self, canvas, button, data_x, data_y):
+        self.xcoord = int(data_x)
+        self.ycoord = int(data_y)
+
+        # Exclude points outside boundaries
+        shape = self.image.shape
+        if 0 <= self.xcoord < shape[1] and 0 <= self.ycoord < shape[0]:
```
ejeschke: Recommend:
```python
wd, ht = self.image.get_size()
if 0 <= self.xcoord < wd and 0 <= self.ycoord < ht:
```

ginga/misc/plugins/LineProfile.py
```diff
@@ -0,0 +1,220 @@
...
+    def _slice(self, naxes):
+        header = self.image.get_header()
+
+        # Build N-dim slice
+        axes_slice = [0] * naxes
+
+        t_step = self.pl_obj.play_idx
+
+        # Build base slice
+        ctypes = [header.get('CTYPE%d' % i) for i in xrange(1, naxes+1)]
```
ejeschke: Recommend:
```python
ctypes = [header.get('CTYPE%d' % i, '') for i in xrange(1, naxes+1)]
```
in the case that `CTYPE` keywords are not present.

ginga/misc/plugins/LineProfile.py
```diff
@@ -0,0 +1,220 @@
...
+            axes_slice[ea-1] = slice(None, None, None)
+
+        return axes_slice
+
+    def get_axis(self, i):
+        header = self.image.get_header()
+        axis = header.get('CRVAL%d' % i) + \
+               np.arange(0, header.get('NAXIS%d' % i), 1) * \
+               header.get('CDELT%d' % i)
+        return axis
```
ejeschke: Need to handle the case that these keywords may not be present.

ginga/misc/plugins/LineProfile.py
```diff
@@ -0,0 +1,220 @@
...
+        self.plot.clear()
+        self.plot.plot(axis_data, mddata[axes_slice])
+
+    def _slice(self, naxes):
+        header = self.image.get_header()
+
+        # Build N-dim slice
+        axes_slice = [0] * naxes
+
+        t_step = self.pl_obj.play_idx
```
ejeschke: Suggest:
```python
t_step = self.image.revnaxis[-1] + 1
```
and remove stuff above having to do with starting `MultiDim`.

nabobalis:
@rupak0577 I would do @ejeschke  suggestions. The LP plugin works but if it needs Multidim open, it should open it at startup.

LP does work but I don't think the output is correct. If I use an IRIS raster, the naxis buttons produce an output but it seems to be indexing wrongly. 

Also LP needs to always have [X,Y]. To my mind, a line profile is at two spatial co-ordinates.

Cadair:
@nabobalis what do you mean by has to have X,Y?

nabobalis:
@Cadair A line profile really cant be in [x,t], it would make little sense. To me, its the how that pixel value varies in time or wavelength.

Cadair:
If you mean what I think you mean, if you had a [x, t, lambda] cube you could still have a line profile for a given position in [x,t] space. Take for instance a sit and stare raster from EIS or IRIS.

nabobalis:
Hmm, true. I had not considered that.

Ok, so ignoring that. Using LP in an IRIS raster, does not give much sensible results. Also, it seems to work in a small region of the data.

rupak0577:
@nabobalis Updated the slicing along with a few fixes. Will add the marking support soon.

rupak0577:
@nabobalis @ejeschke Added marking support. Needs a review.

nabobalis:
I like the mark, but the ability to move it once its been set, would be nice.

nabobalis:
It looks good. However, I have a two issues.

First one, I think it is sensible to disable (if possible) that you can use this with the mark set as None.
This has caused Ginga to segmentation fault twice. One on a large and small fits file. Since I just move the mouse cursor around the image and it keep updating the plugin without stopping. It freezes the rest of the GUI. 

Secondly, is there a keyboard shortcut to add another mark? If so, it should be added to the instructions at the top for the plugin.

rupak0577:
@nabobalis Updated

There is no keyboard shortcut but is it really required?

nabobalis:
@rupak0577 Keyboard shortcut would be nice but its not not necessary right now.

rupak0577:
@nabobalis That's strange - try deleting and checking it out again.

nabobalis:
@rupak0577 Even with a clean clone of your repo and checkout to this branch. There is still no LP plugin. 

@Cadair Can you reproduce this or is something wrong on my end?

Cadair:
@nabobalis I can't reproduce, the plugin opens for me ok. I haven't got data to hand to test it at the moment.

nabobalis:
@Cadair Thanks. Some strange local issue.

ejeschke:
@nabobalis, can you point me to a link to a FITS file that I can test with?  I'd like to try to reproduce the segfault and check all functionality.  I really don't want to release anything that could potentially cause a crash.

nabobalis:
@ejeschke I do not know why it segfaulted. But I assume that the loop since it was updating at every little mouse update, by moving the mouse everywhere quickly meant that even if you moved the mouse off the image, the LP would still update since it has not been able to keep up with the mouse changing so rapidly. 

With regards to a fits file. [This one should do.](http://www.lmsal.com/solarsoft/irisa/data/level2_compressed/2015/10/09/20151009_041433_3620263378/iris_l2_20151009_041433_3620263378_SJI_1330_t000.fits.gz)

Cadair:
It would be good to have the option to turn on "live" mouse updates like crispex can?

nabobalis:
That would also work instead. Add it to the preferences.

ejeschke:
I don't see any major problems with this one except that it suffers from the same problem (in PR #129) when you press "Play" in `MultiDim` it locks up.  But I think solving that problem in #129 will also fix it here, and that is a `MultiDim` issue.

ejeschke:
@Cadair, when you mention the "live" updates, does this mean just updating the plot as you move the cursor around the image? @rupak0577, do you want to try to add that before we merge or leave it for after?

nabobalis:
It was in originally. But that caused the segfault issue i had. The plugin
could not keep up with my mouse changing.

Cadair:
Yeah, updates as the cursor moves is what i meant. I don't consider it a blocker though.

ejeschke:
@rupak0577, it looks like when you are moving the mark, it ends up being 5 pixels to the left of where you clicked.  I think this is because you are creating a compound object with the text label 10 pixels to the right of the point, so when you move it the center of the compound object is in the middle.  You can see this if you zoom in greatly so the pixels are exaggerated.  See my simple fix shown in the code commentary.

ginga/misc/plugins/LineProfile.py
```diff
@@ -0,0 +1,425 @@
...
+            self.del_all_btn.set_enabled(True)
+        return True
+
+    def update(self, canvas, event, data_x, data_y):
+        tag = self.mark_selected
+        if tag is None:
+            return
+        idx = int(tag.strip('mark'))
+        obj = self.canvas.getObjectByTag(tag)
+        obj.move_to(data_x, data_y)
```
ejeschke: You probably want to change this to:
```python
obj.move_to(data_x+5, data_y)
```
because you are moving the compound object *center*.

ejeschke:
@Cadair, wish matplotlib were a little faster for plotting. I think this is the bottleneck.

Cadair:
@ejeschke If I can find some time I will look at drilling down to speed it up, I have had some success with this in the past.

rupak0577:
@ejeschke @nabobalis Fixed above mentioned issue. Ready to be merged (ignoring the speed issues)

ejeschke:
Looks good.  When you have time @rupak0577, consider adding a class docstring for the plugin (like is done for `Cuts`).  I know most of the older plugins don't have it yet, but I have in mind a kind of future help utility that will pull out the docstring.

pllim:
I briefly tried this with your test data... and my own regular 2D stuff. From the instructions alone, it is not clear that this plugin only works with data cubes. Some kind of indication that it is not supposed to be used with 2D images (e.g., greyed out boxes or status message) would  be nice.

pllim:
Also, I see this in Ginga's log when I load 2D image:
```python
2015-10-26 12:48:03,031 | E | Callback.py:89 (make_callback) | Traceback:
  File ".../ginga-2.5.20151022033143-py2.7.egg/ginga/misc/Callback.py", line 72, in make_callback

  File ".../ginga-2.5.20151022033143-py2.7.egg/ginga/misc/plugins/LineProfile.py", line 151, in new_image_cb
    self.build_axes()

  File ".../ginga-2.5.20151022033143-py2.7.egg/ginga/misc/plugins/LineProfile.py", line 154, in build_axes
    self.hbox_axes.remove_all()
```

rupak0577:
@pllim Looks like it does not work with 2D images. Thanks, I will look into it.

PR #149, opened 2015-08-16
Add extension only if path is valid

rupak0577:
Fixed a bug in the Qt Save dialog code where the extension is added to the file name even if the user has pressed 'Cancel' which creates hidden files instead of dropping the operation.


ejeschke:
Tnx

Issue #128, opened 2015-06-25
Feature Request: Annulus shape


pllim:
As discussed in #89 , I am trying to use `arr_masked = image.cutout_shape(shape_obj)` but to extract a region defined by an annulus. Since there is no direct support for this shape, I use two circles:
```python
icirc_masked = image.cutout_shape(obj.objects[0])  # inner circle
ocirc_masked = image.cutout_shape(obj.objects[1])  # outer circle
data = icirc_masked[~icirc_masked.mask & ~ocirc_masked.mask].data
```
But this doesn't work because they have bounding boxes with different shapes:
```
2015-06-25 11:51:56,856 | E | Callback.py:88 (make_callback) | Error making callback 'draw-event': operands could not be broadcast together with shapes (200,199) (220,219) 
```
I thought maybe I could write my own shape, but looking at `DrawingMixin.py` and `CanvasObject.py`, adding a new shape does not seem trivial.

pllim:
Alternately, instead of a new shape object, if there is a way to just return the full-size mask for each of the circle, I can build the annulus mask myself.

pllim:
Just so you know, I already have my own functions to extract annulus using your circle objects. So if you don't think such a support would benefit Ginga in general, feel free to close this issue.

ejeschke:
@pllim, I think there is support for this via the `CompoundObject` canvas object.  You just need to define an annulus shape by combining the two circles:
```python
CompoundObject = viewer.getDrawClass('compoundobject')
annulus = CompoundObject([circ1, circ2])
# hack to change the default interpretation of containment from OR to AND
annulus._contains_reduce = numpy.logical_and
```
(you can also just import CompoundObject from `ginga.canvas.CanvasObject`).
```

This is largely uncharted waters.  Let me know how it goes!

**Comment by ejeschke** [comment]
Sorry, this won't work  because the AND containment test is not right.  So we have to write a custom containment test.  I'll work up an example.

**Comment by ejeschke** [comment]
Tested this a little bit, and I think it should work:
```python
import numpy
from ginga.canvas.CanvasObject import CompoundObject

class Annulus(CompoundObject):

    def contains(self, x, y):
        circ1, circ2 = self.objects
        return circ1.contains(x, y) and numpy.logical_not(circ2.contains(x, y))
```
You need to place the object on the canvas (using ``add()``) before the `contains()` method will work (because it will get initialized with a coordinate mapper).

ejeschke:
[...]

PR #153, opened 2015-08-18
New Annulus shape object

pllim:
Added a new shape object that is a special compound object to handle an annulus shape that consists of two circles with the same centroid. This was discussed in detail in #128 . The methods are sufficient for my needs, but people can always expand it as needed in the future.

Example usage:
```python
tag = canvas.add(Annulus(Circle(100, 200, 10), Circle(100, 200, 15)))
obj = canvas.getObjectByTag(tag)
arr_masked = image.cutout_shape(obj)
```

### DIFF
--- a/ginga/canvas/types/basic.py
+++ b/ginga/canvas/types/basic.py
@@ -159,6 +159,12 @@ class PolygonMixin(object):
         # See: http://alienryderflex.com/polygon/
         xa, ya = x_arr, y_arr
 
+        # promote input arrays dimension cardinality, if necessary
+        if len(xa.shape) == 1:
+            xa = xa.reshape(1, -1)
+        if len(ya.shape) == 1:
+            ya = ya.reshape(-1, 1)
+
         result = numpy.empty((ya.size, xa.size), dtype=numpy.bool)
         result.fill(False)
 
@@ -169,12 +175,14 @@ class PolygonMixin(object):
                 numpy.logical_or(numpy.logical_and(yi < ya, yj >= ya),
                                  numpy.logical_and(yj < ya, yi >= ya)),
                 numpy.logical_or(xi <= xa, xj <= xa))
-            # TODO: get a divide by zero here for some elements whose tf=False
+            # NOTE: get a divide by zero here for some elements whose tf=False
             # Need to figure out a way to conditionally do those w/tf=True
+            # Till then we use the warnings module to suppress the warning.
             with warnings.catch_warnings():
                 warnings.simplefilter('ignore', RuntimeWarning)
                 cross = ((xi + (ya - yi).astype(numpy.float) /
                           (yj - yi) * (xj - xi)) < xa)
+
             result[tf == True] ^= cross[tf == True]
             xj, yj = xi, yi
 
--- a/ginga/misc/log.py
+++ b/ginga/misc/log.py
@@ -54,7 +54,7 @@ class NullLogger(object):
 def get_logger(name='ginga', level=None, null=False,
                options=None, log_file=None, log_stderr=False):
 
-    if null or options.nulllogger:
+    if null or (hasattr(options, 'nulllogger') and options.nulllogger):
         # User wants a Null Logger
         return NullLogger()
 


pllim:
The way this is implemented, it does not quite fit into the `Drawing` plugin without some special handling. I hope this is not a deal-breaker.

ejeschke:
@pllim, I was hoping to add `Annulus` to the `astro` section of the `ginga.canvas.types` submodule.  Also, we can (and should) add the drawing support by adding an `idraw` and `get_params_metadata` class methods.

ejeschke:
I think that a variety of different kinds of annuli (box, ellipse, etc.) could be supported by an `AnnulusMixin` class.  But that can be an iteration after getting the basic circular annulus in there.

In the basic case, can we consider the annulus to be two compatible shapes with the outer one sharing the center point of the inner one, but having different radius/radii?

pllim:
It did not even cross my mind that annulus can be non-circular. Is there a need for other shapes?

https://en.wikipedia.org/wiki/Annulus_%28mathematics%29

pllim:
@ejeschke , I moved `Annulus` to `astro.py` and deleted `annulus.py`. I added commented stubs for `idraw()` and `get_params_metadata()` methods but do not know how to implement them.

The annulus is a compound object of two circles of different radii. There is no way for me to draw it interactively without a fixed annulus width, which is undesirable in a fully interactive mode. Does this make sense?

ejeschke:
ds9 provides a Box Annulus and an Elliptical Annulus in addition to the normal circular one.  Not sure how much these are actually used in practice, but I could imagine them being useful.  And if you look at your code for the three methods (`contains`, `contains_arr` and `get_llur`) you can see that these would work with boxes or ellipses without changing any code.

ejeschke:
@pllim, I'll work up an example of the Annulus with interactive drawing.

ejeschke:
@pllim, can you try commit 6df08dc8dd9398534a3f43e1a14ea15c5730bad0 ?  Invoke the `Drawing` plugin and select 'annulus' as the drawing type.  This object can also be used for your plugin if you just change the way the constructor is called.  See `ginga.canvas.types.astro` for details.  Let me know what you think.

pllim:
@ejeschke , it works. Thanks! Some comments about its behavior in `Drawing` plugin below:

* When I left-click on the main display, only one circle or square box is drawn (not two that are needed to define an annulus).
* There needs to be documentation to tell users to enter the `width` value *before* drawing an annulus, because there is no way to modify the width interactively.
* Just the `Drawing` GUI in general, it is pointless to have some of the attributes editable because editing them using text boxes do nothing because they are interactively set by the mouse. For example, `x`, `y`, and `radius`.

ejeschke:
@pllim, some good points.  Please try the editing function of the canvas to adjust the width of the annulus--click the Edit button at the bottom of the plugin, then click in the annulus to select it--this should show control points which can be used for adjusting position, radius and width.

pllim:
@ejeschke , I select `Edit` but I do not see any "control points". Also, the `Attributes` box becomes empty, so I cannot manually set the value either.

pllim:
Also, in trying to use it in my custom plugin (that uses "old style" drawing), the following no longer works:

* Changing annulus width of the displayed annulus by setting `obj.width = new_width`
* Changing annulus line style of the displayed annulus by setting `obj.linestyle = new_style`

They worked when I directly accessed `annulus.objects` and set the parameters that way, but I imagine one good thing about your implementation is that I no longer need to do it the ugly way.

ejeschke:
@pllim, after clicking "Edit", *click on the annulus to select it*.

With the current implementation, after assigning `obj.width`, etc. call `obj.sync_state()` to propagate the relevant changes to the sub-objects.

pllim:
Ah, yes, I missed the clicking to select part. :)  In playing with the editing mode, I realized that the reason I only see 1 circle created when I left-click in the drawing mode is because the inner radius is very small (probably 1 pixel) no matter what I enter in the text box.

pllim:
Back to the `Annulus` object, `sync_state()` only works up to some point. The following still does not work:

* `obj.move_to()` fails to update `obj.x` and `obj.y`. It reverts to the pre-move coordinates as soon as I release the mouse.
* `image.cutout_shape(obj)` returns masked array with *all* pixels masked (no "good" data to process)

pllim:
Update: I fixed `cutout_shape()` problem by reordering the mixin order (see my updated PR). But I am still having problem with `move_to()`.

pllim:
Update: I fixed the `move_to()` problem as well. See my PR.

pllim:
Update: I added a fix for Polygon here too.

ginga/canvas/types/basic.py
```diff
@@ -158,7 +159,7 @@ def contains_arr(self, x_arr, y_arr):
         # See: http://alienryderflex.com/polygon/
         xa, ya = x_arr, y_arr
 
-        result = numpy.empty(ya.shape, dtype=numpy.bool)
+        result = numpy.empty((ya.size, xa.size), dtype=numpy.bool)
```
ejeschke: @pllim, this change breaks the basic polygon containment test.  I'm pretty certain the result is supposed to be a 1D array of bools that is as long as the number of coordinates being tested--that length is `ya.shape` or `xa.shape` (they should be equivalent).  This would be consistent with the other canvas items.

Can you describe the case where it needs to be otherwise?

ejeschke:
Otherwise the PR looks good.

ginga/canvas/types/basic.py
pllim: Without this "fix", I get the following error:
```
2015-09-01 10:16:24,820 | E | Callback.py:89 (make_callback) | Traceback:
  File ".../ginga-2.4.20150827203255-py2.7.egg/ginga/misc/Callback.py", line 72, in make_callback

  File "~/.ginga/plugins/SurfBgRatio.py", line 390, in update
    self.draw_cb(canvas, tag)

  File "~/.ginga/plugins/SurfBgRatio.py", line 469, in draw_cb
    return self.redo()

  File "~/.ginga/plugins/SurfBgRatio.py", line 307, in redo
    sci_masked = image.cutout_shape(sig_obj)

  File ".../ginga-2.4.20150827203255-py2.7.egg/ginga/BaseImage.py", line 330, in cutout_shape
    view, mask = self.get_shape_view(shape_obj)

  File ".../ginga-2.4.20150827203255-py2.7.egg/ginga/BaseImage.py", line 318, in get_shape_view
    contains = shape_obj.contains_arr(xi, yi)

  File ".../ginga-2.4.20150827203255-py2.7.egg/ginga/canvas/types/basic.py", line 181, in contains_arr
    result[tf == True] ^= cross[tf == True]
```

This is the info I get when I added a print function in `canvas/types/basic.py`:
```python
>>> print(xa.shape, ya.shape, tf.shape, result.shape, cross.shape)
(1, 136), (131, 1), (131, 136), (131, 1), (131, 136)
```

ejeschke: @pllim, why do `xa` and `ya` have different shapes?  `xa` (xarr) should be an array of the X values of all the points, and `ya` (yarr) should be an array of all the Y values of the points.  If there are N points, then there should be an equal number of X and Y values.

Usually when you have an array of points X, Y you can get these two arrays at "no cost" by simply taking the transpose view of the array and indexing the 0 (X) and 1 (Y) elements. i.e.:
```python
contains = obj.contains_arr(points.T[0], points.T[1])
```

pllim: @ejeschke , I don't know why, the `xa` and `ya` are actually `xi` and `yi` in `BaseImage.get_shape_view()`:
```python
x1, y1, x2, y2 = map(int, shape_obj.get_llur())

if avoid_oob:
    # avoid out of bounds indexes
    wd, ht = self.get_size()
    x1, x2 = max(0, x1), min(x2, wd-1)
    y1, y2 = max(0, y1), min(y2, ht-1)

# calculate pixel containment mask in bbox
yi = numpy.mgrid[y1:y2+1].reshape(-1, 1)
xi = numpy.mgrid[x1:x2+1].reshape(1, -1)
contains = shape_obj.contains_arr(xi, yi)  #  <-------- CRASHES HERE
```

pllim: Below are the data in the method when the crash happened:
```
>>> shape_obj.points
[(1201.3878954607978, 739.6162310866575),
 (1257.7015130674004, 686.1182943603851),
 (1232.360385144429, 621.3576341127923),
 (1170.4154057771664, 607.2792297111416),
 (1142.2585969738652, 646.6987620357634),
 (1122.5488308115544, 714.2751031636864),
 (1122.5488308115544, 717.0907840440166)]
>>> xi
[[1122 1123 1124 1125 1126 1127 1128 1129 1130 1131 1132 1133 1134 1135
  1136 1137 1138 1139 1140 1141 1142 1143 1144 1145 1146 1147 1148 1149
  1150 1151 1152 1153 1154 1155 1156 1157 1158 1159 1160 1161 1162 1163
  1164 1165 1166 1167 1168 1169 1170 1171 1172 1173 1174 1175 1176 1177
  1178 1179 1180 1181 1182 1183 1184 1185 1186 1187 1188 1189 1190 1191
  1192 1193 1194 1195 1196 1197 1198 1199 1200 1201 1202 1203 1204 1205
  1206 1207 1208 1209 1210 1211 1212 1213 1214 1215 1216 1217 1218 1219
  1220 1221 1222 1223 1224 1225 1226 1227 1228 1229 1230 1231 1232 1233
  1234 1235 1236 1237 1238 1239 1240 1241 1242 1243 1244 1245 1246 1247
  1248 1249 1250 1251 1252 1253 1254 1255 1256 1257]]
>>> yi
[[607]
 [608]
 [609]
 [610]
 [611]
 [612]
 [613]
 [614]
 [615]
 [616]
 [617]
 [618]
 [619]
 [620]
 [621]
 [622]
 [623]
 [624]
 [625]
 [626]
 [627]
 [628]
 [629]
 [630]
 [631]
 [632]
 [633]
 [634]
 [635]
 [636]
 [637]
 [638]
 [639]
 [640]
 [641]
 [642]
 [643]
 [644]
 [645]
 [646]
 [647]
 [648]
 [649]
 [650]
 [651]
 [652]
 [653]
 [654]
 [655]
 [656]
 [657]
 [658]
 [659]
 [660]
 [661]
 [662]
 [663]
 [664]
 [665]
 [666]
 [667]
 [668]
 [669]
 [670]
 [671]
 [672]
 [673]
 [674]
 [675]
 [676]
 [677]
 [678]
 [679]
 [680]
 [681]
 [682]
 [683]
 [684]
 [685]
 [686]
 [687]
 [688]
 [689]
 [690]
 [691]
 [692]
 [693]
 [694]
 [695]
 [696]
 [697]
 [698]
 [699]
 [700]
 [701]
 [702]
 [703]
 [704]
 [705]
 [706]
 [707]
 [708]
 [709]
 [710]
 [711]
 [712]
 [713]
 [714]
 [715]
 [716]
 [717]
 [718]
 [719]
 [720]
 [721]
 [722]
 [723]
 [724]
 [725]
 [726]
 [727]
 [728]
 [729]
 [730]
 [731]
 [732]
 [733]
 [734]
 [735]
 [736]
 [737]
 [738]
 [739]]
```

pllim: If you do not want to change `PolygonMixin.contains_arr()`, then the fix will either have to be in `BaseImage.get_shape_view()`, or even in `PolygonMixin.get_llur()`?

ejeschke: Thanks, that's very helpful.  I'll take a look and see where this is different from the other shape cases.

ejeschke:
@pllim, I modified the PR a little to promote the arrays in `contains_arr()` if necessary and it seems to be working fine (merged into c776a30cb2c0d5d09d196f7225d3b3cfe28cec32, although this PR doesn't seem to be updating).

Can you test?  Seems working for my use cases.  Sorry for the confusion above--I think the case of different lengths for X and Y arrays is normal and true for most object containment tests when the size of the bounding box is not square.

pllim:
@ejeschke , I rebased my PR and don't see anymore remaining unmerged changes. My problem is fixed. Thank you!

ejeschke:
Good.  Weird that it didn't record the merge of the PR.  Usually even if I commit on top of the PR github recognizes the merge.

pllim:
Yeah, but it doesn't matter. Looks like my commits are still listed in master regardless.

PR #154, opened 2015-08-20
Added Cuts legend

rupak0577:
A legend for the cuts plot using matplotlib.


nabobalis:
This looks fine.

ejeschke:
I've made a slight modification to only show the legend if a preference is set.  I would be willing to set the default preference to True if there is enough interest, but for now you can simply set it as shown [here](https://github.com/ejeschke/ginga/blob/master/examples/configs/plugin_Cuts.cfg)

pllim:
:+1:  

This is what I preferred but it does not have to be default:
```python
label_cuts = False
show_cuts_legend = True
```

PR #169, opened 2015-10-02
Turn ginga.*.plugins into namespace packages

pllim:
Turn `ginga.*.plugins` into namespace packages, as discussed in #92.

With this change, I can add my own custom plugins under the `ginga.qtw.plugins` namespace with this directory structure:
```
+- mygingastuff/
    +- ginga/
    |   +- __init__.py
    |   +- qtw/
    |   |   +- __init__.py
    |   |   +- plugins/
    |   |   |   +- MyGlobalPlugin2.py
    |   |   |   +- MyLocalPlugin2.py
    |   |   |   +- __init__.py
    +- setup.py
```

The contents of each `__init__.py` (as pointed out by @embray):
```python
try:
    # As long as we're using setuptools/distribute, we need to do this the
    # setuptools way or else pkg_resources will throw up unncessary and
    # annoying warnings (even though the namespace mechanism will still
    # otherwise work without it).
    # Get rid of this as soon as setuptools/distribute is dead.
    __import__('pkg_resources').declare_namespace(__name__)
except ImportError:
    pass
__path__ = __import__('pkgutil').extend_path(__path__, __name__)
```

This is an example `setup.py`:
```python
#! /usr/bin/env python
#
try:
    from setuptools import setup
except ImportError:
    from distutils.core import setup

setup(
    version = '0.1.0',
    name = 'mygingastuff',
    namespace_packages = ['ginga'],
    packages = ['ginga', 'ginga.qtw', 'ginga.qtw.plugins'],
    author = 'J. Doe',
    description = 'Ginga Qt plugins',
    long_description = 'Ginga Qt plugins'
)
```

To start Ginga with these plugins from command line (specifying the full path to the custom plugins is a must due to how Ginga finds them via `main.py -> Control.py -> misc/ModuleManager.py`):
```
ginga --loglevel=20 --log=ginga.log --plugins=ginga.qtw.plugins.MyLocalPlugin2 --modules=ginga.qtw.plugins.MyGlobalPlugin2 myimage.fits
```

As far as Python is concerned, it is the same namespace although the plugins reside in two separate actual packages:
```python
>>> from ginga.qtw.plugins import Contents
>>> Contents
<module 'ginga.qtw.plugins.Contents' from '.../ginga-2.5.20151001032930-py2.7.egg/ginga/qtw/plugins/Contents.pyc'>
>>> from ginga.qtw.plugins import MyLocalPlugin2
>>> MyLocalPlugin2
<module 'ginga.qtw.plugins.MyLocalPlugin2' from '.../mygingastuff-0.1.0-py2.7.egg/ginga/qtw/plugins/MyLocalPlugin2.pyc'>
```

### DIFF
--- a/ginga/Control.py
+++ b/ginga/Control.py
@@ -432,7 +432,8 @@ class GingaControl(Callback.Callbacks):
             name = spec.setdefault('name', spec.get('klass', spec.module))
             self.local_plugins[name] = spec
 
-            self.mm.loadModule(spec.module, pfx=pluginconfpfx)
+            pfx = spec.get('pfx', pluginconfpfx)
+            self.mm.loadModule(spec.module, pfx=pfx)
 
             hidden = spec.get('hidden', False)
             if not hidden:
@@ -447,7 +448,8 @@ class GingaControl(Callback.Callbacks):
             name = spec.setdefault('name', spec.get('klass', spec.module))
             self.global_plugins[name] = spec
 
-            self.mm.loadModule(spec.module, pfx=pluginconfpfx)
+            pfx = spec.get('pfx', pluginconfpfx)
+            self.mm.loadModule(spec.module, pfx=pfx)
 
             self.gpmon.loadPlugin(name, spec)
             self.add_plugin_menu(name)
--- a/ginga/__init__.py
+++ b/ginga/__init__.py
@@ -1,7 +1,7 @@
 """
 Ginga is a toolkit designed for building viewers for scientific image
-data in Python, visualizing 2D pixel data in numpy arrays.  
-The Ginga toolkit centers around an image display class which supports 
+data in Python, visualizing 2D pixel data in numpy arrays.
+The Ginga toolkit centers around an image display class which supports
 zooming and panning, color and intensity mapping, a choice of several
 automatic cut levels algorithms and canvases for plotting scalable
 geometric forms.  In addition to this widget, a general purpose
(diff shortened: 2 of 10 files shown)


embray:
Well, it looks like it didn't blow up at least? :)

ejeschke:
@pllim, does this work under python 3 ?

ejeschke:
@embray, does every `__init__.py` have to have this declaration or can it just be included at some upper-level?  I'm thinking that plugins might want to add data files, icons, etc.-- in which case do we have to add this to every directory affected in the ginga source tree?  Would it just be sufficient to have it at top-level?

embray:
It only applies to the namespace package itself, not submodules.

The reason is that if you type, say, "import ginga", there's no guarantee which "ginga" package will be imported first (since now every package you install that belongs to the ginga package will have a ginga/ directory with an `__init__.py`).  So whichever one gets imported first has to declare the namespace package, (which then also searches sys.path for other paths for that package and adds them to `ginga.__path__`).

pllim:
> does this work under python 3 ?

Yes.

> does every `__init__.py` have to have this declaration?

When I remove the incantation from submodule init, the following fails, so as far as I am concerned, it needs to be repeated on all levels.
```python
>>> from ginga.qtw.plugins import Contents, MyLocalPlugin2
```

pllim:
**Hold your horses!** Even though the import works (namespace is somehow shared), Ginga still cannot find my local plugins. It worked last Friday because I forgot I had that plugin already copied to my `~/.ginga/plugins` directory! I removed that today and it does not work anymore. Ops!

pllim:
The question now becomes, how does Ginga actually find modules that share namespace but not actually in the core distribution? @ejeschke ?

pllim:
Update: It turns out the way `misc/ModuleManager.py` works, a prefix of the full import path is needed for the custom plugin to be recognized. I have updated the PR. This works now (for me, at least).

pllim:
I added a piggy-back commit so Qt file dialog forgets old directory from old Qt session. For some reason, the file dialog on my machine remembers a directory that wasn't even related to my work using Ginga. Setting `directory=os.curdir` seems to solve that problem.

ejeschke:
Ok, merged.  I think we'll add a plugin auto discovery feature as a separate issue/PR.

Issue #175, opened 2015-10-13
Adding FITS convenience function to Ginga util


pllim:
I have been using this in my Ginga plugins to find and load an extension of interest in my multi-extension FITS images:
```python
from astropy.io import fits 

def fits_extnamever_lookup(filename, extname, extver):
    """Return ext num for given name and ver."""
    extnum = -1
    with fits.open(filename) as pf:
        for i, hdu in enumerate(pf):
            if hdu.name.startswith(extname) and hdu.ver == extver:
                extnum = i
                break
    return extnum
```
This is because sometimes `(DQ, 1)` is in extension 2, sometimes in extension 3, which is beyond my control.

There are 2 ways we can do this:

1. If there is already a way for Ginga FITS file handlers to open an extension by `(extname, extver)` pair instead of `extnum` only, then this is unnecessary (I couldn't find such functionality when I looked, but I could have just missed it).

2. We can add this somewhere in `ginga.util.io_fits` module. Since you have two different FITS file handlers, it is not obvious to me what would be the most elegant way to implement this in a way that I do not break existing behaviors.

ejeschke:
I think we ought to follow astropy's conventions for this; namely, that you can tack on [name, ver] to the path and it will do the right thing.  And ideally these parameters just get passed to index into the astropy `HDUList` in the opener .  The fitsio opener should also be modified to allow the syntactical convention, as best as we can do it.

pllim:
Actually, this lookup function might still be useful, depending on how flexible you want `Datasrc` dictionary lookup to be. Currently I have to convert, say, `image.fits[('DQ', 1)]` to `image.fits[3]` so I can grab the cached data using `chinfo.datasrc['image.fits[3]']`.

ejeschke:
@pllim, can you try commit 593c2fc50ca47b96b4020ec8a9fe108725560ef6?  This adds support for loading by `(extname, extver)`.  This also adds support for the image name so that the data src should contain a reasonable name for it.

Note that `MultiDim` is still not updated (yet) to open the HDU by this method (it is still using straight indexes), so you might need to keep your function around for a bit longer.  But I think it can be refactored to open them via (extname, extver) so that you get more reasonable names in the Contents and Thumbs.

pllim:
@ejeschke , the latest version broke `MultiDim` (as you pointed out?) and some of my plugins. This is because the keys in `Datasrc` changed from
```python
['image1[1]', 'image2[1]', 'image1[3]']
```
to
[...]

PR #176, opened 2015-10-13
Flex FITS extension naming

pllim:
Alternate solution for #175. Seems to work when I tested this as follows:
```python
im = AstroImage()
im.load_file('image001.fits', ('DQ', 1))
dat = im.get_data()
plt.imshow(dat)
```
Also works on:
```python
im.load_file('image001.fits', 'PRIMARY')
im.load_file('image001.fits', 1)
```

### DIFF
--- a/ginga/AstroImage.py
+++ b/ginga/AstroImage.py
@@ -135,6 +135,8 @@ class AstroImage(BaseImage):
         name = self.get('name', None)
         if name is None:
             name = info.name
+            if ('[' not in name):
+                name += '[{0}]'.format(numhdu)
             self.set(name=name)
 
         self.set(path=filepath, idx=numhdu)
--- a/ginga/util/iohelper.py
+++ b/ginga/util/iohelper.py
@@ -18,29 +18,48 @@ def get_fileinfo(filespec, cache_dir='/tmp', download=False):
     """
     Parse a file specification and return information about it.
     """
-    idx = None
-    name_ext = ''
+    # Loads first science extension by default.
+    # This prevents [None] to be loaded instead.
+    idx = ('SCI', 1)
+    name_ext = '[SCI,1]'
 
     # User specified an HDU using bracket notation at end of path?
-    match = re.match(r'^(.+)\[(.+)\]$', filespec)
+
+    # This matches
+    #   image.fits[("NAME", 1)]
+    #   image.fits[('NAME', 1)]
+    #   image.fits['NAME', 1]
+    #   image.fits["NAME", 1]
+    #   image.fits[NAME, 1]
+    #   image.fits[(NAME, 1)]
+    match = re.match(r'^(.+)\[\(?[\'\"]?([A-Za-z]+)[\'\"]?\,([0-9])\)?\]$',
+                     filespec)
     if match:
         filespec = match.group(1)
-        idx = match.group(2)
-        if ',' in idx:
-            hduname, extver = idx.split(',')
-            hduname = hduname.strip()
-            extver = int(extver)
-            idx = (hduname, extver)
-            name_ext = "[%s,%d]" % idx
-        else:
+        hduname = match.group(2)
+        extver = int(match.group(3))
+        idx = (hduname, extver)
+        name_ext = "[%s,%d]" % idx
+
+    else:
+        # This matches
+        #   image.fits[1]
+        #   image.fits["NAME"]
+        #   image.fits['NAME']
+        #   image.fits[NAME]
+        match = re.match(r'^(.+)\[\(?[\'\"]?([A-Za-z0-9]+)[\'\"]?\)?\]$',
+                         filespec)
+        if match:
+            filespec = match.group(1)
+            idx = match.group(2)
             if re.match(r'^\d+$', idx):
                 idx = int(idx)
                 name_ext = "[%d]" % idx
             else:
                 idx = idx.strip()
                 name_ext = "[%s]" % idx
-    else:
-        filespec = filespec
+        else:
+            filespec = filespec
 
     url = filespec
     filepath = None
@@ -93,7 +112,7 @@ def name_image_from_path(path, idx=None):
             hduname, extver = idx
             hduname = hduname.strip()
             extver = int(extver)
-            name += "[%s,%d]" % idx
+            name += "[%s,%d]" % (hduname, extver)
         else:
             if isinstance(idx, str):
                 name += "[%s]" % idx.strip()
--- a/ginga/util/iohelper.py
+++ b/ginga/util/iohelper.py
@@ -24,42 +24,25 @@ def get_fileinfo(filespec, cache_dir='/tmp', download=False):
     name_ext = '[SCI,1]'
 
     # User specified an HDU using bracket notation at end of path?
-
-    # This matches
-    #   image.fits[("NAME", 1)]
-    #   image.fits[('NAME', 1)]
-    #   image.fits['NAME', 1]
-    #   image.fits["NAME", 1]
-    #   image.fits[NAME, 1]
-    #   image.fits[(NAME, 1)]
-    match = re.match(r'^(.+)\[\(?[\'\"]?([A-Za-z]+)[\'\"]?\,([0-9])\)?\]$',
-                     filespec)
+    match = re.match(r'^(.+)\[(.+)\]$', filespec)
     if match:
         filespec = match.group(1)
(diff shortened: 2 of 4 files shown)


ginga/AstroImage.py
```diff
@@ -142,7 +142,7 @@ def load_file(self, filepath, numhdu=None, naxispath=None):
             if '.' in name:
                 name = name[:name.rindex('.')]
             if numhdu is not None:
-                name += ('[%d]' % numhdu)
+                name += '[{0}]'.format(numhdu)
             self.set(name=name)
 
         self.set(path=filepath, idx=numhdu)
```
ejeschke: Not sure, but I think there might be some places where `idx` is assumed to be an integer.  I'll dig around a little bit.  Also, does this work for `fitsio`-based loader?

pllim: > does this work for fitsio-based loader?

I tried both settings by modifying `general.cfg`. So, if testing the way I did actually grabs the setting from the config file, then the answer is yes.

ejeschke:
@pllim, I think we need to modify a few more bits of code.  For example, `ginga.util.iohelper` is used to parse a filespec and pull out the HDU information.  Currently it is only matching against an integer.  So we should enumerate the possible conventions:
1) integer: numerically indexes the HDU
2) name: names the HDU
3) name,integer: names the HDU with an extver

Does that about cover it?  I verified that your example does work with the fitsio loader, so that's good news.

pllim:
@ejeschke , thanks for following up on this. I updated this PR with some fixes to your latest commit, as mentioned in #175. Yes, I think the three conventions you listed are pretty much it AFAIK.

ejeschke:
Hmm, with these updates `Thumbs` doesn't seem to be picking up the name, and so I get the same thumbail being updated as I load HDUs from `MultiDim`.

pllim:
Sorry, maybe you have an older commit. I just pushed more changes, which should work now. I like to squash intermediate commits, so it's not obvious from commit messages.

pllim:
I am still working on it... I think the final hurdle is to get `Datasrc` to realize that "image.fits[1]" is the same as "image.fits['SCI', 1]".

ejeschke:
I think ultimately what would be most desirable is that the way you open the file, the image name follows the image throughout ginga.  So if I open the image with "[DQ,1]" appended the thumbnail shows that, it shows up that way in `Contents`, in the data src, etc.  `MultiDim` has to be modified so that it opens by (extname, extver) if possible.

ejeschke:
I think if you modify `MultiDim` to load by creating an (extname, extver) for the `idx` parameter then the datasrc will automatically be in sync.

pllim:
I also get something like this once in a while. I am starting to think it is a lot easier if we just internally convert all the extension names to the old index in `iohelper` (like what I wrote in #175).
```python
  File ".../ginga/qtw/plugins/Thumbs.py", line 71, in mouseReleaseEvent
    self.thumbs_cb()
  File ".../ginga/qtw/plugins/Thumbs.py", line 140, in <lambda>
    image_future)
  File ".../ginga/misc/plugins/ThumbsBase.py", line 195, in load_file
    image_future=image_future)
  File ".../ginga/Control.py", line 926, in switch_name
    self.change_channel(chname, image=image)
  File ".../ginga/Control.py", line 1058, in change_channel
    self._switch_image(chinfo, image)
  File ".../ginga/Control.py", line 996, in _switch_image
    raise_initialize_errors=False)
  File ".../ginga/ImageView.py", line 553, in set_image
    profile = image.get('profile', None)
AttributeError: 'exceptions.TypeError' object has no attribute 'get'
```

ejeschke:
I think that exception may be because you removed the `idx` parameter from `Control.py`'s `load_image` method.  I think `Thumbs` and `Contents` try to use that to restore the correct HDU.

pllim:
I changed `idx` to `numhdu` because they are starting to confuse me (they were inter-changable at a lot of places).

But the point is `Datasrc` needs a consistent naming scheme to work properly. Whether that name is `[idx]` or `[extname, extver]` is unimportant.

pllim:
To elaborate further, there is no way you can force people to only open by `idx` or `[extname, extver]` exclusively. There needs to be a way to cross-reference them. I think `Datasrc` is the best place for this to happen, but how, I am not sure. I am thinking along the line of storing a list of alternate keys or something. Ideas welcome!

ejeschke:
Before we merge #175, can you give commit 1f961dc65072c24cf5af06dc93088a92709963e4 a whirl?  I believe this fixes the issues with the names not being consistent.  Now when you open HDUs with `MultiDim` the HDU names and versions should show up consistently. Would be interested to hear your take on this.  Personally, I think it makes it easier to know which thumbnail is what.

I think the datasrc is probably not the place to resolve this issue.  It's really a FITS issue and having to do with multiple ways to access subparts of FITS file, so probably belongs in the `io_fits` openers--they should return a consistent "index".  As long as the image is given a consistent `path`, `name` and `idx` metadata it should work fairly consistently between the various plugins.

As an aside, the viewer part of ginga proper doesn't really "know" the images it is looking at come from FITS files--it's `AstroImage` that knows about the origin of the file.  Some of the plugins are FITS-specific, of course.  I think Mike D. had a nice test version of ASDF-aware `MultiDim` running at the last Python in Astronomy workshop.  I thought that was pretty cool.

pllim:
@ejeschke , thank you! I rebased and now my PR shows the remaining changes I wish to merge.

ginga/util/iohelper.py
```diff
@@ -18,29 +18,48 @@ def get_fileinfo(filespec, cache_dir='/tmp', download=False):
...
+
+    # This matches
+    #   image.fits[("NAME", 1)]
+    #   image.fits[('NAME', 1)]
+    #   image.fits['NAME', 1]
+    #   image.fits["NAME", 1]
+    #   image.fits[NAME, 1]
+    #   image.fits[(NAME, 1)]
+    match = re.match(r'^(.+)\[\(?[\'\"]?([A-Za-z]+)[\'\"]?\,([0-9])\)?\]$',
+                     filespec)
```
ejeschke: What about hyphens and underscore? Or periods?

@embray, do you know what the FITS standard allows for naming of HDUs?

pllim: Hmm. I did not consider the possibilities of special characters in EXTNAME. It took me quite a while to get that regex working, so I thought I would not want to "waste" the effort. :ghost: But if it is not practical, I can revert this change. Let me know.

ejeschke: @pllim, I just want to make sure that we cover all the cases for naming because I'm sure I'll get a bug report from someone that has a strangely named HDU.  FITS is terrible that way.

pllim:
@ejeschke , okay, that is a fair point. I removed my fancy `regex`.

ginga/util/iohelper.py
```diff
@@ -18,8 +18,10 @@ def get_fileinfo(filespec, cache_dir='/tmp', download=False):
     """
     Parse a file specification and return information about it.
     """
-    idx = None
-    name_ext = ''
+    # Loads first science extension by default.
+    # This prevents [None] to be loaded instead.
+    idx = ('SCI', 1)
+    name_ext = '[SCI,1]'
```
ejeschke: @pllim, I had a chance to test this today.  It fails to load any FITS file that does not have:
1) an extension
2) an extension named "SCI"

This includes many FITS files (most of our instruments, actually).  I think what we want to do is make the "canonical name" be the first HDU that can be opened.  So for HST files, if the first extension that can be opened is "[SCI, 1]" that becomes the suffix.  I'll work on this a bit more.

ejeschke: To clarify, I think I can modify `io_fits` to return the `numhdu` as (in your case) "[SCI, 1]" if `numhdu` is passed in as `None`.

pllim: That would be great if you can force `None` to be something meaningful in the name. The other route I was considering is making default "canonical name" a user preference somewhere, which seems unnecessarily complicated.

ejeschke:
@pllim, I merged a version of this with suggested fixes.  Can you try commit 86f9e0bf00fc283df5a5cf335beee5a325a60662 from master?

pllim:
@ejeschke , yeah, it works. Thanks!

Given that this is technically merged, I'll open a new PR for `nstools.py` and close this PR.

PR #177, opened 2015-10-15
Fixed thumbs paths and missing chinfo.readout error messages

pllim:
Fixed thumbs paths and missing `chinfo.readout` error messages.

1. For some reason, sometimes `chinfo` is returned as `None` and this causes an error in `Control.py`. Otherwise, does not seem to affect functionality.
2. Hardcoding thumbs directory to "/dev/null" in `Mosaic.py` is a bad idea. Changed it to working directory. Also, need to create nested directories if necessary.
3. There is a need to create nested directories if thumbnails are saved to `~/.ginga/thumbs` instead of local directory.

### DIFF
--- a/ginga/Control.py
+++ b/ginga/Control.py
@@ -340,6 +340,8 @@ class GingaControl(Callback.Callbacks):
         ColorBar in the Readout (if any).
         """
         chinfo = self.get_channelInfo()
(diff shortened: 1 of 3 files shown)


ejeschke:
Looks good.

Issue #166, opened 2015-10-01
Ginga data files


pllim:
If I want to develop a plugin that comes with data file(s), e.g., ASCII table(s), what is the best way to do it? Does Ginga have a scheme for distributing associated data files? Thanks!

ejeschke:
@pllim, I think whatever scheme we come up with for Issue#92 should also handle associated data for plugins.  I really want to solve this, but in a way that is clean and does not rely heavily on the changes between versions of python or python package manager *du jour* (python packaging seems to change as often as the wind).

pllim:
Now that #169 is merged, should we revisit this?

ejeschke:
How about adding a data directory that uses the namespace trick?

pllim:
Where should this `data` reside? I can give that a try and see.

ejeschke:
How about `ginga.data`?

Don't forget to add it to `setup.py`.  For now it can probably go into the `packages` list.

ejeschke:
Hmm, rethinking this.  It might need to go in `package_data`, probably following the way that ginga examples are now installed.  At least, that's most likely how you would specify the data files in *your* setup.py that was installing.  I bet @embray would know the answer to this.

pllim:
We can always utilize `astropy.utils.data.get_pkg_data_filename()` but I don't know if that will work with shared namespace, nor the best place to place the `data` directory in this situation.

ejeschke:
Let's use `ginga.data`.  Try to add it to your `package_data` in your install script for your plugin and see if you can install data files.

pllim:
That works. I put my data file under `mymodule/ginga/data` directory. And added `{'ginga': ['data/myfile.txt']}` to my `package_data` in `setup.py`.

The catch is that I need the latest `astropy` to find it because I need the new `package='ginga'` keyword in `get_pkg_data_filename()`. Otherwise, it goes into "search online" mode and gives me a weird `"cannot find _bsddb"` error.

pllim:
@ejeschke , if you are okay with what I reported above, you can close this issue.

ejeschke:
I think this is ok.  You can also (since you need an `__init__.py` in the directory, right?) do something like:
```python
mydatafile = "specialflags.dat"
from ginga.data import __file__
datadir, _x = os.path.split(__file__)
myfile = os.path.join(datadir, mydatafile)
```

pllim:
[...]

PR #178, opened 2015-10-19
Added data directory

pllim:
Added data directory and updated setup script. This is in response to a request in #166.

This directory will be used for Ginga data files, which can be accessed via Astropy as follows. The example below uses a custom local plugin but data file may belong to any sub-module within Ginga, as desired:
```python
from astropy.io import ascii
from astropy.utils.data import get_pkg_data_filename

from ginga import GingaPlugin


class MyLocalPlugin2(GingaPlugin.LocalPlugin):
    def __init__(self, fv, fitsimage):
        super(MyLocalPlugin2, self).__init__(fv, fitsimage)

        # Load associated data file in ginga/data
        # In this example, it is a plain ASCII table.
        # The "package" keyword is optional? I needed it for shared namespace but
        # unsure if it is part of the core distribution.
        datafile = get_pkg_data_filename('data/mylocalplugin2_table.txt', package='ginga')
        self.tab = ascii.read(datafile, ...)       
```

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -69,7 +69,7 @@ setup(
                      'ginga.doc': ['manual/*.html'],
                      'ginga.web.pgw': ['templates/*.html', 'js/*.js'],
                      'ginga.fonts': ['*/*.ttf', '*/*.txt'],
-                     'ginga': ['examples/*/*'],
+                     'ginga': ['data/*', 'examples/*/*'],
                      },
     scripts = ['scripts/ginga', 'scripts/grc', 'scripts/gris'],
     install_requires = ['numpy>=1.7'],


pllim:
@eteq, you might also be interested in this as you consider the alternate `stginga` packaging idea.

ejeschke:
@pllim, what is the status of this PR?  Did you resolve the issues with finding the data directory or do you want to pursue alternate implementation?  Just going through some remaining PRs now.

pllim:
This just adds a placeholder directory for your package data, as you requested. I do not really need this for my own package data in the shared namespace. Looking up my own package data is resolved by #183.

pllim:
If #183 is "no go", I think we can close this as well without merging.

ejeschke:
Sounds good.  I'll be interested to hear how your method works out.  I like the idea of plugins registering themselves somehow on the install so that there is no need for autodiscovery.

Are you using the shared namespace feature still?

pllim:
No, @eteq changed the structure such that it is contained within `stginga` itself. So our "discovery" uses the existing custom plugin architecture.

pllim:
I opened #207 for pursuing "registering".

PR #183, opened 2015-10-23
Added namespace tools module

pllim:
Added namespace tools module.

This is a solution for:

* Finding data files in shared namespace (see #166 )
* Finding custom plugins in shared namespace (see #92) -- Can only find them, cannot auto-load them

### DIFF
--- a/ginga/util/nstools.py
+++ b/ginga/util/nstools.py
@@ -1,0 +1,145 @@
+#
+# Copyright (c) Eric R. Jeschke.  All rights reserved.
+# This is open-source software licensed under a BSD license.
+# Please see the file LICENSE.txt for details.
+#
+"""Tools for shared Ginga namespace.
+
+These can be used by other developer writing plugins
+under the same Ginga namespace but in a separate package.
+"""
+
+# STDLIB
+import os
+
+
+# Like astropy.utils.data.get_pkg_data_filename() but not quite.
+def get_pkg_data_filename(data_name, package_name='ginga'):
+    """Get package data filename in shared namespace.
+
+    Parameters
+    ----------
+    data_name : str
+        Data path and filename. For example: ``'data/mydata.txt'``
+
+    package_name : str
+        The default value of ``'ginga'`` assumes data files
+        are in ``ginga/data`` directory.
+
+    Returns
+    -------
+    data_fn : str
+        Path to package data filename, if found.
+
+    """
+    data_name = os.path.normpath(data_name)
+    data_fn = ''
+
+    # Traverse through packages in shared namespace.
+    # Get the first match.
+    for path in __import__(package_name).__path__:
+        fn = os.path.join(path, data_name)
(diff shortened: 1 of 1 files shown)


ginga/util/nstools.py
```diff
@@ -0,0 +1,145 @@
+#
+# Copyright (c) Eric R. Jeschke.  All rights reserved.
+# This is open-source software licensed under a BSD license.
+# Please see the file LICENSE.txt for details.
+#
+"""Tools for shared Ginga namespace.
+
+These can be used by other developer writing plugins
+under the same Ginga namespace but in a separate package.
```
pllim: Perhaps more suitable to be part of `ginga/misc/PluginManager.py`?

ginga/util/nstools.py
```diff
@@ -0,0 +1,145 @@
...
+    # No custom plugins
+    if len(paths) < 2:
+        return plugins
+
+    # Identify core distribution path
+    coremod = __import__('ginga.Control')  # Only core has this?
+    corepath = os.path.dirname(coremod.__file__)
+
+    # Find custom plugins in shared namespace.
+    # TODO: How to separate global and local plugins?
```
pllim: If we can figure out how to identify and separate global and local plugins, perhaps this is better off as a method within `PluginManager` class?

ginga/util/nstools.py
```diff
@@ -0,0 +1,145 @@
...
+            if pfx in plugins:
+                plugins[pfx].union(pnames)
+            else:
+                plugins[pfx] = pnames
+
+    # Search HOME (*py and *pyc give duplicates)
+    if include_home:
+        home = os.path.join(os.path.expanduser('~'), '.ginga', 'plugins')
+        if os.path.isdir(home):
+            pnames = set(os.path.splitext(fn)[0] for fn in os.listdir(home))
```
pllim: If this becomes a method, the repeated incantation of finding `pnames` and inserting it into the dictionary can be a hidden method and is called here?

pllim:
@eteq, the "auto discovery" function is in this PR. FYI.

ginga/util/nstools.py
ejeschke: Well, they do have different constructor signatures.  We probably can use python's `inspect` module to determine whether it is a local or global plugin.

pllim:
It appears that STScI decided to not use shared namespace after all as per spacetelescope/stginga#18 . Do you still want to pursue this in case others are interested, or should we close this as "no go"?

ejeschke:
If STScI doesn't need this now then I think we can just close it.  I suspect it will resurface again as an issue, but maybe then I can learn from your experience.

pllim:
Do you want me to revert #169 as well?

ejeschke:
No, I think we'll leave that one for now.

pllim:
Okay.

Issue #184, opened 2015-10-26
Add "short name" display to Thumbs preferences


pllim:
@hcferguson pointed out that when image names are very long, thumbnails are displayed such that you only see one thumbnail per row (instead of the usual two or three for the same GUI dimension). The problem is STScI often encounters very long image names, especially for HLA and JWST.

So, I am wondering if it is possible to add a preference to `Thumbs` plugin so it only displays a short version of image name that is configurable by the user. Possible options, where the display length and display pattern are set by user:

1. First ``n`` characters, followed by ``...`` and the extension name.
2. ``...``, followed by the last ``n`` characters and the extension name.

Currently:
```
very_very_very_long_name[ext]
    +-------------+
    | thumbnail   |
    |   image     |
    +-------------+
```

Desired display (option 1):
```
    very_ve...[ext]
    +-------------+
    | thumbnail   |
    |   image     |
    +-------------+
```

Desired display (option 2):
```
    ...g__name[ext]
    +-------------+
    | thumbnail   |
    |   image     |
    +-------------+
```

This is just an idea. Perhaps @hcferguson can provide more details or correct me if he disagrees, since he is the one who reported this.

On the other hand, maybe we can instead utilize the full control on `AstroImage` name in `Datasrc`, as you mentioned in the last comment in #175. But, automatically determining an unique short name for an arbitrary image file, or how/where we let user to customize it would need more thoughts.

Or we can close this if you think the implementation would be too complicated.

hcferguson:
I think maybe a mouse hover showing the name might be a better solution, if that's feasible.

pllim:
@hcferguson, that is actually easier than what I had in mind, which is a good thing! Maybe you could give my PR in #185 a try and see if that is what you wanted?

ejeschke:
I was already thinking about option 1 combined with the mouseover, so maybe that is a good sign.

option 1 should not be hard, just need to save the full name in the `Bunch` that is created to store the thumb info and modify the label that is set, based on some name-mangling method.  Might be good to put that mangling method in a util file so it can be easily shared by `Contents` as well (although maybe `Contents`, with its resizable name column, doesn't need this modification).

Mouseover support means modifying the tooltip to add the name as the first item.

pllim:
@ejeschke, I have an idea about the mouseover part in #185

pllim:
Fixed by #185.

PR #185, opened 2015-10-26
Thumbs preferences to hide labels

pllim:
This is a proposed solution for #184, as suggested by @hcferguson.

* Added `Thumbs` preferences to shorten labels for very long image names.
* Added image name to tooltip (not configurable).
* Added previously undocumented `thumb_length` setting to `config` example.

To enable shortening of thumbnail labels, add the following to `~/.ginga/plugin_Thumbs.cfg`:
```python
# Mandatory unique image identifier in tooltip
mouseover_name_key = 'NAME'

# Thumbnail label length in num of characters (None = no limit)
label_length = 25

# Cut off long label ('front' or 'back')
#label_cutoff = 'front'
label_cutoff = 'back'
```

### DIFF
--- a/ginga/misc/plugins/ThumbsBase.py
+++ b/ginga/misc/plugins/ThumbsBase.py
@@ -41,10 +41,11 @@ class ThumbsBase(GingaPlugin.GlobalPlugin):
                                   auto_scroll=True,
                                   rebuild_wait=4.0,
                                   tt_keywords=tt_keywords,
+                                  mouseover_name_key='NAME',
                                   thumb_length=150,
                                   sort_order=None,
-                                  label_name=True,
-                                  mouseover_name_key='NAME')
+                                  label_length=None,
+                                  label_cutoff='back')
         self.settings.load(onError='silent')
         # max length of thumb on the long side
         self.thumbWidth = self.settings.get('thumb_length', 150)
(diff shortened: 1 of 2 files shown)


ginga/misc/plugins/ThumbsBase.py
```diff
@@ -127,6 +129,12 @@ def add_image(self, viewer, chname, image):
             self.thumb_generator.set_image(image)
             imgwin = self.thumb_generator.get_image_as_widget()
 
+        if not self.settings.get('label_name', True):
+            thumbnamekey = self.settings.get('mouseover_name_key', 'NAME')
+            self.keywords.insert(0, thumbnamekey)
```
ejeschke: I think `self.keywords` is supposed to be a static list.  If you keep inserting here at this point in the code, I think the list will keep growing.  Maybe do it once in `__init__`?

ejeschke:
It would be nice to add the shortened name thing as well, because then you have some small visual indicator of the name.

pllim:
@ejeschke , okay. I replaced the option to hide label with the option to shorten it. I'd rather not have both for simplicity.

ejeschke:
Merged with some cleanup and a bug fix.  Notable changes:
- shortening names is now the default, and defaults to 25 characters
- label_cutoff should be "left" or "right" (I feel this is more intuitive in regard to the name--if "right" (the default) the name is cut off on the right)

pllim:
Thank you, @ejeschke ! Yes, your defaults make more sense. :+1:

PR #187, opened 2015-10-26
Line profile fixes

rupak0577:
Mainly disables controls for 2D images

### DIFF
--- a/ginga/misc/plugins/LineProfile.py
+++ b/ginga/misc/plugins/LineProfile.py
@@ -155,8 +155,12 @@ class LineProfile(GingaPlugin.LocalPlugin):
                 chkbox = Widgets.CheckBox('NAXIS%d' % i)
                 self.hbox_axes.add_widget(chkbox)
 
-                # Add callback
-                self.axes_callback_handler(chkbox, i)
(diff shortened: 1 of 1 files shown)


ginga/misc/plugins/LineProfile.py
```diff
@@ -148,7 +148,10 @@ def new_image_cb(self, fitsimage, image):
         elif image is None:
             return False
         else:
-            self.build_axes()
+            try:
+                self.build_axes()
+            except:
+                pass
```
ejeschke: This sort of construct, simply squelching the exception, may hide other exceptions that would be good to see.  Instead, we need to fix `build_axes()` so that it doesn't throw an exception in the 2D case.

rupak0577:
@ejeschke It also throws other exceptions related to `hbox_axes` being not initialised. The problem is that I used `new_image_cb()` to reset the axes if a new image is loaded. But when the plugin is started, `new_image_cb()` is called instead of `build_gui()` which initialises the `hbox_axes` variable.

ejeschke:
Sorry for the long wait, @rupak0577.  I got busy with other changes and neglected this PR.

Issue #186, opened 2015-10-26
Thumbs and/or Contents to highlight currently displayed image


pllim:
As pointed out by @hcferguson, currently `Thumbs` and `Contents` (at least in Qt) currently do not highlight their thumbnail/entry that is being currently displayed. It would be convenient to have the active image highlighted so user knows its corresponding thumbnail and `Contents` entry without having to manually cross-reference via "Synopsis" (under `Info`) first.

pllim:
**Thumbs**

Maybe we can update `focus_cb()` to change thumbnail label color, say, to red, for the active image. However, I cannot find a trivial way to do this for Qt (since I don't really care about GTK) because `focus_cb()` is in `ThumbsBase` that is supposed to be toolkit-indepedent.

One idea is to add a method, say, `highlight_thumb()`, to do the actual highlighting. Call it from `Thumbsbase.focus_cb()`, implement it in `qtw/plugins/Thumbs.py`, and just have the GTK version to `pass` without actually doing anything.

@ejeschke, what do you think?

**Contents**

See #190 for Qt solution.

ejeschke:
This is actually a more tricky problem than it may first appear.  One complicating factor is that with alternative layouts you can have several channel windows visible--then you need to keep track of, and highlight, all the thumbs and contents entries for all the images in visible windows (and it is very easy to make an alternative layout--you can do it by just specifying a new layout in your `.ginga/ginga_config.py`).  Because plugins have a lot of latitude over their channel windows as well, they can manipulate them to load new images (think `MultiDim`, or `Mosaic`) so this also complicates the viewer trying to keep track of whether an image is still "being viewed" somewhere.

We could experiment with this by adding a callback that would be the counterpart to the `image-set` callback (i.e. `image-unset`?). This would be called when an image is displaced in the viewer by another one.  Either that or change the method signature of `image-set` to also include the image that was being displaced.

gtk version definitely needs to follow what the qt version is doing, as much as possible, in the general scheme of things for the reference viewer. But it's pretty straightforward to change the background of a label in both desktop widget sets.

pllim:
I don't think STScI has a need for GTK. If I submit the solution for Qt, is it possible for someone else to implement GTK later on?

ejeschke:
[...]

PR #190, opened 2015-10-26
Highlight active image in Contents (Qt)

pllim:
~~Qt~~ solution for `Contents` for #186.

~~If we want to be even more flexible, we can make the font color configurable, but will that be overkill?~~

* Enable user preference to set highlighted text color.
* Highlight active image properly when `highlight_tracks_keyboard_focus = True`. Highlighting logic is now like `Thumbs` (see https://github.com/ejeschke/ginga/compare/1c68ac3...1d6849d).
* Log failed highlighting as error instead of warning.
* Fixed error when image is removed from a channel, or when a channel is deleted.
* Also fixed similar error for `Thumbs` (follow up of #195).
* Added `sort_order` example in `channel_Image.cfg` (follow up of #226).
* Corrected verbiage in both `Contents` and `Thumbs` config examples.
* There is no need to make channel name all lowercase in the lookup keys in both `Contents` and `Thumbs`.

### DIFF
--- a/ginga/examples/configs/channel_Image.cfg
+++ b/ginga/examples/configs/channel_Image.cfg
@@ -103,6 +103,12 @@ numImages = 10
 follow_focus = True
 
 # ---------------
+# General - Listed here but has nothing to do with general.cfg
+
+# Sort channel history ('loadtime' or 'alpha')
+sort_order = 'alpha'
+
+# ---------------
 # Remember - When set to True, GUI remembers the settings for each image
 
 profile_use_scale = False
--- a/ginga/examples/configs/plugin_Contents.cfg
+++ b/ginga/examples/configs/plugin_Contents.cfg
@@ -10,7 +10,7 @@ columns = [ ('Name', 'NAME'), ('Object', 'OBJECT'), ('Filter', 'FILTER01'), ('Da
 # If set to True, will always expand the tree in Contents when new entries are added
 always_expand = True
 
-# Default is for Contents to highlight images that are displayed in channels.
+# Option to highlight images that are displayed in channels.
 # If set to True this option will only highlight the image that is in the
 # channel with the keyboard focus
 highlight_tracks_keyboard_focus = False
@@ -18,3 +18,6 @@ highlight_tracks_keyboard_focus = False
 # If True, color every other row in alternating shades to improve
 # readability of long tables
 color_alternate_rows = True
+
+# Highlighted row colors (in addition to bold text)
+row_font_color = 'green'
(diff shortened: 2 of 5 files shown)


pllim:
Like #195, I added user preference for highlight text color. But in this case, it is not practical to highlight the background, so I make the highlighted text bold instead. The behavior is similar to the aforementioned PR.

pllim:
Rebased. Still works.

pllim:
@ejeschke, after you merged the Qt and GTK versions into one, I cannot find an obvious way to highlight the active `Contents` entry anymore. My Qt version of the `focus_cb()` method is as follows:

```python
    def focus_cb(self, viewer, fitsimage):
        image = fitsimage.get_image()
        if image is None:
            return

        imname = image.get('name')
        chname = viewer.get_channelName(fitsimage)

        normal_tf = QtGui.QFont()
        normal_tc = QtGui.QColor('black')

        # The Bold and the Beautiful
        hlight_tf = QtGui.QFont()
        hlight_tf.setBold(True)
        hlight_tc = QtGui.QColor(self.settings.get('row_font_color', 'green'))

        # Traverse all channels.
        for key in self.nameDict:
            #key = key.lower()  # Channel name is sometimes capitalized?
            fileDict = self.name_dict[chname]
            chitem = fileDict['_chitem']

            # Highlight active image listing.
            # Cannot highlight background because listing has alternate-row
            # background color scheme.
            # Cannot use setSelected() because that causes circular callbacks.
            for i in range(chitem.childCount()):
                item = chitem.child(i)
                name = item.text(0)

                if key != chname or name != imname:
                    txfont = normal_tf
                    txcolor = normal_tc
                else:
                    txfont = hlight_tf
                    txcolor = hlight_tc

                for j in range(item.columnCount()):
                    item.setFont(j, txfont)
                    item.setTextColor(j, txcolor)
```

Any ideas? Thanks.

ejeschke:
@pllim, we are going to need to add a method to `ginga.qtw.Widgets` in the `TreeView` class to highlight a row.  This class in under heavy churn right now as I try to finish up the unification of `FBrowser` and `Catalogs`.  I'll add that function in the next commit, so hold off just a bit. Tnx.

pllim:
@ejeschke , rebased and ready for review. I updated my PR info on top. Thanks!

pllim:
@ejeschke , if possible, I would like this reviewed and merged as soon as possible. I find that I am spending more time than I like rebasing and resolving conflicts with every Ginga master updates. Thanks.

pllim:
@ejeschke , if possible, I would like this reviewed and merged as soon as possible, to avoid rebasing conflicts. As you suggested in #195, I updated my PR to use `set()` like `Thumbs`, among other things, which I listed in the PR description on top. Please give this a try and let me know if more changes are needed. Thanks!

ejeschke:
Looks good!

PR #195, opened 2015-10-27
Highlight active image in Thumbs (Qt)

pllim:
~~Qt~~ solution for `Thumbs` for #186.

~~If we want to be even more flexible, we can make the font color configurable, but will that be overkill?~~

Also see #190

### DIFF
--- a/ginga/Control.py
+++ b/ginga/Control.py
@@ -897,6 +897,7 @@ class GingaControl(Callback.Callbacks):
             self.logger.info("Large image update: %.4f sec" % (
                 split_time1 - start_time))
 
+            self.make_callback('active-image', channel.fitsimage)
 
     def change_channel(self, chname, image=None, raisew=True):
         self.logger.debug("change channel: %s" % (chname))
--- a/ginga/misc/plugins/Thumbs.py
+++ b/ginga/misc/plugins/Thumbs.py
@@ -50,6 +50,7 @@ class Thumbs(GingaPlugin.GlobalPlugin):
                                   sort_order=None,
                                   label_length=25,
                                   label_cutoff='right',
+                                  highlight_tracks_keyboard_focus=False,
                                   label_font_color='black',
                                   label_bg_color='yellow')
         self.settings.load(onError='silent')
@@ -65,6 +66,9 @@ class Thumbs(GingaPlugin.GlobalPlugin):
         # different instruments have different keywords of interest
         self.keywords = self.settings.get('tt_keywords', tt_keywords)
         self.keywords.insert(0, self.settings.get('mouseover_name_key', 'NAME'))
+
+        self.highlight_tracks_keyboard_focus = self.settings.get(
+            'highlight_tracks_keyboard_focus', False)
 
         fv.set_callback('add-image', self.add_image_cb)
         fv.set_callback('remove-image', self.remove_image_cb)
@@ -376,12 +380,18 @@ class Thumbs(GingaPlugin.GlobalPlugin):
         path = image.get('path')
         chname = viewer.get_channelName(fitsimage)
         thumbkey = self.get_thumb_key(chname, name, path)
+        chname = chname.lower()  # for matching with thumbkey
 
         hlight_bg = self.settings.get('label_bg_color', 'yellow')
         hlight_fg = self.settings.get('label_font_color', 'black')
 
         with self.thmblock:
             for tkey in self.thumbDict:
+                # Do not mess with other channels unless asked to
+                if (not self.highlight_tracks_keyboard_focus and
+                        chname != tkey[0]):
+                    continue
+
                 if tkey == thumbkey:
                     bgcolor = hlight_bg
                     fgcolor = hlight_fg
--- a/ginga/misc/plugins/Thumbs.py
+++ b/ginga/misc/plugins/Thumbs.py
@@ -49,7 +49,9 @@ class Thumbs(GingaPlugin.GlobalPlugin):
                                   thumb_length=150,
                                   sort_order=None,
                                   label_length=25,
-                                  label_cutoff='right')
+                                  label_cutoff='right',
+                                  label_font_color='black',
+                                  label_bg_color='yellow')
         self.settings.load(onError='silent')
         # max length of thumb on the long side
         self.thumbWidth = self.settings.get('thumb_length', 150)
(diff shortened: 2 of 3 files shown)


ejeschke:
@pllim, I think that red is not a great default choice for a highlight color.  It makes me feel that there is an error message I should be looking for somewhere.

ejeschke:
Try opening a new workspace with grid of 4 images.  Now you can be displaying 4 images by dragging an image into each one.  With this PR the highlight in `Thumbs` follows the current focus, but it does not highlight all the images that are visible.

This same situation will exist if the user uses a custom layout.  I'm not sure what the exact intention of the PR is, so just checking here.  If highlighting the focused image is sufficient, I think we wouldn't be too far from a merge, but if the intention is to highlight all visible images there is a ways to go.

ejeschke:
My preference would be to change the color of the background of the thumb label and not the font color.

pllim:
@ejeschke 

For me, it makes sense for the highlight to be consistent with the displayed header under "Synopsis". In addition, I want the highlighting to answer the question, "If I do such and such now, which image will be affected?" I tried using local plugins on a workspace but they do not seem to work properly there (and maybe they are not supposed to be used there at all?), so the consistency with "Synopsis" is good enough for now. Does this make sense?

I added user preferences so both label background and font colors can be configured, as shown in this updated PR.

ejeschke:
@pllim, can you rebase against master and make sure this PR still works?

pllim:
Rebased. Still works.

ejeschke:
@pllim, sorry to drag this PR out, but I'm in the process of unifying Thumbs, Contents and Catalogs, so I appreciate your patience and incremental rebase for a successful outcome.

`Thumbs` is now a unified (wrapped) plugin.  Can you rebase this PR against master again?  qt and gtk specific versions are gone, so the entire plugin lives in `ginga.misc.plugins.Thumbs`.  You can use the `set_color(fg=, bg=)` method of a label to set the foreground and background colors of the label.

pllim:
@ejeschke , I rebased. The changes are consistent with what I added before, but now with this PR, I cannot drag and drop existing image into new channel anymore. I am not sure why that is the case. Any idea?

ejeschke:
Drag and drop is temporarily lost with this update, but all other functionality should be there. Did you use that feature much? It was fairly new.  I am thinking about a different way to organize channels.

pllim:
@ejeschke, I just want to make sure that this PR didn't break anything. Other than the drag-drop thing, the highlighting still works as before. I think this is ready for review.

pllim:
@ejeschke , I think once we get the `Contents` sorted out, the callbacks behavior here should be changed to be consistent.

ejeschke:
Agreed.

pllim:
@ejeschke , the highlighting behavior now should be consistent with `Contents`. It is rebased and ready for review.

pllim:
Any chance we can resolve this soon?

ejeschke:
@pllim, I would like to resolve it.  It is a bit tied in with how channels are navigated, so any change to that should be reflected in the highlighting.  Can you try this against PR #226?  Since that PR adds the alphabetical sorting and image history (beyond datasrc) to the channels that you folks requested in other issues, and since `Thumbs` should highlight based on the channel navigation it would be good to finalize the channel navigation/history first.

ejeschke:
Also, I would be interested in whether you feel the highlighting is correct with any viewer configuration that exposes more than one image to the user.  For example, if you open a new workspace in MDI mode and have several channels visible there (or open a grid workspace), does your highlighting show you what you expect to see?

pllim:
@ejeschke 

>>  Can you try this against PR #226?

> No problem found.

>> highlighting is correct with any viewer configuration (?)

> Yes.

If possible, I would like this reviewed and merged as soon as possible. I find that I am spending more time than I like rebasing and resolving conflicts with every Ginga master updates. Thanks.

ginga/Control.py
```diff
@@ -897,6 +897,7 @@ def channel_image_updated(self, channel, image):
             self.logger.info("Large image update: %.4f sec" % (
                 split_time1 - start_time))
 
+            self.make_callback('active-image', channel.fitsimage)
```
ejeschke: @pllim, this callback is not for when an image is set in a channel (the closest one for that is "image-set", available on the channel viewer).  This callback (perhaps unfortunately named) is for when the user changes the channel, even to one where there is an old image present.  Perhaps we should rename this callback.

ginga/misc/plugins/Thumbs.py
```diff
@@ -331,11 +337,13 @@ def focus_cb(self, viewer, fitsimage):
             # No memory of this thumbnail, so regenerate it
             chname = viewer.get_channelName(fitsimage)
             self._add_image(viewer, chname, image)
+            self.highlight_thumbnail(viewer, fitsimage)
             return
 
         # Else schedule an update of the thumbnail for changes to
         # cut levels, etc.
         self.redo_delay(fitsimage)
```
ejeschke:
```python
if self.highlight_tracks_keyboard_focus:
    self.highlight_thumbnail(viewer, fitsimage)
```

This is a callback for keyboard focus.  So if you are wanting highlight based on that, we check preference and if set, highlight.

ginga/Control.py
ejeschke: In any case, it should not be invoked here.

ginga/misc/plugins/Thumbs.py
```diff
@@ -365,6 +373,34 @@ def copy_attrs(self, fitsimage):
...
+        name = image.get('name')
+        path = image.get('path')
+        chname = viewer.get_channelName(fitsimage)
+        thumbkey = self.get_thumb_key(chname, name, path)
+        chname = chname.lower()  # for matching with thumbkey
+
+        hlight_bg = self.settings.get('label_bg_color', 'yellow')
+        hlight_fg = self.settings.get('label_font_color', 'black')
+
+        with self.thmblock:
```
ejeschke: This looks like it is iterating over the entire set of thumbnails every time the keyboard focus changes from one widget to another.  At our telescope in the course of a night there are hundreds of thumbs present, and people are changing widgets (tabs) constantly.  I think this is not an efficient way to go about this.

Why don't we keep a set of thumbkeys (one set per channel).  In each set are the keys for the currently highlighted thumbs. When this function is called it is called with the new set of keys that should be highlighted (this is figured out elsewhere, according to the highlight scheme).  All this function has to do is take the difference of the sets (in two ways) to figure out which ones to unhighlight and which ones to highlight.

For example:
```python
In [5]: foo = set(['a', 'b', 'c', 'd'])

In [6]: goo = set(['b', 'c', 'd', 'e'])

In [7]: foo - goo
Out[7]: {'a'}

In [8]: goo - foo
Out[8]: {'e'}
```
If `foo` is the old set of highlighted thumbkeys, and `goo` is the new set, then (`foo` - `goo`) are the ones to *un*highlight, and (`goo` - `foo`) are the ones we need to highlight.

ejeschke:
Seemed simplest to simply merge and then modify the PR for the efficiency improvements.  Hope you don't mind, @pllim.  I've set the default for `highlight_tracks_keyboard_focus` to be True.  We'll do something similar in `Contents` so that they match.

Please check that this works according to your understanding.

pllim:
@ejeschke , yes, it works. Thanks!

Issue #189, opened 2015-10-26
Header plugin sorting is not working for me


hcferguson:
The manual says: "Clicking on a column header will sort the table by values in that column, which may be useful for quickly locating a particular keyword." 

This isn't working for me (running using Qt on a Mac.)

pllim:
Did you check the "Sortable" checkbox at the bottom left? Just wanting to make sure.

ejeschke:
Works for me on your huge header example if I click the `Sortable` box in lower left.  Maybe I should mark this as a documentation fix?

hcferguson:
Yes. I didn't notice the checkbox. It works for me when I click ```Sortable```.

ejeschke:
Maybe we should make this a stored preference?  Currently you have to check the box whenever you start up ginga.  It persist for the entire session, but not between sessions.

hcferguson:
Seems sensible. Or just make it sortable by default?

ejeschke:
Hmm, when I make it sortable by default, it sorts the header (as expected).  I think for many of our users (at least), they are familiar with the order in which the instrument wrote the file, and having ginga sort it by default might be confusing.

pllim:
I agree with @ejeschke that sorting by default is not desirable. I am one of those people who already familiar with the header layout and wish to keep it that way.

hcferguson:
Agreed that I would want to have the first view be in the order they appear in the header. But more intuitive to me would be to have it so that  just clicking on a column sorts by that column without having to click another button to enable that functionality. And then I would want the option to put it back in the original order (i.e. rename the Sortable button something like "reset" and have it undo the sort, if that's feasible).

pllim:
Currently, unchecking `Sortable` undoes the sorting. I think for now, it is sufficient to make the checkbox a preference.

ejeschke:
Merged #196.  Leaving this issue open until the documentation is updated.

ejeschke:
I believe the documentation is addressed in commit 10fe0ceb9bb051d0832915b6f21ef6b73d87c43c.  If there is no objection, I will close this issue.

pllim:
@ejeschke, there appears to be a broken reference in the sentence just above your commit.

PR #196, opened 2015-10-27
Added preference to toggle header sort on load

pllim:
Added preference to toggle header sort on load. This is based on a request in #189. c/c @hcferguson

@ejeschke, I don't have GTK but theoretically it should work on both Qt and GTK. Can you please double check? Thanks!


ejeschke:
Well done, @pllim!  I had to fix one small thing.

pllim:
:smile:

Issue #201, opened 2015-10-28
How to directly open multi-ext FITS with non-default ext?


pllim:
This was brought up during a sprint involving @hcferguson, @gderosa2004, @eteq, @perrygreenfield, @stscieisenhamer, and myself.

Let's say we have an image named `image01.fits` with the following multi-extension FITS layout:

Ext. Num. | EXTNAME | EXTVER
------ | ------ | ------
0 | PRIMARY | --
1 | SCI | 1
2 | ERR | 1
3 | DQ | 1
4 | SCI | 2
5 | ERR | 2
6 | DQ | 2

And the user wishes to open extension 4 directly. With DS9, there are two ways to do this (that I actually used, could  be more):

* Via PyRAF/IRAF: 
```
--> !ds9 &
--> display image01.fits[4] 1
```

* Via DS9 GUI:
```
File -> Open -> Click on the filename -> Manually append "[4]" to the name -> OK
```

I have yet found such shortcut in Ginga without having to load the default extension first (e.g., `[SCI,1]`) and then use the `MultiDim` plugin. I tried passing in `image01.fits[4]` via command line, using `File -> Open`, and `FBrowser`; None of those let me open extension 4 directly.

ejeschke:
Did you try it from the command line?  e.g. 
```bash
$ ginga foo.fits[1]
```

pllim:
Yes, I did. I get an error:
```
ginga: No match.
```

ejeschke:
I was able to do it by quoting the brackets, because depending on your shell (e.g. `bash`) they are special characters.  For example, if I do:
```bash
$ ginga .../testdata/HST/"iaby02x7q_flt.fits[4]"
```
it does just what you would expect.  You can probably also use backslash to escape the brackets.

ejeschke:
The `load_file` method in ginga does parse the HDU extension syntax, so it's simply a matter of getting the path with the "[...]" bits on the end in to that function.   I think it is also supported by drag and drop if the URL happens to have the syntax at the end.

Probably we should modify the file open dialog and the `FBrowser` plugin to allow you to specify the HDU extension.

pllim:
Ah, okay. I can confirm that this also works:
```
ginga "image01.fits[4]"
ginga image01.fits\[4\]
```

Yes, would be nice for `FBrowser` to support this, if possible.

ejeschke:
Remote control also works.  Be sure to invoke the "RC" plugin from the "Plugins" menu, then:
[...]

PR #204, opened 2015-10-29
Improve FBrowser and File menu (Qt)

pllim:
This addresses #201 for Qt.

Improved `FBrowser` plugin:

* Removed unused `FBrowserBase.py`.
* If user set invalid home directory in `plugin_FBrowser.cfg`, it defaults to internal setting instead of throwing error.
* Added `open_files()` method. 
* Load button can now open directory if a single sub-directory is selected in the tree view, instead of throwing error.
* User can now enter their own filter in entry text box and "Load" will open all matching files with the correct extension, if given. Example: `/my/path/*.fits`, `/my/path/*/*.fits`, `/my/path/*.fits[4]`, `/my/path/*.fits[SCI,2]`, etc.
* Pressing "Enter" on the text box will result in similar behavior as "Load".
* "Save" button to report error to log file when there is no image, instead of throwing error on GUI.

Improved `GingaQt` File menu (Qt only):

* File menu now accepts wildcard and extension.
* Only opens first match if wildcard returns multiple matches (otherwise, I get `QPainter` errors on the terminal).

Modified `AstroImage` and FITS I/O:

* Consistent naming of cache no matter how user provided extension info (int, str, or tuple). For example, `image.fits[1]`, `image.fits[SCI]`, and `image.fits[SCI,1]` will all result in the same cache name of `image[SCI,1]`.

For GTK version, shouldn't be too hard for you to add the similar functionality given the diff in this PR, right, @ejeschke ?


pllim:
A side effect is that if user enters `[3]`, it does not convert it to, say, `[DQ,1]`. I will look more into this.

ejeschke:
@pllim, that is because if the users specifies the hdu the opener tries to use that specification.  The thinking goes: "if the user specified `[1]` then when that shows up in the viewer it makes sense. If they specified `[5]` and `[DQ, 2]` shows up in the viewer, they may be confused.".  If the user knows they want `[DQ, 1]`, I think they should specify that.

pllim:
@ejeschke , that complicates things when I have plugins that automatically loads, say, the corresponding `DQ` to the active `SCI`. I can look for all the possible extension naming, but it is easier if the cache has a consistent way to name the extensions.

pllim:
@ejeschke, I think this is ready for review. Please let me know what you think.

pllim:
Rebased to the latest master.

ejeschke:
Sorry, @pllim can you rebase once more against unified plugin? I can then look more closely at this.

pllim:
@ejeschke , before I proceed, I notice that in the latest dev version, `FBrowser` file ordering now appears random (screenshot attached). The most obvious is that the "one level up" (`..`) symbol that is supposed to be listed first, is not anymore.

untitled

ejeschke:
Sorry, I had the sort set on the wrong column.  Fixed in 58b4a6c4c9936b62fb67413b324f32f4f5355bd4.

pllim:
@ejeschke , rebased and appears to work (in Qt anyway). This is ready for review.

ginga/AstroImage.py
```diff
@@ -139,8 +139,7 @@ def load_file(self, filepath, numhdu=None, naxispath=None):
             if name is None:
                 name = info.name
                 if ('[' not in name):
-                    if numhdu is None:
-                        numhdu = numhdu_
+                    numhdu = numhdu_
```
ejeschke: This is the only change that worries me. Because you are overriding the `numhdu` that the user specified.  Since this is an important method that can be called from many places, the user may want to force the use of a specific `numhdu`.

To preserve the possibility of forcing a specific `numhdu`, but defaulting to an override, how about adding another kwarg parameter `allow_numhdu_override` which defaults to `True`.  Then the logic looks like:
```python
if (numhdu is None) and allow_numhdu_override:
    numhdu = numhdu_
```
That way it will work in the default with the reference viewer, but if someone has a special use case with their custom viewer they can always force the `numhdu`.

ejeschke:
Otherwise looks ok to me.  If you are ok with the change I mentioned, I can add that and the gtk bits in the PR merge.

BTW, in testing I noticed that the "File"->"Load Image" dialog does not allow multiple selection.  I rarely use this dialog since I prefer `FBrowser`, but it seems like it should support multiple selection.  Wondering if we should open up another issue or PR on that?

ginga/AstroImage.py
pllim: Sounds, good. Except I think the logic should be `or`, not `and`.

pllim:
Regarding `File -> Load Image` loading only one image even if there are multiple matches using wildcard, it is done on purpose for the following reasons:

1. If I open all of them, I get `QPainter` errors using Qt. (No testing done with GTK.)
2. "Load Image" is singular, hence suggests it should only load one.

:+1: to addressing this particular feature on a separate issue.

pllim:
Thanks, @ejeschke . I just want to point out again that this PR did not address GTK version.

Issue #171, opened 2015-10-07
Add channel_Image.cfg to examples


pllim:
Add `channel_Image.cfg` to `examples/config`.

ejeschke:
Hmm.  Not opposed to this, but what would we put in it?  It's basically a dump of the `Preferences` for the generic, starting channel.  As such, it could be configured a lot of ways, and will get overwritten if the user ever saves preferences for that channel.

Are you thinking just to document what settings affect what things?

pllim:
Yes, it's more for documentation. I opened up my `channel_Image.cfg` and not sure what is what for some of the entries.

pllim:
Here is a dump out when I save my channel preference (I think they should mostly be default values):
```python
auto_orient = True
autocenter = 'on'
autocut_method = 'zscale'
autocut_params = {}
autocuts = 'override'
autozoom = 'on'
color_algorithm = 'linear'
color_hashsize = 65535
color_map = 'gray'
cuts = (0.0, 0.0)
defer_lagtime = 0.025
defer_redraw = True
flip_x = False
flip_y = False
follow_focus = True
genthumb = True
icc_black_point_compensation = False
icc_output_intent = 'perceptual'
icc_output_profile = None
icc_proof_intent = 'perceptual'
icc_proof_profile = None
image_overlays = True
intensity_map = 'ramp'
interpolation = 'basic'
numImages = 10
onscreen_ff = 'Sans Serif'
output_icc = None
pan = (1.0, 1.0)
pan_coord = 'data'
profile_use_cuts = False
profile_use_pan = False
profile_use_rotation = False
profile_use_scale = False
profile_use_transform = False
raisenew = True
rot_deg = 0.0
scale = (1.0, 1.0)
scale_max = 10000.0
scale_min = 1e-05
scale_x_base = 1.0
scale_y_base = 1.0
show_pan_position = False
swap_xy = False
switchnew = True
wcs_coords = 'icrs'
wcs_display = 'sexagesimal'
zoom_algorithm = 'step'
zoom_rate = 1.4142135623730951
zoomlevel = 1.0
```

Would be nice to:

1. Group them into relevant categories and insert useful comments.
2. Mention which ones can also be inherited from `general.cfg` (e.g., `numImages`).

pllim:
Resolved by #211.

PR #211, opened 2015-11-02
Added channel_Image.cfg example

pllim:
This addresses #171.

* Added channel config file example.
* Minor verbiage change in general config to be consistent with the new example.

@ejeschke, my organization of the config file is mainly educated guesses, so please double check. I also added notes below on the entries that I am really not sure about.

### DIFF
--- a/ginga/examples/configs/channel_Image.cfg
+++ b/ginga/examples/configs/channel_Image.cfg
@@ -105,7 +105,6 @@ icc_output_intent = 'perceptual'
 icc_proof_profile = None
 icc_proof_intent = 'perceptual'
 icc_black_point_compensation = False
-output_icc = None
 
 # ---------------
 # Miscellaneous
--- a/ginga/examples/configs/channel_Image.cfg
+++ b/ginga/examples/configs/channel_Image.cfg
@@ -19,14 +19,29 @@ color_algorithm = 'linear'
 # Color Mapping
 
 color_map = 'gray'
-color_hashsize = 65535
 intensity_map = 'ramp'
+
+# In most cases, with the default value of the hash size as high as it is and
+# the limitation of current displays, there is no need for the user to set this
+# value.
+#
+# As the hash size grows smaller, and depending on the values in the image, the
+# cut levels, and the color distribution algorithm (linear, log, etc.), distinct
+# image values will be hashed into the same buckets, so subtle gradations could
+# show up as posterization in the image, especially with the standard rgbmapper
+# mapping to 24 bpp displays.
+#
+color_hashsize = 65535
 
 # ---------------
 # Auto Cuts
-
+#
+# Stores parameters specific to the chosen autocut_method. For example:
+#   autocut_method = 'histogram'
+#   autocut_params = [('usecrop', True), ('numbins', 2048), ('pct', 0.993)]
+#
 autocut_method = 'zscale'
-autocut_params = {}
+autocut_params = []
 cuts = (0.0, 0.0)
 
 # ---------------
@@ -58,7 +73,10 @@ interpolation = 'basic'
 
 # ---------------
 # Panning
-
+#
+# If pan_coord is set to 'wcs', the pan position cannot exist in sexagesimal
+# format in the config file.
+#
 pan = (1.0, 1.0)
 pan_coord = 'data'
 show_pan_position = False
@@ -81,12 +99,8 @@ genthumb = True
 numImages = 10
 
 # Channel focus follows window focus
-# Same as channelFollowsFocus in general.cfg???
+# Same as channel_follows_focus in general.cfg
 follow_focus = True
-
-# Preference for fixed and sans fonts
-# Same as sansFont in general.cfg???
-onscreen_ff = 'Sans Serif'
 
 # ---------------
 # Remember - When set to True, GUI remembers the settings for each image
@@ -109,7 +123,16 @@ icc_black_point_compensation = False
 # ---------------
 # Miscellaneous
 
+# Preference for fixed and sans fonts
+# Font used for onscreen messages (zoom, etc.)
+onscreen_ff = 'Sans Serif'
+
+# Ensure that RGB images are properly flipped in Y on load
+# (e.g., a JPEG image). Not applicable to FITS.
 auto_orient = True
+
+# Performance will be terrible if set to False, but doing so is useful
+# for debugging the core widget and some special backends.
 defer_redraw = True
 defer_lagtime = 0.025
 
--- a/ginga/examples/configs/channel_Image.cfg
+++ b/ginga/examples/configs/channel_Image.cfg
@@ -1,0 +1,118 @@
+#
+# Image channel preferences file
+#
+# Place this in file under ~/.ginga with the name "channel_Image.cfg" for
+# channel name "Image". This file WILL BE OVERWRITTEN by the Preferences
+# plugin (see below). NOT ALL PARAMETERS IN GUI CAN BE ADJUSTED HERE.
+#
+# Alternately, the settings can be adjusted via Preferences local plugin
+# within Ginga. NOT ALL PARAMETERS CAN BE ADJUSTED VIA GUI but users can
+# normally ignore the hidden ones.
+#
+
+# ---------------
+# Color Distribution
+
+color_algorithm = 'linear'
+
+# ---------------
+# Color Mapping
+
+color_map = 'gray'
+color_hashsize = 65535
+intensity_map = 'ramp'
+
+# ---------------
+# Auto Cuts
+
+autocut_method = 'zscale'
+autocut_params = {}
+cuts = (0.0, 0.0)
+
+# ---------------
+# Transform
+
+flip_x = False
+flip_y = False
+swap_xy = False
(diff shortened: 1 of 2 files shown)


ginga/examples/configs/channel_Image.cfg
```diff
@@ -0,0 +1,118 @@
+#
+# Image channel preferences file
+#
+# Place this in file under ~/.ginga with the name "channel_Image.cfg" for
+# channel name "Image". This file WILL BE OVERWRITTEN by the Preferences
+# plugin (see below). NOT ALL PARAMETERS IN GUI CAN BE ADJUSTED HERE.
```
pllim: It is a bit confusing to me that this config file does not address everything in `Preferences` and vice versa. So I added notes here in the comments. Please correct me if I misunderstood.

ginga/examples/configs/channel_Image.cfg
```diff
@@ -0,0 +1,118 @@
...
+
+color_map = 'gray'
+color_hashsize = 65535
+intensity_map = 'ramp'
+
+# ---------------
+# Auto Cuts
+
+autocut_method = 'zscale'
+autocut_params = {}
```
pllim: Not very sure what this is for but it shows up when I click Save.

ginga/examples/configs/channel_Image.cfg
```diff
@@ -0,0 +1,118 @@
...
+scale = (1.0, 1.0)
+scale_min = 1e-05
+scale_max = 10000.0
+interpolation = 'basic'
+
+# ---------------
+# Panning
+
+pan = (1.0, 1.0)
+pan_coord = 'data'
```
pllim: Cannot find the setting that corresponds to the "WCS sexagesimal" checkbox. Did I miss something?

ginga/examples/configs/channel_Image.cfg
```diff
@@ -0,0 +1,118 @@
...
+# ---------------
+# General - Inherits from general.cfg if not specified, otherwise overwrite.
+
+# Number of images to keep in memory per channel (0 = unlimited)
+# Same as numImages in general.cfg
+numImages = 10
+
+# Channel focus follows window focus
+# Same as channelFollowsFocus in general.cfg???
+follow_focus = True
```
pllim: Really unsure about this one.

ginga/examples/configs/channel_Image.cfg
```diff
@@ -0,0 +1,118 @@
...
+
+# Preference for fixed and sans fonts
+# Same as sansFont in general.cfg???
+onscreen_ff = 'Sans Serif'
```
pllim: Really unsure about this one as well.

ginga/examples/configs/channel_Image.cfg
```diff
@@ -0,0 +1,118 @@
...
+
+# ---------------
+# ICC Profiles
+
+icc_output_profile = None
+icc_output_intent = 'perceptual'
+icc_proof_profile = None
+icc_proof_intent = 'perceptual'
+icc_black_point_compensation = False
+output_icc = None
```
pllim: This one does not seem to be settable from `Preferences`?

ginga/examples/configs/channel_Image.cfg
```diff
@@ -0,0 +1,118 @@
...
+# ---------------
+# Color Distribution
+
+color_algorithm = 'linear'
+
+# ---------------
+# Color Mapping
+
+color_map = 'gray'
+color_hashsize = 65535
```
pllim: Why would there be a need to change the hash size?

ginga/examples/configs/channel_Image.cfg
```diff
@@ -0,0 +1,118 @@
...
+icc_output_intent = 'perceptual'
+icc_proof_profile = None
+icc_proof_intent = 'perceptual'
+icc_black_point_compensation = False
+output_icc = None
+
+# ---------------
+# Miscellaneous
+
+auto_orient = True
```
pllim: Should I really put this under "New Images" section?

ginga/examples/configs/channel_Image.cfg
```diff
@@ -0,0 +1,118 @@
...
+defer_redraw = True
```
pllim: What is this setting used for? To increase performance?

ginga/examples/configs/channel_Image.cfg
ejeschke: These two items control which autocut algorithm is used and what the control parameters are.

ginga/examples/configs/channel_Image.cfg
ejeschke: I think the "WCS sexagesimal" box just controls how the pan wcs is formatted in the `Preferences` plugin.  It's not actually a channel setting.  It could be made into a `plugin_Preferences.cfg` preference though.

ginga/examples/configs/channel_Image.cfg
ejeschke: If `follow_focus` is True then the channel will be changed when the mouse cursor enters the channel window.  I'm pretty sure this is passed on from the `general.cfg` setting for `channel_follows_focus`.  You won't actually notice this effect unless you use gridded workspaces or alternate viewer layouts, because clicking a tab also changes the focus.

ginga/examples/configs/channel_Image.cfg
ejeschke: Font used for onscreen messages (zoom, etc.)

ginga/examples/configs/channel_Image.cfg
ejeschke: No need in the usual case.

ginga/examples/configs/channel_Image.cfg
ejeschke: There is no GUI element in preferences for this, and it does nothing currently for `AstroImage` objects, I think.  It ensures that RGB images are properly flipped in Y when you load one of those (e.g. a JPEG image).

ginga/examples/configs/channel_Image.cfg
ejeschke: Yes, very important.  Performance will be terrible if turned off.  But useful for some debugging of the core widget and some special back ends.

ginga/examples/configs/channel_Image.cfg
pllim: Can you please give an example of an unusual case?

ginga/examples/configs/channel_Image.cfg
pllim: Can you please give an example on how one would define the dictionary?

ginga/examples/configs/channel_Image.cfg
pllim: I still find this a bit confusing. Because here, the user can theoretically set `pan_coord = 'wcs'` but there is no way to use this config file to say whether `pan` in WCS mode is sexagesimal or not.

ginga/examples/configs/channel_Image.cfg
pllim: Is it `channelFollowsFocus` or `channel_follows_focus`? Does this mean `general.cfg` need updating too?

ginga/examples/configs/channel_Image.cfg
ejeschke: The pan position cannot exist in sexagesimal form internally.  The checkbox only effects how it is displayed in the the `Preferences` plugin.  So there is no option for this in an `ImageViewer`-based object.

ginga/examples/configs/channel_Image.cfg
ejeschke: yes, looks like `general.cfg` needs updating.  It's `channel_follows_focus`.

ginga/examples/configs/channel_Image.cfg
ejeschke: Hmm, I just checked and actually this is supposed to be a list of tuples.  Example might be:
```python
[('usecrop', True), ('numbins', 2048), ('pct', 0.993)]
```
if `autocut_method` is set to 'histogram'.

I think there is a bug in restoring the params from the saved settings--looking into that.

ginga/examples/configs/channel_Image.cfg
ejeschke: As the hash size grows smaller, and depending on the values in the image, the cut levels and the color distribution algorithm (linear, log, etc), distinct image values will be hashed into the same buckets, so subtle gradations could show up as posterization in the image, especially with the standard rgbmapper mapping to 24 bpp displays.  

Practically speaking, with the default value of the hash size as high as it is and the limitation of current displays there is no need for the user to set the value.  That is why I removed it from the `Preferences` plugin.

ejeschke:
@pllim, let me know when you are done with any changes to this PR and I'll merge it

ginga/examples/configs/channel_Image.cfg
pllim: Okay, I'll change it to empty list then in the config.

pllim:
@ejeschke, I'm done. Feel free to merge if you are okay with this. Thank you!

Issue #215, opened 2015-11-06
Add "Modified" column to Contents


pllim:
I see that Ginga as a `'modified'` callback. I propose that `Contents` adds a new column at the end named "Modified", which would change from "no" to "yes" when an image issues the aforementioned callback. What do you think, @ejeschke? If you agree, I can attempt a PR since I already have something similar going on in my subclass. Please let me know.

ejeschke:
There are more than one `modified` callbacks.  There is one issued by an `AstroImage` class object to indicate that the file has been modified in memory (but not on disk).  This is not typically done by most plugins that I know of.  

There is a `modified` issued by a canvas to indicate that a redraw is necessary for any viewers looking at that canvas.

Are you thinking of the first case?  Do you have a plugin that modifies the image's data and resets it using `set_data`?

pllim:
> the file has been modified in memory (but not on disk)

Yes, that one.

> Do you have a plugin that modifies the image's data

Yes, I have multiple.

pllim:
Currently, I manually set the status by directly accessing the subclassed `Contents` plugin method. But since you already have such a callback, I thought it would be more elegant to use the callback. And there is no reason not to include in core to show that Ginga already support such a thing?

ejeschke:
I have no objections.  I would keep in mind that 1) it is possible for an image to be in more than one channel at a time, and 2) image names are not necessarily unique.  AFAIK, the `modified` callback comes from the image directly, with no information about which channel it is in, so that might complicate figuring out which contents entries to modify.

How about instead adding a modification time (`time.time()`) to the image object itself and showing that in the `Contents`? That certainly would be easier, and very accurate.

pllim:
Hmm. Modification time tells me nothing unless I can somehow memorize all the original mod times and mentally compute whether it has changed. However, if you are saying that we can use modification time as a basis to decide whether the status is "yes" or "no", then it works for me.

ejeschke:
If you set the initial modification time of the image to `None`, then in the column you could show "no" or if modified, the time.

pllim:
Yes, that sounds great! :+1:

pllim:
@ejeschke , I started a prototype at #223 but it needs some help.

pllim:
This is resolved.

PR #223, opened 2015-11-12
Add new MODIFIED column in Contents

pllim:
Proposed solution for #215.

As @ejeschke said, "AFAIK, the modified callback comes from the image directly, with no information about which channel it is in, so that might complicate figuring out which contents entries to modify." ~~That is something I am not sure how to address yet.~~

Changes in `Control.py`:

* Added new `image-modified` callback, as @ejeschke suggested.
* Added `Channel.image_data_modified()` method to handle this new callback. (L1482, L1687)
* ~~Added new `MODIFIED` keyword to image header to keep track of modified timestamp. This is set to `None` if not modified yet. (L871)~~
* Fixed bug in `advertise_image()`. (L879)
* Auto-close local plugins when channel is deleted.
* Added check to prevent deleting a channel when there is no channel left.
* PEP8 stuff, some doc fixes.

Changes to `Contents.py`:

* Added new MODIFIED column to display modified timestamp, if any.
* Added `image_modified_cb()` method to handle the new `image-modified` callback.
* Change `add-image-info` to use `add_callback()` (see comments in 17c6e7d3f).

`plugin_Contents.cfg` -- Added the new MODIFIED column in the settings.

### DIFF
--- a/ginga/Control.py
+++ b/ginga/Control.py
@@ -807,6 +807,18 @@ class GingaControl(Callback.Callbacks):
                 self.logger.error("Failed to continue operation: %s" % (str(e)))
                 # TODO: log traceback?
 
+    def _close_plugins(self, channel):
+        """Close all plugins associated with the channel."""
+        opmon = channel.opmon
+        for key in opmon.get_active():
+            obj = opmon.getPlugin(key)
+            try:
+                self.gui_do(obj.close)
+
+            except Exception as e:
+                self.logger.error("Failed to continue operation: %s" % (str(e)))
+                # TODO: log traceback?
+
     def channel_image_updated(self, channel, image):
 
         with self.lock:
@@ -1009,17 +1021,24 @@ class GingaControl(Callback.Callbacks):
         return channel
 
     def delete_channel(self, chname):
+        """Delete a given channel from viewer."""
         name = chname.lower()
-        # TODO: need to close plugins open on this channel
+
+        if len(self.channelNames) < 1:
+            self.logger.error('Delete channel={0} failed. '
+                              'No channels left.'.format(chname))
+            return
 
         with self.lock:
             channel = self.channel[name]
+
+            # Close local plugins open on this channel
+            self._close_plugins(channel)
 
             # Update the channels control
             self.channelNames.remove(chname)
             self.channelNames.sort()
             self.w.channel.delete_alpha(chname)
-
             self.ds.remove_tab(chname)
             del self.channel[name]
             self.prefs.remove_settings('channel_'+chname)
--- a/ginga/Control.py
+++ b/ginga/Control.py
@@ -492,11 +492,6 @@ class GingaControl(Callback.Callbacks):
             #channel.fitsimage.onscreen_message("Failed to load file", delay=1.0)
             raise ControlError(errmsg)
 
-        # Initialize MODIFIED header keyword for image-modified callback.
-        # This is for consistent keyword display in Contents, and used in
-        # 'image-modified' callback.
-        image.metadata['header']['MODIFIED'] = None
-
         self.logger.debug("Successfully loaded file into image object.")
         return image
 
@@ -758,11 +753,6 @@ class GingaControl(Callback.Callbacks):
                 raise ValueError("Need to provide a channel name to add "
                                  "the image")
             chname = channel.name
-
-        # Initialize MODIFIED header keyword for image-modified callback.
-        # This is for consistent keyword display in Contents, and used in
-        # 'image-modified' callback.
-        image.metadata['header']['MODIFIED'] = None
 
         # add image to named channel
         channel = self.get_channel_on_demand(chname)
@@ -1366,15 +1356,12 @@ class Channel(Callback.Callbacks):
             if channel != self:
                 self.fv.change_channel(self.name)
 
-    def image_data_modified(self, image, reset=False):
+    def image_data_modified(self, image, reason='', reset=False):
         """Call this when data in buffer is modified.
 
-        Header metadata, ``MODIFIED``, will be updated with UTC timestamp info.
-        Plugins with ``redo()`` methods will pick this up.
+         **Callbacks**
 
-        **Callbacks**
-
-        Both ``'image-modified'`` and ``'modified'`` callbacks will be issued.
+        The ``'image-modified'`` callback will be issued.
         To use this from a plugin:
 
         .. code-block:: python
@@ -1382,15 +1369,24 @@ class Channel(Callback.Callbacks):
             image = self.fitsimage.get_image()
             data = image.get_data()
             new_data = do_something(data)
+
+            # This issues a 'modified' callback, which calls redo()
             image.set_data(new_data, metadata=image.metadata)
+
             chname = self.fv.get_channelName(self.fitsimage)
             channel = self.fv.get_channelInfo(chname)
-            channel.image_data_modified()
+
+            # This issues a 'image-modified' callback, which sets
+            # the UTC timestamp and reason
+            channel.image_data_modified(image, reason='Did something')
 
         Parameters
         ----------
         image
             Image object to update.
+
+        reason : str
+            Reason for modification.
 
         reset : bool
             Set to `True` on init or if image fell out of cache.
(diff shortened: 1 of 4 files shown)


pllim:
@ejeschke , if I want to change the `modified` callback to accept channel info, not just the image, where to start?

ejeschke:
Hmm...I think the thing to do is to use a chained callback.  When an image is modified there is no context necessarily about the channel it's in.  Probably the place to do this is in `Control` where an image is added to a channel--perhaps even in the new `Channel` class there. You could add a new `image-modified` callback.  When an image is added to a channel it should register for the `modified` callback on the image.  In *it's* callback handler, it can then make its own channel callback, for `image-modified`, including the `image` object in the callback (which will also include itself (channel object) by default).  In `Contents`'s new channel callback it can register on that channel for `image-modified`.  It should then get a callback for every channel containing the image which gets modified.

One detail: a channel should check during it's callback handling of `modified` whether the image is actually in its own history (it may have been there initially, but was moved to a different channel subsequently).  Only if it is in it's own history should it make the chained callback.

pllim:
Just brainstorming here...

Another alternative could be that we make image name in `Datasrc` (i.e., the lookup key) unique across the board, not just per channel. This can possibly be achieved by attaching the current channel name to the image name to construct the key. So, `fileprefix001[SCI,1]` becomes something like `channelname_fileprefix001[SCI,1]`.

Doing this means, when image is moved into a different channel (say, from `Image` to `Image0`), its key changes, say, from `image_myfile001[SCI,1]` to `image0_myfile001[SCI,1]`. All the usual mechanisms still apply (check for duplicate, reload if needed, etc). This way, from the key alone, we can immediately identify both the channel and the image affected in `modified` callback (and all the other callbacks as well).

This would make more sense, no? What do you think, @ejeschke ?

pllim:
Or we can also consider attaching channel name to its metadata instead? Although, I think putting it in the key is more definitive.

pllim:
@ejeschke , I updated the PR to use the new callback, as you suggested. I tested it against spacetelescope/stginga#61 and it works for me.

I did not test the its behavior for when an image is copied or movied to another channel, because I don't know how to do that since drag-and-drop no longer works.

Would like to get this resolved soon if possible, since the changes can be complicated to rebase.

ginga/Control.py
```diff
@@ -602,20 +606,20 @@ def load_image(self, filepath, idx=None):
...
+        filespec : str
+            The path of the file to load (can be a URL).
+
+        dldir
 
         Returns
         -------
-        res:
-            a Bunch object
+        res : `~ginga.misc.bunch.Bunch`
```
ejeschke: I think this should be `ginga.misc.Bunch.Bunch` (caps typo--B).

pllim: Good catch, @ejeschke . It is fixed. Thanks!

pllim:
In the current state of this PR, when an image buffer is modified and then reloaded from file after it fell out of `Datasrc`, its modifed status does not change from modified timestamp back to None until it is in focus again (when Ginga actually looks for it and attempts to reload). 

I thought about having `Contents` look at *all* the latest timestamps (or lack of them) each time an image is clicked, which would give more accurate modified status info, but also a big hit on its performance. This is not an issue if user sets `numImages=0` (unlimited cache size). FYI.

ginga/Control.py
```diff
@@ -1419,6 +1479,52 @@ def _add_image_update(self, image, info):
...
+        ----------
+        image
+            Image object to update.
+
+        reset : bool
+            Set to `True` on init or if image fell out of cache.
+            This will set the modified timestamp to `None`.
+
+        """
+        if reset:
```
pllim: @ejeschke , is it worth adding a lock? I don't see a lock in other callback methods. Theoretically, multiple plugins could issue this callback at the same time, but then again, it is quite impossible for the user to use multiple plugins simultaneously, right?

ejeschke:
@pllim, I can appreciate the code "clean ups", but they complicate reviewing a functional change.  I would request that code clean ups be generally done as a separate PR, unless the code in question is part of the functional change.  The one exception might be documentation--I think ginga's documentation needs all the help it can get, so adding/cleaning the docstrings on functions that you are modifying is :+1:

ejeschke:
> In the current state of this PR, when an image buffer is modified and then reloaded from file after it fell out of Datasrc, its modifed status does not change from modified timestamp back to None until it is in focus again (when Ginga actually looks for it and attempts to reload). 

One option would be to also subscribe to the callback for `image-unset` (this is a viewer cb, not a ginga shell cb) and then check whether the image that was just "unset" is still in memory.  That would be a lot more efficient.  

But perhaps the best would be to add callbacks to the `Datasrc` when something gets pushed out it invokes the callback.  Then the `Channel` object can subscribe and do a chained callback.  plugins can subscribe to the chained cb to find out when an image has been removed from memory.  I'll take a look at that because I think that would be the most efficient.  The problem with adding lots of checks to these callbacks in the plugins means that overall image changing gets slower and slower with each one you add.

ginga/Control.py
```diff
@@ -850,16 +866,22 @@ def add_image(self, imname, image, chname=None, silent=False):
...
             if channel is None:
-                raise ValueError("Need to provide a channel name to add the image")
+                raise ValueError("Need to provide a channel name to add "
+                                 "the image")
             chname = channel.name
 
+        # Initialize MODIFIED header keyword for image-modified callback.
+        # This is for consistent keyword display in Contents, and used in
+        # 'image-modified' callback.
+        image.metadata['header']['MODIFIED'] = None
```
ejeschke: First, if this is being added to the FITS header, it would be better to use `image.set_keyword()` which doesn't expose the implementation detail that the FITS header is stored in the metadata "header" attribute--which could change in the future.

Second, this will make MODIFIED be shown in the `Header` plugin.  Is this desirable?  Because it will make users think that it is actually part of the FITS header.

Third, because of #1 and #2, I think it might be better to make 'modified' be a regular metadata attribute that is assigned in the `AstroImage` class (or even `BaseImage`) as `None`.  And it should be updated (set to a `time.time()`) when the `set_data()` method is called in that class.  That class already invokes a `modified` callback.

pllim: If I make it too specific to FITS specific calls, then it might break if JPEG emits the same callback, no? That is the reason I didn't use `set_keyword()`. Good points on the rest, I'll have to think about it more.

ejeschke: That's probably even more reason to use a "regular" metadata element, since all images have that.  Could you try a test modification of your PR with commit 2931482903b4a61708ddd1928d047b8988b5685a?  This adds the "time_modified" metadata item to `BaseImage` and sets it when the image is modified with `set_data()`.  Will that work fairly cleanly with your PR?

pllim: Now I recall why I had to stick MODIFIED in the header. This is so Contents can pick it up for display. Contents plugin assumes all columns info can be extracted from the header.

ejeschke: Well, `name` is not in the FITS header either, but it extracts that.

pllim:
>  I can appreciate the code "clean ups", but they complicate reviewing a functional change.

Sorry! My OCD got the better of me there.

>  I'll take a look at that because I think that would be the most efficient. 

That would be great. Since you already have an idea on how to do it, let's address that as a separate issue.

I'll make some changes to this PR based on your feedback. Thank you, @ejeschke !

pllim:
@ejeschke , my latest commit to this PR (d498f08) should address your concerns. Please let me know what you think!

pllim:
@ejeschke , any chance we can resolve this before the holidays?

ejeschke:
I think we can get this merged in the next couple days.  I'll look at your updates later today.

ejeschke:
@pllim, please *don't* rebase this--yet.

pllim:
Okay, thanks for the warning.

ejeschke:
@pllim, this PR has been merged, with some modifications.  See the commit message in c6c8e772e16fee6d6b9be2acb73463aa6f2ccb93 for the main points.  Can you test with your image-modifying plugins and give feedback?  Very briefly, you will need to comment out your call from the plugin setting the modified status; images already generate a modified callback when `set_data()` or other internal data modifying methods are used, so it is not necessary with this version.  This allows the modification time to be shown, e.g. when you tile each image in a mosaic--the Mosaic plugin knows nothing about this, yet Contents modification time updates properly.

If you really need the "reason" metadata for why an image was modified I would recommend attaching that to the image_info structure (a Bunch) stored in the channel history, which is where the modified time is now stored.  This can be done in the plugin, using the `get_image_info(name)` method in the `Channel` object.

Right now (as in your PR) the modification time should be reset to None when the image is reloaded.  Can you verify?  We should think about whether it should go to None when it actually is unloaded from memory, but that would be a different PR (if it is even necessary).

pllim:
Thanks, @ejeschke ! Since the changes are in master, I am closing this PR. I'll test it and let you know.

PR #226, opened 2015-11-17
(reference viewer) Refactor channels 

ejeschke:
This PR is intended to address several outstanding inter-related issues in Ginga.

Basically, it adds a full history to channels that goes beyond the simple `Datasrc` that is used to hold the current images in memory. This history contains information about every image that has been loaded into a channel, and can be used to recreate images that have been pushed out of the data source.  Furthermore, the history can be sorted.  This facilitates the following:

1) you can now step through all the images in a channel using the channel up/down navigation arrows, not just the ones that are in memory.
2) the images in a channel can be sorted according to different criteria, like alphabetically or by time loaded, depending on how the user wants to see them or move through them. The stepping function obeys this order.
3) it should eventually simplify the `Thumbs` and `Contents` plugins somewhat, which maintain their own histories and indexes, by creating a common index and history that can be shared by plugins
4) it makes it possible to copy or move image information between channels for images that are out of memory.
 
To do this a `Channel` class is added, and it becomes more of a "smart object" than simply a placeholder of attributes.


ejeschke:
@pllim, @stscieisenhamer, would appreciate you testing your existing plugins against this PR.  I think this addresses some of the concerns about stepping through images in a channel and having that be consistent with what shows in the channel in `Contents` and `Thumbs`.  This should eventually simplify those plugins as well.

ejeschke:
To get the alpha sorting behavior, go to `Preferences` for the `Image` channel and in the "General" settings there is now a "Sort Order".  Set this to "alpha", save settings and restart ginga.  Now stepping through images in a channel using up/down follows the alpha-sorting order.

pllim:
Thanks, @ejeschke ! I'll check it out when I get the chance.

I need some clarification about the `sort_order` preference. So now, I can set it in at least *four* different places, i.e., `general.cfg`, `plugin_Thumbs.cfg`, `plugin_Contents.cfg`, and `channel_Image.cfg`? Which overwrites which? Can I set it to `'alpha'` in one but time in another?

ejeschke:
By sharing the index, changes made in the `Contents` plugin could propagate to the channel sort order.  For example, if you clicked on `Filter` in the `Contents` to sort by filter, it could set that as the sort order that you get when you navigate by arrow key in the image.  The only problem with that scheme is that since clicking a column in `Contents` affects the sort order of all channels it effectively prevents you from having one channel sorted by `OBJECT` say, and one by filter.

pllim:
@ejeschke , do you mind resolving the conflicts on this PR, so I can test it with the latest dev version? Thanks!

ejeschke:
Updated.

pllim:
So, it appears that defining `sort_order` in `~/.ginga/general.cfg` does not do anything, so why is it listed under "General" section in `Preferences`?

pllim:
Setting `sort_order` in `~/.ginga/channel_Image.cfg` does not appear to do anything as well. I tried both "alpha" and "loadtime". The PR as-is always uses "loadtime", and this PR + #235 always uses "alpha".

ejeschke:
Yes, I think the capability is there, but the processing of the sort order via Preference is missing.  There are probably a few tweaks yet for this PR, but the main thing is to find out whether it breaks plugins or other functionality.

pllim:
> but the main thing is to find out whether it breaks plugins or other functionality

I think we got a "chicken and egg" thing going here. I can't confirm whether it breaks anything on my end until I have a functional PR to play with.

ejeschke:
Sorry, I wasn't clear.  I mean whether the new history breaks things like your custom `Contents` etc.  In my testing of the PR with `stginga` plugins, they seem to still work (as far as I am able to test).

I've updated the PR so that if you change the sort order in `Preferences` it will immediately take effect for the channel.  An easy way to see this is if you have a bunch of images that have a logical alphabetical naming order.  Load some of them from the middle of the range, then some later, then some earlier, etc.  Set the preference for `sort_order` to "alpha" then move to `Contents`. Watch the highlight as you use the up/down arrow keys (in the image or from the `Toolbar`) to navigate the images.  They should highlight in "round-robin" fashion alphabetically.  Now change the sort order to "loadtime" and repeat.  You should see them being highlighted in the order they were loaded.

ejeschke:
The `sort_order` in `plugin_Thumbs.cfg` still just controls the sort order overall in `Thumbs`, which doesn't (yet) pay any attention to channel sort order.

I'm open to a different heading for the parameters than "general" in `Preferences`--this has nothing to do with the `general.cfg` config file, which is for overall ginga configuration. It just seemed like a "general" heading for things that didn't fall under more specific headings.

ejeschke:
The bold green highlight *should* change when you step up and down through the images, but this may only work in the stock ginga `Contents`.

pllim:
Not sure what you mean by "stock". I was testing your PR directly, not using my own subclassed version.

ejeschke:
This PR addresses the problem of using the `Datasrc` as the only known history of the channel--it still has the benefits of the `Datasrc` on memory usage, but allows a full history to be maintained and a user-specifiable sort-order on that history.  So navigation between channels and images becomes more intuitive.  But.. we have to decide how `Thumbs` and `Contents` will behave wrt to the channel history and sort order.  It seems like the default should be to simply follow the channel sort order (e.g. alpha) unless the user overrides by sorting on a different column.

ejeschke:
Ok, I think I will merge this, since it doesn't look like there is any major disruption.

pllim:
I found out that the highlighting was not working for me because I had `highlight_tracks_keyboard_focus = True` in my `plugins_Contents.cfg`. I fixed it as part of #190. I also added this new `sort_order` in `channel_Image.cfg` example over there.

ejeschke:
Good to know.  This is one reason I wanted to do this PR: we are getting so many different fixes in each plugin to try to order things, having an ordered history on the channel might make it possible to simplify the plugins and have them all be somewhat consistent.  We'll get it figured out in the other PRs.

Issue #207, opened 2015-10-30
Registering custom plugins


pllim:
"I like the idea of plugins registering themselves somehow on the install so that there is no need for autodiscovery." -- @ejeschke 

https://pythonhosted.org/setuptools/setuptools.html#dynamic-discovery-of-services-and-plugins -- @embray

Instead of shared namespace solution (as discussed in #92).

c/c: @eteq

pllim:
OMG I just c/c-ed three Eric/Erik's

embray:
I've mentioned before that this can be done with entry_points.  Here's a decent primer: http://pylons-webframework.readthedocs.org/en/latest/advanced_pylons/entry_points_and_plugins.html  (not specific to Pylons in any way).

pllim:
> *"Now that the plugin is written we need to write the code in the CMS package to load it."*

Does this mean Ginga's setup needs to acknowledge the external plugins for this to work?

What if the external plugin comes with its own data files or utility functions?

eteq:
@embray - I've always been afraid of these entry points on the theory that they are a setuptools thing that doesn't clearly have a future.  Do you know if the `distutils2` et al. discussions have said this will continue to work in the future?

@pllim - yes, ginga would have to support this in its plugin loading machinery, but I think it would be pretty straightforward to just iterate over the entry points and have them return plugin classes.

My main concern is that I'm not sure this is *desired*: that is, should ginga really automatically show *all* plugins that are installed without the user having to do anything?  It might be that, at least for the reference viewer, the answer is yes.  But it might be confusing for a lot of users when they start seeing a bunch of plugins and have no idea where they came from.

pllim:
> ginga would have to support this in its plugin loading machinery

Not sure if I like this part. That means @ejeschke would have to review and sanction every single external plugin that people write. Maintainability can be a problem here.

eteq:
@pllim - not quite manual - he'd just have to set up the *machinery* - that is, he just writes some code that says "if some package on the pythonpath registers into the ``ginga::local_plugin``, call the `load_plugin` method on it".  Then the plugin packages just say "I register with ``ginga::local_plugin``", and then every time ginga starts up it should load that plugin.  But there's no need for `ginga` to *explicitly* keep track of specific plugins.

ejeschke:
@pllim, I certainly don't want to be in the business of doing that!

[...]

PR #230, opened 2015-11-18
Allow plugins and the GUI layout to be configured via a directory

ejeschke:
This modification allows ginga plugins and the ginga layout to be discovered via a directory.  In this first pass, the directory is configured to be in the user's $HOME/.ginga, but this could be easily changed in the future.

To write the set of standard plugins, invoke ginga with the command:

```
$ ginga --write-plugin-conf
```

This will start ginga, but will also write the configuration for the plugins to $HOME/.ginga/plugin_conf and a layout configuration to $HOME/.ginga/layout.json

Upon subsequent starts, ginga will see that these files exist and will use them to configure itself.

This is a possible solution to Issue #207


eteq:
Just to make sure I understand, @ejeschke, this would be a directory that *only* has plugin information and GUI config, or you're thinking it would also contain  ``ginga_config.py`` info?

And your thinking is that plugins would then *not* use the entry_points in ``setup.py``, is that right?  Just leave it to the user to opt-in?

pllim:
What if the plugin has its own data file?

eteq:
@pllim - I *think* that doesn't matter here: the plugins are installed independently in their own namespaces, and this PR just adds the machinery that lets a user say "the python package called pllims.amazing.new.plugin is a ginga plugin and should be loaded as such".  So then the data file would happily live in `pllims.amazing.new.plugin` however you want it to.

ejeschke:
@eteq: correct.  This PR doesn't change anything regarding `ginga_config.py`: that is still available to do custom configuration.

To make an easy opt-in system we can add a preference (possibly backed by a GUI) that would basically control which directories to look at (or skip) to discover the plugins and layout.

pllim:
I gave it a try, as follows. After I ran `ginga --write-plugin-conf`, I see the following file listings:
```
$ ls ~/.ginga
...   layout.json   ...   plugin_conf/  ...
$ ls ~/.ginga/plugin_conf/
Blink.json      Errors.json       Mosaic.json       Ruler.json
Catalogs.json   FBrowser.json     MultiDim.json     SAMP.json
Compose.json    Header.json       Overlays.json     Thumbs.json
Contents.json   Histogram.json    Pan.json          Toolbar.json
Crosshair.json  Info.json         Pick.json         WBrowser.json
Cuts.json       IRAF.json         PixTable.json     WCSMatch.json
Debug.json      LineProfile.json  Preferences.json  Zoom.json
Drawing.json    Log.json          RC.json
```

To have it find, say, `DQInspect` local plugin from `stginga` (already installed), I added a new file named `~/.ginga/plugin_conf/DQInspect.json` with the following contents:
```python
{
    "module": "DQInspect",
    "pfx": "stginga.plugins",
    "type": "local",
    "workspace": "dialogs"
}
```

Now, starting Ginga normally (i.e., `ginga --loglevel=20 --log=ginga.log`) shows the custom `DQInspect` plugin, as expected. It has no problem reading the data file from `stginga`.

#### Notes

* Removing the `~/.ginga/layout.json` file has no effect on plugins. I guess it is handy to modify Ginga viewer layout for expert users, but I doubt a regular user would bother.
* Removing a JSON from `plugin_conf/*.json` removes the associated plugin. Not sure how I feel about this feature; Could be too prone to user error.
* After I deleted `plugin_conf/Contents.json` (while keeping everything else), I can still add `Contents` back in via `~/.ginga/ginga_config.py`. I think now we are faced with *too many* ways to configure Ginga. I cannot decide which one is better.
* Running `ginga --write-plugin-conf` again restores deleted built-in JSON files (e.g., `Contents.json`) while retaining custom JSON (e.g., `DQInspect.json`). This, I like.

#### Questions

* This breaks `stginga` in the sense that it does not load its own custom plugins anymore. Maybe @eteq can have a look at this?
* Perhaps `ginga --write-plugin-conf` should not start Ginga at all (only write the JSON files)? Since it is run at a setup stage, I find myself just closing Ginga immediately afterwards anyway, and then mess around in `~/.ginga`, before manually running Ginga viewer again.

ejeschke:
>Perhaps ginga --write-plugin-conf should not start Ginga at all (only write the JSON files)? Since it is run at a setup stage, I find myself just closing Ginga immediately afterwards anyway, and then mess around in ~/.ginga, before manually running Ginga viewer again.

We could easily choose this behavior. What I am thinking is that a directory can be written with the plugin configurations when the user installs either ginga or a separately distributed plugin.  It should certainly be easy to reinstall a plugin just by rerunning an installer. 

> Removing the ~/.ginga/layout.json file has no effect on plugins. I guess it is handy to modify Ginga viewer layout for expert users, but I doubt a regular user would bother.

Only necessary if a plugin wants to use a new workspace that is not in the "standard" layout--much like `MultiImage` is doing.

> After I deleted plugin_conf/Contents.json (while keeping everything else), I can still add Contents back in via ~/.ginga/ginga_config.py. I think now we are faced with too many ways to configure Ginga. I cannot decide which one is better.

I think `ginga_config.py` is around for "experts" to configure their own ginga environment. It's like a `.emacs`.

eteq:
@pllim - can you clarify *which* mode of `stginga` is no longer working? That is, of the methods listed on http://stginga.readthedocs.org/en/latest/, which stopped working?

pllim:
@eteq , it is `stginga [args]`

pllim:
@ejeschke , regarding `layout.json` being used by `MultiImage` (in `stginga`), it is unlikely to happen that way. As a distributed package, `stginga` should not actively modify `~/.ginga` contents, so the layout setup will probably happen within `stginga` instead. We can provide an alternative to user by telling them that they can also copy our version of `layout.json` to their `~/.ginga` directory, but I suspect most users will just run `stginga` directly anyway.

pllim:
With this PR, if a user wants to use third-party plugins from various developers, they will have to copy and paste all the JSON files of interest into their `~/.ginga/plugin_conf` directory. But what if the different developers are distributing different versions of `layout.json`? Is this the workflow that we desire?

ejeschke:
@pllim, I wasn't thinking that shuffling files around would be the user's responsibility (although that certainly is a fairly easy to understand way to manage things).  This PR is more proof of concept of using a directory and group of files to auto-discover the available plugins.  I am thinking that a plugin's install script would write its json files to a known area.  We could expand the idea to include a way for the user to enable or disable the plugins on an individual basis, possibly by a configuration file in their own $HOME/.ginga. We could even do something like: if a new plugin is discovered in the plugins directory, then ginga could pop up a one-time box asking whether the user wants to enable or disable the plugin (could have a description, etc.).  Then it would write this decision to the user's own configuration file and only load the plugins enabled there.  There could even be GUI for reviewing which plugins are available and which are enabled or disabled, and allow you to toggle them.

pllim:
I just want to beat that entry-point horse one more time...

As @embray mentioned, "In Trac, plugins are discovered automatically via `entry_points`, but still have to be activated manually by updating the config file. The `entry_points` can still be used to show the user what plugins are available, but they can enable/disable them on an as-desired basis."

It sounds very similar to what you suggested above. Is there a reason *not* to just use `entry_points` then? One of the concerns voiced was whether this feature will still exist in the future, but @embray pointed out that the answer is probably "yes" given that many others use it too.

But either way, the idea of "finding but not loading until explicitly asked to" sounds good. The question is where is a good place to register or write those JSON files that would work for all users or developers?

embray:
Yeah, I wouldn't reinvent the wheel here.

eteq:
I think I agree that using `entry_points` would be more familiar, as there's already documentation/use cases about it...

That does *not* necessarily address the question of where json configuration files go, though, right?  I guess I see two questions there, and I'm not sure which is the most relevant (and which @pllim or @ejeschke  are worried about):

* Should configuration always go in `.ginga/something` so that *users* know where to find it if they want to modify user settings?
* Should configuration always go in `.ginga/something` so that *developers* know where they're supposed to look for configuration information?

Both are certainly acheivable within *any* of these frameworks, but I think it may be somewhat orgthogonal to the question of how `ginga` *detects* plugins.

ejeschke:
Superseded by #426

PR #232, opened 2015-11-20
Fix QT_API for PyQt5

astrofrog:
This looks like a typo, and it caused all kinds of fun when trying to import the IPython qt console afterwards :)

### DIFF
--- a/ginga/qtw/QtHelp.py
+++ b/ginga/qtw/QtHelp.py
@@ -39,7 +39,7 @@ if toolkit in ('qt5', 'choose') and (not configured):
             pass
 
         # for Matplotlib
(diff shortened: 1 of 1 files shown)


ejeschke:
@astrofrog, thanks for this fix!  In my testing with python3 the previous setting seems to work fine--I see the matplotlib plots.  Wondering how that managed to work...falls back to using Qt4 somehow in the same program using Qt5?

astrofrog:
@ejeschke - I'm not 100% sure, but some packages might treat ``QT_API`` as a preference so it could be that it would fall back to PyQt5 if PyQt4 was not available. IPython however doesn't do that, so this should fix my issue.

PR #235, opened 2015-11-25
Qt TreeView to be sorted by leaf index

pllim:
Qt `TreeView` should be sorted by the given leaf index, not "Column 1".


ejeschke:
Thanks, @pllim!

Issue #228, opened 2015-11-17
Add important modules to API doc


pllim:
As discussed in #224.

@ejeschke --  *I think the most important modules for developers using Ginga (as far as API is concerned), are `ImageView`, `Control`, `AstroImage` and various ones in `ginga.canvas` and `ginga.canvas.types`.*

Also see #182.

Issue #182, opened 2015-10-23
Improve developer doc


pllim:
As part of my work at STScI, I will attempt to improve the documentation aimed at developers interested to write their own Ginga plugins.

ejeschke:
This is excellent!

pllim:
I didn't get a chance to do this myself during the spring, but @hcferguson (and maybe even @gderosa2004) made some edits. @hcferguson plans to open a PR for his edits.

ejeschke:
See also Issue #85.

pllim:
Also #693

Issue #219, opened 2015-11-11
Does Ginga readthedocs reflect latest dev version?


pllim:
I was under the impression that https://ginga.readthedocs.org/en/latest/ reflects the latest dev version. However, I am unable to find any documentation on `LineProfile` plugin. Am I missing something?

ejeschke:
I think the `readthedocs.org` sphinx documentation is not being automatically rebuilt on a github commit. I've manually triggered a rebuild, but this needs to be fixed.  @cdeil or @bsipocz, you both seem to be well versed in the arts of travis ci.  Any chance you could have a quick look at the [`.travis.yml`](https://github.com/ejeschke/ginga/blob/master/.travis.yml) and see if you see anything obvious?

cdeil:
Readthedocs runs the sphinx build and puts up the HTML docs if it was successful.
If there's another sphinx build on travis-ci, that is just for testing, it's not the case that travis-ci publishes the output html somewhere.

It looks like the sphinx builds on readthedocs work fine:
https://readthedocs.org/projects/ginga/builds/
https://readthedocs.org/projects/ginga/builds/3475917/

If I build the docs locally and put `LineProfile` in the search field, also nothing is found.
You have to look why `LineProfile` isn't part of the Sphinx API docs.

If someone has time to fix the Sphinx errors and warnings, that would be great!
```
$ python setup.py build_sphinx
running build_sphinx
creating /Users/deil/code/ginga/doc/_build
creating /Users/deil/code/ginga/doc/_build/doctrees
creating /Users/deil/code/ginga/doc/_build/html
Running Sphinx v1.3.1
loading pickled environment... not yet created
WARNING: 'default' html theme has been renamed to 'classic'. Please change your html_theme setting either to the new 'alabaster' default theme, or to 'classic' to keep using the old default.
building [mo]: all of 0 po files
building [html]: all source files
updating environment: 15 added, 0 changed, 0 removed
reading sources... [100%] quickref                                                                                                                                                     
/Users/deil/code/ginga/doc/WhatsNew.rst:17: ERROR: Unexpected indentation.
/Users/deil/code/ginga/doc/WhatsNew.rst:19: WARNING: Block quote ends without a blank line; unexpected unindent.
/Users/deil/code/ginga/doc/WhatsNew.rst:27: ERROR: Unexpected indentation.
/Users/deil/code/ginga/doc/WhatsNew.rst:28: WARNING: Block quote ends without a blank line; unexpected unindent.
[...]

PR #237, opened 2015-12-01
Added full API doc for ImageView

pllim:
Added full API doc for `ImageView`, as part of #228. Also see #182.

I am not sure when I have time to do the rest, such that the cross-references work. So, we can just merge this first, and I'll open new PRs as able.

### DIFF
--- a/doc/api/imageview.rst
+++ b/doc/api/imageview.rst
@@ -1,0 +1,10 @@
+.. _api_imageview:
+
+############
+Image Viewer
+############
+
+.. currentmodule:: ginga.ImageView
+
+.. automodule:: ginga.ImageView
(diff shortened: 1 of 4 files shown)


pllim:
@ejeschke , if you plan to modify `ginga/ImageView.py`, could you please merge this before you do so? Resolving conflicts with this one can be a pain. Thanks.

ejeschke:
This is *so* awesome.

@pllim, I don't plan to change the API of `ImageView` much.  In fact, it has been fairly stable for a while.

pllim:
You are welcome, @ejeschke . Given that #219 is still outstanding, you might want to do a manual RTD update so the official documentation reflects this addition. FYI.

PR #238, opened 2015-12-02
Fix error when channel tries to reload image that fell out of cache

pllim:
Fix error when channel tries to reload image that fell out of cache. Traceback:
```python
2015-12-02 11:14:40,474 | E | Task.py:387 (execute) | Task '<ginga.main.Ginga object at 0x7f18bb7230d0>.task0' terminated with exception: 'Channel' object has no attribute 'switch'
2015-12-02 11:14:40,476 | E | Task.py:391 (execute) | Traceback:
  File ".../ginga-2.5.20151124191302-py2.7.egg/ginga/misc/Task.py", line 377, in execute
    res = self.func(*self.args, **self.kwdargs)
  File ".../ginga-2.5.20151124191302-py2.7.egg/ginga/Control.py", line 1507, in _load_n_switch
    self.fv.gui_do(self.switch, image)
```

### DIFF
--- a/ginga/Control.py
+++ b/ginga/Control.py
@@ -1503,7 +1503,7 @@ class Channel(Callback.Callbacks):
                 # perpetuate the image_future
                 image.set(image_future=image_future, name=imname,
                           path=path)
-                self.fv.gui_do(self.switch, image)
+                self.fv.gui_do(self.switch_image, image)
 
             self.fv.nongui_do(_load_n_switch, imname, info.path,
                               info.image_future)


ginga/Control.py
```diff
@@ -1503,7 +1503,7 @@ def _load_n_switch(imname, path, image_future):
                 # perpetuate the image_future
                 image.set(image_future=image_future, name=imname,
                           path=path)
-                self.fv.gui_do(self.switch, image)
+                self.fv.gui_do(self.switch_image, image)
```
ejeschke: Good catch!  But I think this is actually supposed to be `_switch` (the function that is defined above).

ejeschke:
I've submitted a quick edit that restores it to `_switch` (commit 6daa3644c10a973426e5d8cf68c359389f38ecc0) .  Working for me.  Can you confirm this fixes it for you?

ginga/Control.py
pllim: Hah, yeah, I debated with myself for a while on whether you really meant `_switch` or `switch_image`. Then I took a bet with 50% chance of winning. Still lost. :ghost:

pllim:
@ejeschke , yes, your fix works. I am closing this then. Thanks!

Issue #227, opened 2015-11-17
Load Image from File menu to load multiple images


pllim:
As discussed in #204.

@ejeschke -- *I noticed that the `"File"->"Load Image"` dialog does not allow multiple selection. I rarely use this dialog since I prefer `FBrowser`, but it seems like it should support multiple selection.*

pllim:
A fix is implemented in #239.

ejeschke:
This was merged.

PR #239, opened 2015-12-02
File menu (Qt) enhancement and FBrowser fix

pllim:
Implement multi-selection and wildcard/extension support for File menu in Qt (see #227). The Qt File menu had limited wildcard/ext support from #204 but it was lost in recent changes; I added that back in this PR. The menu should also be pretty flexible now, which allows user to do `shift+click`, `ctrl+click`, or enter things like `*.fits[2]` in the selection box. I am not aware of anything else that uses `FileSelection` class other than the file menu, but please let me know if this unintentionally breaks something else. As usual, I skipped the GTK portion; someone else can implement it if there is interest.

Also fixed error that occurs in `FBrowser` when directory and files are all selected at the same time (during multi-selection).

I toyed with the idea to also support comma-separated text entries (e.g., `image1.fits,image2.fits`) but I abandoned it because it gets confusing really fast unless you explicitly enter the full path every time (otherwise, Ginga automatically assumes some internal base directory that might not be the currently displayed directory). At this point, it is easier to just do `ctrl+click` or open them separately.

### DIFF
--- a/ginga/qtw/QtHelp.py
+++ b/ginga/qtw/QtHelp.py
@@ -8,10 +8,12 @@
 # Please see the file LICENSE.txt for details.
 #
 from __future__ import print_function
+import glob
 import os
 import math
 
 import ginga.toolkit
+from ginga.util import iohelper
 
 have_pyqt4 = False
 have_pyside = False
@@ -91,7 +93,8 @@ elif have_pyqt4:
 elif have_pyside:
     ginga.toolkit.use('pyside')
 else:
-    raise ImportError("Failed to import qt4, qt5 or pyside. There may be an issue with the toolkit module or it is not installed")
+    raise ImportError("Failed to import qt4, qt5 or pyside. There may be an "
+                      "issue with the toolkit module or it is not installed")
 
 
 tabwidget_style = """
@@ -112,6 +115,7 @@ class TopLevel(QtGui.QWidget):
 
     def setApp(self, app):
         self.app = app
+
 
 class ComboBox(QtGui.QComboBox):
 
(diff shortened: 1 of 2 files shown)


ejeschke:
@pllim, this doesn't appear to work for me under python 3/qt5.

ginga/qtw/QtHelp.py
```diff
@@ -170,19 +176,52 @@ def setSpacing(self, val):
...
 
-    def popup(self, title, callfn, initialdir=None,
-              filename=None):
+        """
         self.cb = callfn
-        if self.filew.exec_():
-            filename = list(map(str, list(self.filew.selectedFiles())))[0]
-            self.cb(filename)
+        filenames = QtGui.QFileDialog.getOpenFileNames(
+            self.parent, title, initialdir, filename)
```
ejeschke: In python3 what is returned here is a tuple containing a list of paths and an empty string, whereas under python2 it's just a list of strings.  This causes the loop to fail where the first item gets whole list and the second item is an empty string, neither of which can be opened.

ejeschke:
Otherwise I'm ok with this.  Not so concerned about gtk FileSelection having this, I think `FBrowser` or drag/drop from system finder is much more convenient.

ginga/qtw/QtHelp.py
pllim: Ops, my bad! From what I see at https://www.reddit.com/r/learnpython/comments/2xhagb/pyqt5_trouble_with_openinggetting_the_name_of_the/ , I added a commit that theoretically fixed this. I cannot test this myself because Anaconda does not appear to support Qt5 yet (I cannot find it at http://docs.continuum.io/anaconda/pkg-docs).

pllim:
@ejeschke , could you please grab the latest commit and see if this PR works in Python3/Qt5 now? Feel free to merge if you are okay with the PR. Thank you!

pllim:
@ejeschke ping

ejeschke:
pong!

pllim:
:smile:

Issue #91, opened 2015-03-17
How to close an image?


pllim:
How to close an open image without deleting the whole channel? This might sound silly but I cannot find the close button!

ejeschke:
@pllim, there is no way currently to delete only a single image from a channel.  In the original design, the images would simply get bumped of the `Datasrc` for a channel as more images get examined.  Then they get reloaded from disk if they got bumped out (e.g. clicking on an "old" thumbnail).

It would be pretty straightforward to add this feature, however.  A `Datasrc` has a mapping interface, so something a `del` on the Datasrc for the channel with the image name as key will do it.  If you want to remove the thumbnail and contents entries, etc. that might entail a little more work, like adding an `image-removed` event that can be subscribed to.

BTW, there is not necessarily anything wrong with treating channels as single image "frames" either.  We could also make it easier to create and delete channels with a button click or keypress.  I've thought about doing that for the grid type workspaces, which would then act very similar to ds9 frames.

pllim:
In situations where user accidentally opens an unwanted image and wishes to de-clutter thumbnails by getting rid of it, a close image button would be nice. Intuitively, I was trying to right click on its thumbnail or `Contents` listing for a "Close" menu. I also looked under `File` menu on top left of Ginga main window. I would expect `Thumbs` and `Contents` to update themselves when an image is closed, so `image-removed` event will be very useful as well.

ejeschke:
Hi @pllim, could you try commit 8c8497adb2c2cc06c487b0d0e7ea7d03681c7d63?  You can choose File->Remove Image to remove the current image, or right click on a thumbnail in the Thumbs tab to get a context menu that allows you to remove the image.

pllim:
@ejeschke , that works. Thank you! However, I don't see the point of `"Load"` option that I see when I right-click on a thumbnail.

ejeschke:
Load is the same as left-clicking the thumb.  Try right-clicking and selecting load for an image that is not currently showing in the channel.  Does it work?

pllim:
Yes, it does. In fact, it preserves the modified buffer, which is nice. I was afraid it was going to reload the data from file. Therefore, `Load` is a little misleading in this context. Maybe another wording like `Focus`, `Display`, `Set Active`, or something that isn't `Load`?

pllim:
Also, somewhat related, when I click `delete channel`, `Contents` is updated but not `Thumbs`.

ejeschke:
[...]

PR #243, opened 2015-12-08
Always require at least one active channel

pllim:
Always require at least one active channel. I think this makes sense, but I am open to alternatives.

When only one channel is present, Ginga will refuse to delete it and emits the following log entry:
```python
<timestamp> | E | GingaGw.py:735 (delete_channel_cb) | Delete channel failed. Need at least one active channel.
```

This was brought up in #91, where I reported that there is no safe-guard on the "Delete Channel" functionality.

### DIFF
--- a/ginga/gw/GingaGw.py
+++ b/ginga/gw/GingaGw.py
@@ -732,6 +732,10 @@ class GingaView(GwMain.GwMain, Widgets.Application):
         w.delete()
         if rsp == 0:
             return
+        if len(self.channelNames) <= 1:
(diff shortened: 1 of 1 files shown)


ejeschke:
What happens if there are two workspaces?  Then do we say you must have one channel in each workspace?  No, I think we should simply handle the case that there can be 0 channels.  Shouldn't be hard to fix any remaining issues with that.

pllim:
Okay, that's fine. Regarding disabling the "delete" menu item, it's just a cosmetic change now that Ginga doesn't emit a traceback anymore, so I don't think it's worth the time.

PR #245, opened 2015-12-09
Fix RTD erroring on numpydoc

pllim:
Replaced `numpydoc` with Sphinx's `napoleon` extension, as suggested by @cdeil . Hopefully this would fix the RTD error reported by @ejeschke in #219.

### DIFF
--- a/doc/conf.py
+++ b/doc/conf.py
@@ -26,11 +26,12 @@ import sys, os
 # Add any Sphinx extension module names here, as strings. They can be extensions
 # coming with Sphinx (named 'sphinx.ext.*') or your custom ones.
 extensions = ['sphinx.ext.autodoc',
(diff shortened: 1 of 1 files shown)


cdeil:
@pllim - Thanks!

Probably it's clear, but just in case: the only way to know if this works on RTD is to merge in master to trigger the build. :four_leaf_clover:

pllim:
Only @ejeschke can merge, I think.

FYI - I did a manual build using this PR on my machine (not RTD, of course) and it was successful.

ejeschke:
Giving it a shot!

pllim:
I already see the changes reflected in https://ginga.readthedocs.org/en/latest/ ! (There is now a new "Reference/API" section on the left menu.) :tada:

Issue #144, opened 2015-07-30
RuntimeWarning in IQCalc (Pick plugin)


pllim:
Sometimes, `Pick` gives the warning below to the screen (not Ginga's log file):
```
.../scipy/optimize/minpack.py:421: RuntimeWarning: Number of calls to function has reached maxfev = 800.
  warnings.warn(errors[info][0], RuntimeWarning)
```
This most likely comes from this line below in `IQCalc.calc_fwhm()` method in `util/iqcalc.py` when the data is not well-behaved:
```python
 p1, success = optimize.leastsq(errfunc, p0[:], args=(X, Y))
```

pllim:
Fixed in #247

PR #247, opened 2015-12-09
Fixes for Histogram

### LINKED ISSUE #134: RuntimeWarning in CanvasObjects.py
This was briefly mentioned in #120 but should be its own issue:
```
.../ginga-2.4.20150626080455-py2.7.egg/ginga/canvas/CanvasObject.py:539: RuntimeWarning: invalid value encountered in double_scalars
  Cx = numpy.sum(cx * (a - b)) / (6. * A)
.../ginga-2.4.20150626080455-py2.7.egg/ginga/canvas/CanvasObject.py:540: RuntimeWarning: invalid value encountered in double_scalars
  Cy = numpy.sum(cy * (a - b)) / (6. * A)
```
This is probably explained in http://stackoverflow.com/questions/27784528/numpy-division-with-runtimewarning-invalid-value-encountered-in-double-scalars . The solution can be:
```python
denom = 6. * A
if denom == 0:
    Cx = x_value_for_this_special_case
    Cy = y_value_for_this_special_case
else:
    Cx = numpy.sum(cx * (a - b)) / denom
    Cy = numpy.sum(cy * (a - b)) / denom
```

### LINKED ISSUE #142: Matplotlib UserWarning in Histogram
When I draw a box for `Histogram` and both "Cut Low" and "Cut High" values are zeroes, I get this warning from `matplotlib` on the screen (not Ginga's log file):
```
.../matplotlib/axes/_base.py:2562: UserWarning: Attempting to set identical left==right results
in singular transformations; automatically expanding.
left=0.0, right=0.0
  'left=%s, right=%s') % (left, right))
```

pllim:
1. Fixed an error message when `Histogram` starts up (due to `self.histtag` set to `None`):

```python
2015-12-09 15:38:54,753 | E | Callback.py:89 (make_callback) | Traceback:
  File ".../ginga-2.5.20151208223618-py2.7.egg/ginga/misc/Callback.py", line 72, in make_callback

  File ".../ginga-2.5.20151208223618-py2.7.egg/ginga/misc/plugins/Histogram.py", line 475, in cutset_ext_cb
    self.redo()

  File ".../ginga-2.5.20151208223618-py2.7.egg/ginga/misc/plugins/Histogram.py", line 246, in redo
    obj = self.canvas.getObjectByTag(self.histtag)

  File ".../ginga-2.5.20151208223618-py2.7.egg/ginga/canvas/CanvasMixin.py", line 92, in get_object_by_tag
    obj = self.tags[tag]
```

2. Minor PEP8 fixes for `Histogram`.

3. Also put its `cut_levels()` error messages in log file.

4. ~~Added a check to prevent `UserWarning` from `matplotlib`~~ Generalize Ginga logger to catch and log all Python warnings (fixes #134, #142, and #144).

### DIFF
--- a/ginga/misc/plugins/Histogram.py
+++ b/ginga/misc/plugins/Histogram.py
@@ -14,6 +14,7 @@ from ginga import GingaPlugin
 from ginga import AutoCuts
 from ginga.util import plots
 
+
 class Histogram(GingaPlugin.LocalPlugin):
 
     def __init__(self, fv, fitsimage):
@@ -55,8 +56,8 @@ class Histogram(GingaPlugin.LocalPlugin):
 
         fitssettings = fitsimage.get_settings()
         for name in ['cuts']:
-            fitssettings.getSetting(name).add_callback('set',
-                               self.cutset_ext_cb, fitsimage)
+            fitssettings.getSetting(name).add_callback(
+                'set', self.cutset_ext_cb, fitsimage)
         self.gui_up = False
 
     def build_gui(self, container):
@@ -86,9 +87,11 @@ class Histogram(GingaPlugin.LocalPlugin):
         vbox.add_widget(w, stretch=1)
 
         captions = (('Cut Low:', 'label', 'Cut Low', 'entry'),
-                    ('Cut High:', 'label', 'Cut High', 'entry', 'Cut Levels', 'button'),
+                    ('Cut High:', 'label', 'Cut High', 'entry',
+                     'Cut Levels', 'button'),
                     ('Auto Levels', 'button'),
-                    ('Log Histogram', 'checkbutton', 'Plot By Cuts', 'checkbutton'),
+                    ('Log Histogram', 'checkbutton',
+                     'Plot By Cuts', 'checkbutton'),
                     ('NumBins:', 'label', 'NumBins', 'entry'),
                     ('Full Image', 'button'),
                     )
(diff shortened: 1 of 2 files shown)


ginga/misc/plugins/Histogram.py
```diff
@@ -441,10 +444,26 @@ def cut_levels(self):
...
             loval = float(self.w.cut_low.get_text())
             hival = float(self.w.cut_high.get_text())
+        except Exception as e:
+            errmsg = 'Error cutting levels: {0}'.format(str(e))
+            self.fv.showStatus(errmsg)
+            self.logger.error(errmsg)
+            return
 
+        # This is needed to avoid UserWarning from matplotlib
+        if loval == hival:
```
ejeschke: I wonder if it would be better to compare these *before* they are converted to `float`s. Because of the complications of comparing two floating point values exactly.  There are a few other places this sort of thing needs to be fixed in ginga, but I'm just thinking this is a chance to avoid adding another such case.

ejeschke:
Otherwise looks ok.

ginga/misc/plugins/Histogram.py
pllim: Comparing just the strings won't work if I enter "0" for low and "0.0" for high.

ejeschke: True, that.  Well, I guess worst case is it doesn't catch the equivalence and we get the same error as before.  It might be worth looking into something like `np.isclose()`.  The only problem is that for cut levels the values *can* be very, very close.

pllim: To truly get rid of the warnings, we can simply just let it happen but catch it and redirect to the logger (instead of letting it spill out to the terminal). I'll give this a try shortly.

pllim:
@ejeschke , I modified this PR to just catch all the warnings from Python and redirect them to Ginga's logger. This conveniently fixes all the issues I opened about warnings on the screen. Is this too drastic? Please review and let me know.

Here's an example log with those warnings captured (from `Histogram` and `Pick`, respectively):

```python
2015-12-10 11:42:02,352 | W | _base.py:2767 (set_xlim) | .../matplotlib/axes/_base.py:2767: UserWarning: Attempting to set identical left==right results
in singular transformations; automatically expanding.
left=0.0, right=0.0
  'left=%s, right=%s') % (left, right))
2015-12-10 11:55:10,762 | W | minpack.py:421 (leastsq) | .../scipy/optimize/minpack.py:421: RuntimeWarning: Number of calls to function has reached maxfev = 800.
  warnings.warn(errors[info][0], RuntimeWarning)
```

ejeschke:
Actually, I think redirecting these to the log is the proper way to handle this. :+1:

PR #248, opened 2015-12-10
Fix Cuts plugin

### LINKED ISSUE #131: Cuts Plugin Improvements
@pllim writes in issue #120:

* When a cut is Drawn, is it possible to set the left drop-down menu back to "None"? Currently, the menu remains on the latest cut, which defaults to the behavior of overwriting the cut when I draw another line. If I forget to re-select "None" before drawing again, I lose my latest cut. Does that make sense?
* Delete buttons can be disabled when there is nothing to delete. Right now, they just do nothing when you press them anyway, so this is more of a cosmetic improvement.
* Plot should clear completely when the last cut is deleted, or when "Delete All" is pressed. Right now, the last cut is displayed if they are deleted one by one, or the whole plot remains if they are all deleted at once. The remaining plot is not meaningful because it is no longer associated with any cut lines. Again, a cosmetic improvement.
* I notice that the cut color changes when I delete another cut. Personally, I am fine with it, but I don't know if other users will find it confusing. For example, Cut1 is green, Cut2 is red; Then, I delete Cut1, and now Cut2 is green.

pllim:
Fixed `Cuts` plot keep shrinking with legend enabled. Fixed "Save" and "Delete All" buttons not toggling on/off properly. Fixed "Save" logic. PEP8 stuff.

This fixes #131.

### DIFF
--- a/ginga/misc/plugins/Cuts.py
+++ b/ginga/misc/plugins/Cuts.py
@@ -25,7 +25,8 @@ class Cuts(GingaPlugin.LocalPlugin):
     A plugin for generating a plot of the values along a line or path.
 
     There are four kinds of cuts available: line, path, freepath and
-    beziercurve.
+    beziercurve:
+
     - The 'line' cut is a straight line between two points.
     - The 'path' cut is drawn like an open polygon, with straight segments
       in-between.
@@ -75,6 +76,7 @@ class Cuts(GingaPlugin.LocalPlugin):
     Changing Width of Cuts
     ----------------------
     The width of 'line' cuts can be changed using the "Width Type" menu:
+
     - "none" indicates a cut of zero radius; i.e. only showing the pixel
       values along the line
     - "x" will plot the sum of values along the X axis orthoginal to the
@@ -87,8 +89,8 @@ class Cuts(GingaPlugin.LocalPlugin):
     The "Width radius" controls the width of the orthoginal summation by
     an amount on either side of the cut--1 would be 3 pixels, 2 would be 5
     pixels, etc.
-    """
 
+    """
     def __init__(self, fv, fitsimage):
         # superclass defines some variables for us, like logger
         super(Cuts, self).__init__(fv, fitsimage)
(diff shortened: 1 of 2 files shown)


pllim:
Note: I am not sure what the "Slit" tab is supposed to do. Please test that I didn't break it first.

ginga/misc/plugins/Cuts.py
```diff
@@ -120,6 +122,7 @@ def __init__(self, fv, fitsimage):
                                   show_cuts_legend=False, enable_slit=False)
         self.settings.load(onError='silent')
         self.colors = self.settings.get('colors', cut_colors)
+        self.show_legend = self.settings.get('show_cuts_legend', False)
```
ejeschke: Just for future reference @pllim, if the preference is handled this way (by assigning it to an instance variable in the plugin object and thereafter only referring to the instance variable) then there is no way short of reloading the plugin to turn off the preference dynamically.  I think that this is fine for many preferences (where the user is likely to edit a file and restart the reference viewer), but it complicates logic for toggling the preference via a GUI.  In such cases it may be better to simply do:
```
if self.settings.get('show_cuts_legend', False):
```
to check the value.  This is also reinforces the idea that it is a preference that you are checking.

pllim: Yes, I realize that. However, I think it is necessary here because the plot area is shrunken 20% in `build_gui()` but the legend is not added/updated until `add_legend()` later. So, if the state of `'show_cuts_legend'` changed in between those two calls, things might get messy. Does this make sense?

pllim: Or perhaps I can move `self.show_legend = self.settings.get('show_cuts_legend', False)` from `__init__` to `build_gui()`?

ejeschke: In this case it's fine.  Just wanted you to realize the tradeoff for future updates to plugins.  If there is any idea that the plugin should respond without restarting/reloading to changes in preferences it is preferable to keep the state confined to the settings object.

ejeschke:
Slit seems to be working, to my understanding of it.  Perhaps @rupak0577 can double check.

pllim:
Okay, thanks!

Issue #246, opened 2015-12-09
Histogram axis tick labels cut off


pllim:
In `Histogram` plugin, when the X or Y axis tick labels display large values (e.g., 1000 or larger), I cannot see the values on the plot, much less the axis labels. This is worse for the X axis because height is less than width. I use Qt toolkit.

Example:
![screenshot](https://cloud.githubusercontent.com/assets/2090236/11698966/ad936076-9e8f-11e5-93bd-c0938078a3d8.png)

pllim:
Actually, I think this problem is global in all the plugins that use `matplotlib`. Perhaps could be addressed by #33?

ejeschke:
I think using `tight_layout()` fixes the problem, but may introduce a different problem--the one you are seeing reported in #131.  At any rate, `tight_layout()` seems to put *too much* padding around the plot so that it ends up being really small.

pllim:
I actually cannot confirm/deny the `tight_layout()` claim because it is only applied to `slit_plot`, not `cuts_plot`, and I never understood how to get Slit to work. I fixed the Cuts plot in #248 though.

ejeschke:
Even with #248 merged I still see the numbers and labels being cut off (for `Cuts`).  We need some matplotlib expertise here, methinks.  If it is possible, we might be able to centralize it in `ginga.util.plots` so that all the plugin plots look nice.

pllim:
@ejeschke , #248 fixed the shrinking, not the labels being cut off. I apologize for the confusion.

@mdboom ?

pllim:
@ejeschke , in the meantime, is it possible to embed the plot within a scroll-able area? I tried the following in the plugin but it does not work (i.e., no scroll bars showed up):

```python
w = Plot.PlotWidget(self.plot)
w.resize(400, 300)
sw2 = Widgets.ScrollArea()
sw2.set_widget(w)
vbox.add_widget(sw2, stretch=1)
```

With the scroll bars, at least user has access to the whole plot until we figure out a better way to visualize the entire plot.

ejeschke:
I think the problem is the way in which the plot is rendered within the space. It's not that the actual canvas (widget) is too large.  That would be why using the scroll area does not work.

I think the solution is going to involve using some matplotlib padding options.  Maybe setting some default left and bottom padding?  If we set it in `ginga.util.plots.Plot` it should fix most of the plugin plots.

ejeschke:
[...]

PR #253, opened 2015-12-11
Tight layout for plots

pllim:
I have identified 4 plugins below with labels cut off, as mentioned in #246. 

- [x] `Cuts` - Adding tight layout requires the legend to be within the plot itself, not outside. Still have to scroll horizontally sometimes to see the rest of the plot, but at least everything is showing now.
- [x] `Histogram` - Adding tight layout works nicely.
- [x] `LineProfile` - Seems to work but I only have 1 data cube file to test with.
- [ ] `Pick` - Has no effect, not sure why.

### DIFF
--- a/ginga/misc/plugins/Cuts.py
+++ b/ginga/misc/plugins/Cuts.py
@@ -122,7 +122,6 @@ class Cuts(GingaPlugin.LocalPlugin):
                                   show_cuts_legend=False, enable_slit=False)
         self.settings.load(onError='silent')
         self.colors = self.settings.get('colors', cut_colors)
-        self.show_legend = self.settings.get('show_cuts_legend', False)
 
         self.dc = fv.getDrawClasses()
         canvas = self.dc.DrawingCanvas()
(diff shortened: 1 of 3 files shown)


pllim:
@ejeschke , what do you think?

ejeschke:
@pllim, I like it!  I tried using `tight_layout()` before and the results were not so good.  But I think your use of the `pad` kwarg seems to set the margins appropriately.  I think we could merge no problem.

pllim:
@ejeschke , I couldn't get the same incantation to work for `Pick`. Do you know why, or care? Should we save that for a future PR?

ejeschke:
I wouldn't worry about `Pick` for now, maybe save it for a separate fix.  I think only the axis titles are sometimes cut off a bit on that anyway.  I can usually see the axis values, which is/was the main annoyance with the others.

pllim:
Okay, @ejeschke . Feel free to merge. I don't have plans to add anymore stuff to this PR.

PR #268, opened 2016-01-19
Auto resize columns in Contents

### LINKED ISSUE #263: Treeview column widths
Is there a way to tell `Treeview` the optimal column widths for some of the columns? Or perhaps tell it that the given column is important and should always be expanded to at least X  number of characters? I find that I always have to manually expand the column widths (by dragging its border), as shown below.

First and last columns not showing meaningful info unless fully expanded:
untitled

First column not showing meaningful info unless fully expanded:
untitled2

pllim:
Allow `Contents` global plugin to automatically resize columns to fit to its contents. Also fixed a bug in Qt version of `TreeView`.

This fixes #263.

### DIFF
--- a/ginga/misc/plugins/Contents.py
+++ b/ginga/misc/plugins/Contents.py
@@ -8,6 +8,7 @@
 # Please see the file LICENSE.txt for details.
 #
 from ginga.util.six import itervalues
+from ginga.util.six.moves import map
 
 from ginga import GingaPlugin
 from ginga.misc import Bunch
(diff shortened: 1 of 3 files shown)


ginga/misc/plugins/Contents.py
```diff
@@ -124,6 +128,20 @@ def recreate_toc(self):
                 new_highlight |= channel.extdata.contents_old_highlight
         self.update_highlights(set([]), new_highlight)
 
+        # Resize column widths
+        n_rows = self.nrows()
```
ejeschke: How about:
```python
n_rows = sum(map(len, self.name_dict.values()))
```?

This avoids a function call and a `for` iteration--the slowest parts of python

ejeschke:
@pllim, could you also add a stub for the `set_optimal_column_widths()` method to both gtk and web/pg `Widgets.py`?  It can be a nop for now.  I'll fix the gtk version before merging.  In general this is the way to proceed with these sorts of widget-specific updates, because the alternative is a broken plugin under the other widget sets.

pllim:
> could you also add a stub...

@ejeschke , I don't understand. I am using a method that *you* added in 759c299c0c3 . GTK, Qt, and PG all have the same methods.

pllim:
@ejeschke , I fixed the calculation algorithm for `n_rows`, as you suggested, which is a very good idea. Thanks!

ejeschke:
@pllim, maybe we should do something similar for `FBrowser`? What do you think?

PR #269, opened 2016-01-21
Fix Header not updating on modified callback

pllim:
Fix `Header` global plugin not updating properly on `'modified'` callback. I encountered this when I modify only the header of the image but not the data. I did something like this:

```python
hdr = image.get_header()
hdr[kwd] = val
image.make_callback('modified')
```

I noticed that the listing in `Header` did not update to reflect the new values until I click the sort checkbox.


ejeschke:
Looks good!

PR #270, opened 2016-01-21
New ChangeHistory global plugin

pllim:
Added new `ChangeHistory` global plugin to log buffer change history. This PR makes the plugin available but not loaded by default. I am already using it in `stginga` but I thought it is general enough that Ginga will find it useful.

If you do not want this, just close this PR without merging. But if you want this, after the merge, I'll remove the copy from `stginga` and use this one instead.

**Here is some doc about the plugin (copied from its API doc):**

Keep track of buffer change history.

History should stay no matter what channel or image is active. New history can be added, but old history cannot be deleted, unless the image/channel itself is deleted.

The `redo()` method picks up a `'modified'` event and displays related metadata here. The metadata is obtained as follows:

```python
channel = self.fv.get_channelInfo(chname)
iminfo = channel.get_image_info(imname)
timestamp = iminfo.time_modified
description = iminfo.reason_modified  # Optional
```

While `'time_modified'` is automatically added by Ginga, `'reason_modified'` is optional and has be to explicitly set by the calling plugin in the same method that issues the `'modified'` callback, like this:

```python
# This issues the 'modified' callback and sets the timestamp
image.set_data(new_data, ...)

# Manually add the description
channel = self.fv.get_channelInfo(chname)
iminfo = channel.get_image_info(image.get('name'))
iminfo.reason_modified = 'Something was done to this image buffer'
```

**Here is a screenshot of it in action:**

untitled

### DIFF
--- a/ginga/examples/configs/plugin_ChangeHistory.cfg
+++ b/ginga/examples/configs/plugin_ChangeHistory.cfg
@@ -1,0 +1,14 @@
+#
+# ChangeHistory plugin preferences file
+#
+# Place this in file under ~/.ginga with the name "plugin_ChangeHistory.cfg"
+
+# If set to True, will always expand the tree in ChangeHistory when
+# new entries are added
+always_expand = True
+
+# If set to True, rows will have alternate colors
+color_alternate_rows = True
+
+# Timestamp column width
+ts_colwidth = 250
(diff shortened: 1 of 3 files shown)


ejeschke:
This looks like a useful plugin.  Is there a way to save the changes to a new file, with the history attached as metadata (maybe an added FITS table or something)?  Thinking this could maybe fit in with some kind of an extensible pipeline plugin.

pllim:
> Is there a way to save the changes to a new file

Not currently in this plugin. I do not think the "save" feature should be part of this plugin. However, it would not be hard for a plugin developer to extract `TreeView` info from this plugin and save the history however they want.

I do have something that saves the changes to output files, but it is very specific to the particular work that requires those outputs.

ejeschke:
Works with `Mosaic`, by my testing.

Issue #87, opened 2015-03-09
Doc: Ginga color palette


pllim:
Would be nice for developer to have access to a color palette in the documentation as defined by `colors.py`.

ejeschke:
Do you mean to be able to define your own colormaps?

pllim:
No, I just want to see a table of supported color names and what the color looks like. I spent 15 minutes today going through the names in `colors.py` trying to find one that looks nice for my plugin. Something like http://matplotlib.org/examples/color/colormaps_reference.html

ejeschke:
I see.  I actually have some code that shows a small preview of the color map, along with the name.  Was planning to use this, but it seems there is a problem with it and recent versions of Qt.  Need to work on it.

Do you want to use a matplotlib colormap with ginga?  You can.

pllim:
I wasn't really looking for an entire colormap. I was just looking for one specific color that would go nicely with the image I am using (for a shape I am trying to superimpose on the image). The small preview you mentioned sounds useful when you get the chance to fix it.

ejeschke:
Just FYI, if you have matplotlib installed, and want to add their selection of colormaps to ginga, add the following line to `$HOME/.ginga/general.cfg`:
```python
useMatplotlibColormaps=True
```

pllim:
I think this has been solved by `ColorMapPicker` plugin.

PR #271, opened 2016-01-23
Some modifications to the ginga/colors.py file and added tests for the same

rajul:
Made a few changes to the functions of the *ginga/colors.py* file
* lookup_color(): Added a meaningful error message if queried with a non-existent color name
* add_color(): Added a _validate_color_tuple() to validate the passed the color before adding it to color_dict
* remove_color():  Added a meaningful error message if a non-existent color name is tried to remove
* scan_rgbtxt(): Refactored this function into two; added a scan_rgbtxt_buf() which takes a string/buffer argument and parses it to read RGB values

**Tests**: Added tests for the old and the new functionalities of the file

### DIFF
--- a/ginga/colors.py
+++ b/ginga/colors.py
@@ -764,9 +764,22 @@ def lookup_color(name, format='tuple'):
     else:
         raise ValueError("format needs to be 'tuple' or 'hash'")
 
+def _validate_color_tuple(tup):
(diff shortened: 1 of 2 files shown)


ejeschke:
This is really excellent.  Thanks!

pllim:
Is this related to #87?

ejeschke:
Don't think so.

PR #272, opened 2016-01-26
Modifications and tests for the Callbacks class

rajul:
I have studied the code and behavior of the functions here. I have made only minor modifications and tried not to make any modifications to the core behavior of any of the functions as this might have impact in other places in the system as *Callbacks* is the parent class of the core components/class of the code, including *BaseImage*. I have instead added a few TODO comments. However, that changes I have made can be summarised as:

* **clear_callback()**: Changed the reassignment of the callback list to empty list to slice assignment of the entire list, to clear the list instead of simply pointing it to a new object (Ref.. http://stackoverflow.com/questions/1400608/how-to-empty-a-list-in-python) 
* **delete_callback()**: Added KeyError handling
* **add_callback()**: Added the key-not-present logic in the function and removed the KeyError exception handling as this would now be pointless
* **set_callback()**: made this to just fallback on the add_callback()

Added **comprehensive tests** for all the functions of the class

Added some **TODO** comments, will be glad on some feedback about those!


ginga/misc/Callback.py
```diff
@@ -28,24 +34,36 @@ def has_callback(self, name):
...
-            if not tup in self.cb[name]:
-                self.cb[name].append(tup)
+            del self.cb[name]
         except KeyError:
             raise CallbackError("No callback category of '%s'" % (
                 name))
 
-    # TODO: to be deprecated ?
+
+    # TODO: Add a argumet validation function for a callback
```
ejeschke: "argument"

ginga/misc/Callback.py
```diff
@@ -28,24 +34,36 @@ def has_callback(self, name):
...
+    # Pointers:
+    #      * Check the name of the event for which callback added
+    #      * Check that fn is a function
+    #      * Check fn accepts at least one argument, the calling object
+    #      * Check fn takes:
+    #           - at least len(args) + len(kwdargs) number of args
+    #           - variable number of args or keyword args
+    #      * Does and what value the callback funtion return
```
ejeschke: "function"

ginga/misc/tests/test_Callback.py
```diff
@@ -0,0 +1,424 @@
...
+		assert expected == actual
+
+	def test_add_callback_arguments(self):
+		test_callbacks = Callback.Callbacks()
+		
+		def test_callback_function(obj, *args, **kwargs):
+			pass
+
+		test_callbacks.add_callback(
+							"test_name",
```
ejeschke: Please use PEP8 python formatting, when possible.

ginga/misc/Callback.py
```diff
@@ -28,24 +34,36 @@ def has_callback(self, name):
         #return self.cb.has_key(name) and (len(self.cb[name]) > 0)
         return name in self.cb
 
+    # TODO: Should this raise CallbackError, or silently pass
```
ejeschke: Should raise a KeyError if the callback category doesn't exist, as now.

ginga/misc/Callback.py
```diff
@@ -28,24 +34,36 @@ def has_callback(self, name):
...
+    #      * Check the name of the event for which callback added
+    #      * Check that fn is a function
+    #      * Check fn accepts at least one argument, the calling object
+    #      * Check fn takes:
+    #           - at least len(args) + len(kwdargs) number of args
+    #           - variable number of args or keyword args
+    #      * Does and what value the callback funtion return
+    #          
+    def add_callback(self, name, fn, *args, **kwdargs):
+        self.enable_callback(name)
```
ejeschke: `add_callback` should raise a `KeyError` if the callback category doesn't exist.

ginga/misc/Callback.py
```diff
@@ -28,24 +34,36 @@ def has_callback(self, name):
...
+        
+        tup = (fn, args, kwdargs)
+        if not tup in self.cb[name]:
+            self.cb[name].append(tup)
+        
     def set_callback(self, name, fn, *args, **kwdargs):
-        if not self.has_callback(name):
-            self.enable_callback(name)
-        return self.add_callback(name, fn, *args, **kwdargs)
+        self.add_callback(name, fn, *args, **kwdargs)
```
ejeschke: `set_callback` should create the category if it doesn't exist and then add the callback.

ginga/misc/Callback.py
```diff
@@ -28,24 +34,36 @@ def has_callback(self, name):
...
 
+    # TODO: Returns True even if any one of the callback succeeds...Is that desired?
```
ejeschke: The return value of callbacks is not well defined (except for one case, I think--I'll have to look that one up and update this comment).  Most callbacks are used as an event notification-type mechanism.  The `Future` class (another good candidate for unit tests!) is more preferentially used to convey results/values between asynchronous tasks in Ginga.

ejeschke:
@rajul, really appreciate you adding these badly needed unit tests!

I think the semantics of the `add_callback`/`set_callback` should be preserved.  That is, `add_callback` should raise a `KeyError` if the callback category does not exist.

rajul:
Hi Eric, thanks for your comments and feedback. 

* Re-formatted using PEP-8. Thanks!!
* Changed add_callback and set_callback to their earlier behavior
* Added/modified tests for add_callback and set_callback covering these
* Corrected the typos you pointed out in TODOs and removed some TODO comments. Thanks again!!

Kindly review this again and tell me if there are any more changes needed.

rajul:
@ejeschke I have added a few tests covering the Future class. Kindly review these. Thanks!!

ginga/misc/tests/test_Future.py
```diff
@@ -0,0 +1,319 @@
...
+# Unit Tests for the Future class
+#
+# Rajul Srivastava  (rajul09@gmail.com)
+#
+import unittest
+import logging
+import threading
+import time
+
+import ginga.misc.Future as gingaMisc
```
pllim: No reason to use camel case here, which is discouraged by PEP8.

pllim: Is there a reason not to simply do `from ginga.misc import Future`?

rajul:
@pllim I have removed the camel case and am importing the class from module now. There was no particular reason for this. Importing module and importing from modules does make a difference when we have a script with a set of class-less functions and variables. However, here importing just the classes I am instantiating does makes more sense. Thanks a lot!!

ginga/misc/Future.py
```diff
@@ -24,7 +27,8 @@ def __init__(self, data=None):
 
     def get_data(self):
         return self.data
-    
+
+    # TODO: Could add some args type/value, return value validation here
```
ejeschke: Except that a future doesn't have a set argument structure or return type.

ejeschke:
@rajul, I'm not sure of the "best practice" for this, but I'm inclined to keep all the tests under `ginga.tests` (with possibly nested structure below that, as needed).  This makes it fairly simple to find/run/maintain the tests.

Can you restructure this PR to move the `test_Callback.py` and `test_Future.py` into `ginga.tests`.  There is already a `test_Task.py` there (which is also from `ginga.misc`, so these should be in good company there.

ginga/misc/Future.py
rajul: @ejeschke Yes, I understand. I meant like we can check when adding the method in freeze() that if it is really a `callable`. Also, in other places we are checking on the value returned by the method. So, we may want to see if it returns a non-null value. I may be over-killing this, but in my opinion Python being a dynamically-typed language, we would explicitly need to protect ourselves against things like this, especially in an open-source project. This class was created for a specific purpose and serves the author's purposes well for now. But we never know and can barely imagine all scenarios about how a new contributor would end up using it. So my point with this, we should add protection agains all undesirable scenarios if and when we can.

ejeschke: Some good points. Fair enough.

rajul:
@ejeschke Hmm. Well, there is no established best practice about this AFAIK. However, I have seen projects that have both types, with separate `test` sub-modules inside each individual egg in a project and the other being having a standalone top level tests package(as you suggested). My personal preference is the first one as in my opinion, it has a few advantages over the second approach.

- It will let the new contributor find tests more easily as they will a separate individual module isolated in one place
- Keep the size of my test modules smaller
- Allow for one to add a feature and can run tests just for that module( well this could be possible if we use a sub-modular structure in our standalone `tests` module too)

Also, you made a very good point about we need a simple way to find/run/maintain the tests. Of course, to run the tests we can use the `discover` provided in the `unittest` module. But I was thinking of working on an Automated Testing Framework over the weekend, as a separate `ginga.testing` module which just at the top of my head:

- will discover and run all tests during the build process (of course, we can choose not to run it like in most build systems)
- let us provide support for other common Python testing frameworks, like `py.test`, other than just `unittest` and let us run all different framework's test at build time
- provide some report about the coverage at the end of the test build process
- (we can extend this module in future with a lot more functionalities)

I suppose this kind of a test framework would be useful.

These are my thoughts on this. If you still insist on a top-level tests module, I can move the test classes.

ejeschke:
@rajul, well, I think you convinced me!

The current "build system" is travis-ci, and if you look at the `.travis.yml` file you can see that it is already using pytest (py.test).  For my testing that is also what I use.  I made the first unit tests in the standard unittest framework because that is a kind of "lowest common denominator" that should continue to work if  a new unittest framework becomes popular.

So if we add tests outside the `ginga.tests` directory, I think we need to modify the `test_suite` line to be "ginga" (?), not "ginga.tests" (and being sure to add the appropriate line to `packages` to make sure the tests are installed).  

There is some TODO type stuff in there for coverage. You might have a look.  I think it is supposed to check coverage with "python setup.py test --coverage".

pllim:
Since Ginga is Astropy affil pkg, theoretically it should follow Astropy org style, which is also putting the tests under their respective subpkg.

pllim:
Related link in Astropy: http://docs.astropy.org/en/latest/development/testguide.html

rajul:
@ejeschke Thanks!! I will take a look in these places and make appropriate changes.
@pllim Thanks to you too!!

pllim:
@rajul , no problem :smile:

rajul:
Have added a function in the module `ginga.tests` and made this as the test_suite entry in `setup.py`. This function effectively discovers all the `unittest` tests inside main `ginga` module for now and runs them on `python setup.py test`

rajul:
The `python3` seem to be failing. Looking into it.

rajul:
The Travis build is passing. This seems ready for merge from my end. Please review and provide feedback if any further changes are needed.

ginga/tests/__init__.py
```diff
@@ -0,0 +1,7 @@
+import unittest
+import ginga
+
+def ginga_test_suite():
+	loader = unittest.TestLoader()
+	suite = loader.discover('ginga')
+	return suite
```
ejeschke: Looks like needs a newline here

ejeschke:
Looks good!  I think that `make_callback()` probably should raise an error if the callback category does not exist.  But that is my commenting out the behavior, so I think there must have been an issue with that--probably involving a race condition regarding the callback--one part of the code trying to use it before it has been created by another part.  Anyway, that can be sorted out later.  Merging.

PR #276, opened 2016-01-28
FBrowser auto resizing column widths

### LINKED ISSUE #274: FBrowser plugin to auto resize columns
Quoting @ejeschke from #268, "... maybe we should do something similar for `FBrowser`?"

Auto resizing columns in `FBrowser` would make sense. However, unlike `Contents`, I do not think there is a need to turn the feature off based on a max number of rows. What do you think, @ejeschke ?

Also see #263.

pllim:
`FBrowser` to auto resize column widths (regardless). This fixes #274.

Also see: #268, #263

### DIFF
--- a/ginga/examples/configs/plugin_FBrowser.cfg
+++ b/ginga/examples/configs/plugin_FBrowser.cfg
@@ -10,7 +10,7 @@ home_path = None
 # listing (slow for large numbers of files)
 scan_fits_headers = True
 
-# If the number of files in the listing is greater than this, don't do 
+# If the number of files in the listing is greater than this, don't do
 # a scan on the headers
 scan_limit = 1000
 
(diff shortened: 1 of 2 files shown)


pllim:
@ejeschke , instead of adding another user preference, I just use the existing `scan_limit` value. Does this look good to you?

ejeschke:
@pllim, I think we'd better introduce a separate preference.  `scan_limit` could be quite slow (it opens every FITS file and tries to read the headers) if we set the default in the thousands, whereas I think we can set the limit for this autosizing to around 5000 or so.  Otherwise looks fine,

pllim:
@ejeschke , okay. New preference added. It is the same name as `Contents`.

ejeschke:
@pllim, I suppose we should think about auto-sizing Header columns too.  What do you think?

pllim:
@ejeschke , does not hurt to have. See #281. 

However, I do notice that with auto resize, you will have to scroll to right to see the comments column. I don't have a problem with that though. Do you?

ejeschke:
I think that's fine and there is no easy way to get around that if the plugin occupies a narrower space than the table.

PR #279, opened 2016-02-05
Added directory browser widget

pllim:
Added directory browser widget (GTK and Qt). I need this for spacetelescope/stginga#87 but it is just generally useful for Ginga to have anyway.

Disclaimer: Untested under GTK but theoretically should work if http://stackoverflow.com/questions/10868167/make-filechooserdialog-allow-user-to-select-a-folder-directory answer is correct.

### DIFF
--- a/ginga/gtkw/GtkHelp.py
+++ b/ginga/gtkw/GtkHelp.py
@@ -498,11 +498,11 @@ class MDIWidget(gtk.Layout):
 
 class FileSelection(object):
 
-    def __init__(self, parent_w, action=gtk.FILE_CHOOSER_ACTION_OPEN):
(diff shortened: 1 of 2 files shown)


ejeschke:
@pllim, are these similar enough to the existing `FileSelection` objects that the latter could just take a keyword argument to make them select folders only?

pllim:
@ejeschke , for Qt widget, `FileSelection` is a bit more complicated (wildcard etc.), so subclassing is not feasible.

pllim:
See #239 for related `FileSelection` PR that was merged.

ejeschke:
Ok, cool.

pllim:
Thanks!

PR #280, opened 2016-02-05
Fixed Datasrc sortedkeys not updated properly when image is added

pllim:
Fixed `Datasrc` sorted keys not updated properly when an image is added to Ginga's data cache. This resulting in its `keys()` method returning empty list in some cases.

### DIFF
--- a/ginga/misc/Datasrc.py
+++ b/ginga/misc/Datasrc.py
@@ -7,11 +7,13 @@
 #
 import threading
 
+
 class TimeoutError(Exception):
     pass
 
-class Datasrc(object):
 
+class Datasrc(object):
+    """Class to handle internal data cache."""
     def __init__(self, length=0):
         self.length = length
         self.cursor = -1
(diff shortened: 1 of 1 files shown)


ginga/misc/Datasrc.py
```diff
@@ -53,6 +53,9 @@ def push(self, key, value):
             self.datums[key] = value
             self._eject_old()
```
ejeschke: @pllim, I believe _eject_old() resorts the keys.

ginga/misc/Datasrc.py
```diff
@@ -53,6 +53,9 @@ def push(self, key, value):
...
 
+            self.sortedkeys = list(self.datums.keys())
+            self.sortedkeys.sort()
```
ejeschke: So this would be a redundant sort.

pllim:
Ah, okay, I missed the `_eject_old()` call. However, the code as-is (without this PR) does not updated `sortedkeys` properly. After I opened an image in Ginga, in my plugin, I tried to iterate over the sorted keys, but it gives me an empty list.

pllim:
Which means in `_eject_old()`, it returned without doing anything:

```python
        if (self.length is None) or (self.length <= 0):
            # no limit
            return
```

pllim:
I updated my PR to be less redundant. Why do you think, @ejeschke ?

ejeschke:
Yeah, this looks better.  Thanks!

PR #283, opened 2016-02-10
Added meaningful change history for Mosaic

pllim:
`Mosaic` now has useful description to go with `ChangeHistory` global plugin. Example:
```
timestamp1 -- Added image1
timestamp2 -- Added image2
```


pllim:
Thanks, @ejeschke !

PR #286, opened 2016-02-22
Thumbs now removes deleted mosaic

### LINKED ISSUE #278: Deleted mosaic still in Thumbs
When a mosaic (created via `Mosaic` plugin) is removed from a channel, its thumbnail is not removed from `Thumbs` even when it has disappeared from both `Contents` and `Datasrc`.

pllim:
Fixes #278

### DIFF
--- a/ginga/misc/plugins/Thumbs.py
+++ b/ginga/misc/plugins/Thumbs.py
@@ -204,9 +204,6 @@ class Thumbs(GingaPlugin.GlobalPlugin):
         if not self.gui_up:
             return
 
-        if path is None:
-            # Currently we need a path to make a thumb key
-            return
         self.logger.info("removing thumb for %s" % (name))
 
         try:


ejeschke:
@pllim, I just wonder about the effect of having a None path.  In the past, it was considered necessary to have a path, because that could help uniquely identify a thumb (because all the images are mixed together in thumbs).  For example, you could have the same named image loaded twice in the same channel, as long as they had different paths.  I suppose the requirement now is to have a unique name, or only have one image with the same name in any channel.  We have some applications of having the same image loaded in different channels.

I need to consider the ramifications of this just a little more.  It might be time to revisit the idea of how the image is managed when it leaves the `Datasrc`.  Instead of dropping the image and trying to reconstruct if the thumb is clicked, it might be better to just drop the data (or close the file, if memmapped, etc). This would allow us to better track an image through different plugins because it would always have the same image object--unless you manually removed it.

pllim:
Regarding dropping data and `astropy.io.fits`, please keep this in mind:
http://docs.astropy.org/en/stable/io/fits/#working-with-large-files

*When opening a file with `memmap=True`, because of how mmap works this means that when the HDU data is accessed (i.e. `hdul[0].data`) another handle to the FITS file is opened by mmap. This means that even after calling `hdul.close()` the mmap still holds an open handle to the data so that it can still be accessed by unwary programs that were built with the assumption that the `.data` attribute has all the data in-memory.*

*In order to force the mmap to close either wait for the containing HDUList object to go out of scope, or manually call `del hdul[0].data` (this works so long as there are no other references held to the data array).*

ejeschke:
After testing, I don't see any major issues with this.  Merging.

Issue #291, opened 2016-03-14
Drawing plugin: Annulus cannot draw when "wcs" is selected


pllim:
In `Drawing` local plugin, select the following:

* Draw type: annulus
* Coord type: wcs
* Mode: draw

Now try to draw it over image display with left mouse button. Nothing happens. However, it appears as usual when "Coord type" is changed to "data".

pllim:
Note: This issue should solve itself if Ginga switches to proposed new Astropy region objects that are WCS-aware, as discussed during Python in Astronomy 2016.

PR #292, opened 2016-03-14
Added mask creation functionality in Drawing

pllim:
In the `Drawing` local plugin, I added a "Create Mask" button to create boolean mask from drawn objects. The resultant mask has 1's where the drawn object contains the pixels, and 0's otherwise. Mask shape is taken from the displayed image when the button was pressed.

Not all drawn objects create mask that makes sense. For example, text or crosshair. But for the shapes that matter (e.g., circle, box, closed polygon, and the like), this should work properly.

The resultant mask will be inserted into Ginga as a new image, not unlike what happens when you use the `Mosaic` plugin.

### DIFF
--- a/ginga/misc/plugins/Drawing.py
+++ b/ginga/misc/plugins/Drawing.py
@@ -4,10 +4,16 @@
 # This is open-source software licensed under a BSD license.
 # Please see the file LICENSE.txt for details.
 #
+
+# THIRD-PARTY
+import numpy as np
+
+# LOCSL
 from ginga import GingaPlugin
 from ginga import colors
 from ginga.gw import Widgets
 from ginga.misc import ParamSet, Bunch
+from ginga.util import dp
 
 draw_colors = colors.get_colors()
 
@@ -16,7 +22,9 @@ default_drawcolor = 'lightblue'
 fillkinds = ('circle', 'rectangle', 'polygon', 'triangle', 'righttriangle',
              'square', 'ellipse', 'box')
 
+
 class Drawing(GingaPlugin.LocalPlugin):
+    """Local plugin to draw shapes on canvas."""
 
     def __init__(self, fv, fitsimage):
         # superclass defines some variables for us, like logger
@@ -48,6 +56,10 @@ class Drawing(GingaPlugin.LocalPlugin):
         self.drawparams_cache = {}
         # holds object being edited
         self.edit_obj = None
+
+        # For mask creation from drawn objects
+        self._drawn_tags = []
+        self._mask_prefix = 'drawing'
 
     def build_gui(self, container):
         top = Widgets.VBox()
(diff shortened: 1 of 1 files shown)


ejeschke:
This looks really great.  Will test it today.

ejeschke:
@pllim, this looks really useful.  How do users save the resulting mask?

ejeschke:
If I draw a circle with wcs coordinates it fails to create a mask for it. It works for ellipses though.

pllim:
> How do users save the resulting mask?

@ejeschke , currently I use a `SaveImage` plugin available in `stginga` (http://stginga.readthedocs.org/en/latest/plugins_manual/saveimage.html). Are you interested to have me submit a PR to port that over as well?

pllim:
> If I draw a circle with wcs coordinates it fails to create a mask for it.

@ejeschke , does your data file has a valid WCS? I cannot reproduce the problem with my test file. But I don't see why that is an issue. All I am doing is to use `get_shape_mask()` method in `BaseImage.py`, which does not use WCS information, as far as I can tell.

ejeschke:
@pllim, I tried a few different files, including some HST images (with circles and ellipses) and I find that if data coordinates are used there is no problem, but if wcs coordinates are used the mask looks odd.  I agree that the `get_shape_mask()` method should work as advertised.  I don't have time to look into this now, but maybe something we could look into at PiA.

pllim:
@ejeschke , I see what you are saying now that I use a different image (from HST). However, I don't think that that bug is related to this PR in particular because even if you don't create a mask, I find that you cannot even edit that circle drawn in WCS space. I think this might be related to #291 instead.

ejeschke:
@pllim, can you add that information about not being able to edit a circle to #291?  If that is the problem then it becomes a simpler issue to look into.

pllim:
@ejeschke , okay, I updated that issue.

pllim:
Looking a bit more into this, I also notice that when a Box is drawn in WCS space, you get a distorted box if the FOV is distorted, which makes sense. But "create mask" creates a undistorted box, which does not match the drawn outline.

My hunch is that all these problems with the shapes in WCS mode boils down to their respective `contains_arr()` method not handling WCS to data conversion properly. But I do not have a good idea on how to fix it across the board.

pllim:
@ejeschke , just wondering if it is possible to merge this soon before a potential demo at STScI showing off this feature later this week. So far, the problem discussed above does not have an obvious solution and does not seem to be caused by this PR, IMHO.

ejeschke:
@pllim, I like the functionality and I don't want to hold up this because of the wcs drawing.  I'm just thinking that it is not clear to the user that after pressing "create mask" that they have to start another plugin to save the resulting mask image--*and* that plugin is not available in stock Ginga.

Tell you what--let's add a "Save As FITS" menu item that starts the "Save Image" plugin (which you should submit as a separate PR which we can merge first).  That way after creating the mask they can either start the save image plugin from the `Operations` menu or they will discover it under the `File` menu.

pllim:
@ejeschke , ah yes, that seems reasonable. Let's discuss the save image plugin in #312.

ejeschke:
Merging, now that #325 is merged and we have a way to save the masks.

Issue #202, opened 2015-10-28
4D FITS and MultiDim


nabobalis:
Creating a 4D FITS file, with shape [10,89,1000,1000] which are [Time,Wavelength,Spatial]. NAXIS3 and NAXIS4 are the correct length in MultiDim but it can not index NAXIS3 beyond the range of NAXIS4.

This seems to me that NAXIS3 and NAXIS4 are switched when they should not be. So somewhere the ordering of the axes is getting incorrectly set but I can not seem to track it down. 

@Cadair @rupak0577

Cadair:
A simple example of this would be:

```python

import numpy as np
from astropy.io import fits

arr = np.random.random([10, 89, 1000, 1000])

hdu = fits.PrimaryHDU(arr)
hdulist = fits.HDUList([hdu])
hdulist.writeto('new.fits')
```

@nabobalis can you check the above?

nabobalis:
@Cadair I already did with some IBIS data. But I can run that.

Cadair:
@nabobalis just to give a way for other people to check it.

nabobalis:
@Cadair I ran it anyway, the same issue.

Cadair:
yes probably best to drop the image down to 100x100

Cadair:
@nabobalis So it seems that:

The callback gets added with the numpy dimension number here: https://github.com/ejeschke/ginga/blob/master/ginga/misc/plugins/MultiDim.py#L255

which then calls: https://github.com/ejeschke/ginga/blob/master/ginga/misc/plugins/MultiDim.py#L210

which calls: https://github.com/ejeschke/ginga/blob/master/ginga/misc/plugins/MultiDim.py#L412

which inverts the axis here: https://github.com/ejeschke/ginga/blob/master/ginga/misc/plugins/MultiDim.py#L423

so I guess there is a bug in that axis inversion code at the end there.

rupak0577:
Hmm, I should have reported this before - I had this issue with a 4D file but thought it was due to the file. Also, another issue is the slider control buttons which seem to work only for NAXIS3.

ejeschke:
Good find!  I'll take a closer look later today, but if you have a fix, please send a PR!

rupak0577:
[...]

PR #293, opened 2016-03-15
Fix for swapped NAXIS3 and NAXIS4 in MultiDim [#202]

rirze:
This patch is a fix for the swap of NAXIS3 and NAXIS4 in the MultiLib window, as noted in #202. The code was inverting the indices, when no inversion was necessary. 

As noted in #202, the control buttons don't work with NAXIS4. Also the NAXIS4 GUI axis skips by two when dragging the pointer across the scale. These problems still persist; the only issue fixed is the axis swap.

### DIFF
--- a/ginga/misc/plugins/MultiDim.py
+++ b/ginga/misc/plugins/MultiDim.py
@@ -292,6 +292,8 @@ class MultiDim(GingaPlugin.LocalPlugin):
             for n in range(2, len(dims)):
                 key = 'axis%d' % (n + 1)
                 self.w[key].add_callback('activated', play_axis_change_func_creator(n))
+                if n == 2:
+                    self.w[key].set_state(True)
 
             
         self.play_axis = 2
--- a/ginga/misc/plugins/MultiDim.py
+++ b/ginga/misc/plugins/MultiDim.py
@@ -18,6 +18,7 @@ from ginga.util import iohelper
 
 import numpy as np
 import matplotlib.pyplot as plt
+import copy
 
 have_mencoder = False
 if spawn.find_executable("mencoder"):
@@ -143,8 +144,8 @@ class MultiDim(GingaPlugin.LocalPlugin):
         b.interval.set_enabled(False)
         vbox.add_widget(w, stretch=0)
 
-        captions = [("Slice:", 'label', "Slice", 'llabel',
-                     "Value:", 'label', "Value", 'llabel'),
+        captions = [("Slice:", 'label', "Slice", 'llabel',),
+                     #"Value:", 'label', "Value", 'llabel'),
                     ("Save Slice", 'button'),
                     ]
         w, b = Widgets.build_info(captions, orientation=orientation)
@@ -227,12 +228,13 @@ class MultiDim(GingaPlugin.LocalPlugin):
                                  #"Choose %s" % (title), 'spinbutton'))
                                  "Choose %s" % (title), 'hscale'))
 
-        radiobuttons = []
-        for i in range(2, len(dims)):
-            title = 'AXIS%d' % (i+1)
-            radiobuttons.extend((title,'radiobutton'))
-        captions.append(radiobuttons)
-        
+        if len(dims) > 3: # only add radiobuttons if we have more than 3 dimensions
+            radiobuttons = []
+            for i in range(2, len(dims)):
+                title = 'AXIS%d' % (i+1)
+                radiobuttons.extend((title,'radiobutton'))
+            captions.append(radiobuttons)
+            
         # Remove old naxis widgets
         for key in self.w:
             if key.startswith('choose_'):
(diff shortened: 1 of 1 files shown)


pllim:
c/c @Cadair , @nabobalis , @rupak0577

nabobalis:
Assuming I have pulled this PR correctly, this has not fixed it.
The third slider still changes time and the 4th axis slider does not change anything.

rupak0577:
@rirze Like @nabobalis said, this does not seem to fix the issue and also breaks the plugin for 3D files. Something else seems to be the problem.

rirze:
@nabobalis @rupak0577 Sorry, I might have misunderstood the issue at first glance. I've added another commit, which fixes the axis switch and is more natural, and works for 3D images. 

Also I'm testing with randomly-generated images as done in #202; is there a better way to test this case?

pllim:
Not sure if this is what you are looking for, but here is a test image from #146 -- http://www.lmsal.com/solarsoft/irisa/data/level2_compressed/2015/10/09/20151009_041433_3620263378/iris_l2_20151009_041433_3620263378_SJI_1330_t000.fits.gz 

Also, when the changes are finalized, I would suggest squashing all these changes in this PR into a single commit before merging.

nabobalis:
That fixes it on my end for the 3D and 4D FITS I have.  :+1:

rirze:
@nabobalis What would be a good name for this PR?

Would "Fix for swapped NAXIS3 and NAXIS4 in MultiDim" be fine?

nabobalis:
I am just being pedantic. Either that, or just stay state the issue number. You do not need to say github.com.

ejeschke:
Well done, @rirze!

@nabobalis, @Cadair, @rupak0577 I notice that the play feature of `MultiDim` only plays the 3rd dimension.  Would it be desirable to be able to pick which axis to play?  If so, that would be a simple addition to this PR that would make it a reasonably significant PR for the astropy GSoC PR requirement.

nabobalis:
It would be nice if we could play a different axis. I guess the question becomes, how much do we want to generalize each plugin with regard to axes.

rirze:
@ejeschke @nabobalis I could play around with the media buttons and add more functionality, such as playing through a particular axis or even a set of axes. 

Just to be clear, implementing something like that would be a different PR, right? (It would be separate from this one. )

nabobalis:
It can be added here.

ejeschke:
@rirze, I would add it to this PR.  May as well fix the problems with the control buttons.  I think this can be done by simply having a radio button to select the axis that will be "played" or using a drop down menu--whatever feels/looks better.

Cadair:
or you could put a play button at the end of each slider for each axis?

ejeschke:
@Cadair, I think that might be a bit confusing. What happens if they start clicking multiple play buttons?  I think it is more logical to have a single set of controls.

This gets me thinking though. Maybe the playing controls and movie stuff ought to be in a separate tab.  Is the current layout getting too crowded?

ginga/misc/plugins/MultiDim.py
```diff
@@ -254,6 +260,27 @@ def build_naxis(self, dims):
...
+                if self.play_axis < len(dims):
+                    self.play_max = dims[self.play_axis]
+                self.play_idx = self.play_indices[n - 2]
+            return play_axis_change
+
+        for n in range(2, len(dims)):
+            key = 'axis%d' % (n + 1)
+            # the following line doesn't trigger, hence, the hack in the following line
+            # self.w[key].set_callback('activated', play_axis_change_func_creator(n) )
+            self.w[key].widget.toggled.connect(play_axis_change_func_creator(n))
```
rirze: For some reason, `set_callback` doesn't activate the function given. Instead, for proof of concept, I used a hack that only works for the qtw radiobutton.

ginga/misc/plugins/MultiDim.py
```diff
@@ -254,6 +260,27 @@ def build_naxis(self, dims):
         # Add vbox of naxis controls to gui
         self.naxisfr.set_widget(w)
 
+        # for storing play_idx for each dim of image. used for going back to
+        # the idx where you left off.
+        self.play_indices = [0 for i in range(len(dims) - 2)]
+
+        def play_axis_change_func_creator(n):
```
rirze: What's the best place to put this? Should it be inside or outside this function?

ginga/misc/plugins/MultiDim.py
ejeschke: @rirze, use `add_callback('activated', callable)`.  `callable` should have the function signature `(widget, value)`: `value` will be True when the radio button is activated.  To make all the radio buttons belong to the same group (so they are mutually exclusive), create the first one and then pass that as the kwarg parameter `group` to all the others.

ginga/misc/plugins/MultiDim.py
ejeschke: Doesn't matter much unless you are capturing lexical environment for use in the function.

rirze:
So I've added a group of RadioButtons for selecting which axis to play. They're mutually exclusive, even though I didn't set a group when creating them (I created them by passing them into `build_captions`, so I don't know how pass a `group` kwarg into it). I also commented out the spectral value, since the slice value would reflect the current image anyways. Let me know what you guys think.

ejeschke:
Looks good, @rirze.  I like how it removes the buttons when there is no option needed: for example, if I first view a 4D file I get the radio buttons. If I then drag a 2D file in they disappear. If I then drag a 3D file in they are still gone, but the player controls are active for the 3rd axis.  Nice.

Can others test and give feedback, please?

ginga/misc/plugins/MultiDim.py
```diff
@@ -415,24 +455,23 @@ def set_naxis(self, idx, n):
             if image is None:
                 raise ValueError("Please load an image cube")
 
-            # invert index
-            naxislen = 2 + len(image.naxispath)
-            m = naxislen - (n+1)
+            m = n - 2
```
ejeschke: @rirze, it would be best not to assume the loop variable has a certain value after exiting the loop (it would be different if it was a non-for variable).  I think it would be more readable and maintainable to say:
```python
n = len(dims) - 1
m = n - 2
```

rirze: But here, `n` is passed in, is it not?

ejeschke: Whups, sorry, you are right.  I was looking at the diff and missed the context.

ejeschke:
Wondering @rirze whether you can fix one final thing: when the axis radio buttons are built, no axis is preselected in the radio buttons.  It seems that it plays axis 3 though from the controls.  Probably the radio button should be set just after building the controls to indicate to the user which axis is selected.

Can you fix this?  Then I think we are ready to merge unless there is any other feedback.

ejeschke:
Merging.

Cadair:
ha! just as I pulll the branch down!

Thanks a lot @rirze !!

ejeschke:
Looks like the problem still exists of skipping certain frames when dragging the sliders for each axis.  I'm going to file that as a separate issue.

ejeschke:
> Looks like the problem still exists of skipping certain frames when dragging the sliders for each axis.

This is fixed in commit 310586d79c23b6936437a8390cc2f60b10f050ac

PR #300, opened 2016-03-24
Reorganized plugins doc

pllim:
Changes include:

* Created a subsection for each plugin.
* Added previously missing plugins sections.
* Moved some plugins sections to the correct category (e.g., global plugins erroneously filed as local in the doc).
* Moved commented `Pipeline` to the end of the list of plugin loading order. This does not affect existing functionality. It is done merely to be consistent with sub-sections order.


ejeschke:
Thanks, @pllim!

PR #301, opened 2016-03-24
Some fixes to Cuts 

Cadair:


ejeschke:
@Cadair, thanks for the fix!

Fixes the Slit not updating correctly when NAXIS sliders are updated in `MultiDim`.

ping @nabobalis, @rupak0577

PR #305, opened 2016-03-25
Enable auto insert of cfg into docstring

pllim:
As discussed during sprint session. This automatically inserts config file contents from examples into plugins doc.

### DIFF
--- a/ginga/misc/plugins/ChangeHistory.py
+++ b/ginga/misc/plugins/ChangeHistory.py
@@ -256,3 +256,10 @@ class ChangeHistory(GingaPlugin.GlobalPlugin):
 
     def __str__(self):
         return 'changehistory'
+
+
+# Replace module docstring with config doc for auto insert by Sphinx.
+# In the future, if we need the real docstring, we can append instead of
+# overwrite.
+from ginga.util.toolbox import generate_cfg_example
+__doc__ = generate_cfg_example('plugin_ChangeHistory', package='ginga')
--- a/ginga/util/toolbox.py
+++ b/ginga/util/toolbox.py
@@ -4,6 +4,18 @@
 # This is open-source software licensed under a BSD license.
 # Please see the file LICENSE.txt for details.
 #
+from __future__ import (absolute_import, division, print_function,
+                        unicode_literals)
+
+# STDLIB
+import io
+import os
+import warnings
+
+# THIRD-PARTY
+from astropy.utils.data import get_pkg_data_contents
+from astropy.utils.exceptions import AstropyUserWarning
+
 
 class ModeIndicator(object):
     """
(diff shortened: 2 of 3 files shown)


ginga/util/toolbox.py
```diff
@@ -89,4 +101,59 @@ def _configure_cb(self, view, width, height):
...
+    cfgname = config_name + '.cfg'
+
+    try:
+        cfgdata = get_pkg_data_contents(
+            os.path.join(cfgpath, cfgname), **kwargs)
+    except Exception as e:
+        warnings.warn(str(e), AstropyUserWarning)
+        return ''
+
+    userpath = '~/.ginga'  # Where user should save it
```
ejeschke: This may cause problems on some plaftorms (e.g. Windows).  Also, user can override the default home directory location by defining `GINGA_HOME`.  So the proper way to do this is:
```python
from ginga.util.paths import ginga_home
```
Then `ginga_home` is the desired directory prefix.

pllim: @ejeschke , inserting `ginga_home` here will insert my own home directory in the RST documentation, which is not desirable. The `~` is only for documentation purpose. I'll add a commit to clarify this shortly.

pllim:
Here is the generated RST for `ChangeHistory` plugin, as part of this PR:

screenshot

ejeschke:
It looks reasonable.  Hopefully people running Windows will be able to figure out where to put their config files.

pllim:
At its current state, it is consistent with the actual content of the config files. If Windows users complain about this, we can always add more clarification in the function.

ginga/misc/plugins/ChangeHistory.py
```diff
@@ -256,3 +256,10 @@ def start(self):
 
     def __str__(self):
         return 'changehistory'
+
+
+# Replace module docstring with config doc for auto insert by Sphinx.
+# In the future, if we need the real docstring, we can append instead of
+# overwrite.
+from ginga.util.toolbox import generate_cfg_example
+__doc__ = generate_cfg_example('plugin_ChangeHistory', package='ginga')
```
ejeschke: @pllim, so once we merge this, this becomes the boilerplate that we add to the bottom of the other plugins?

pllim: Yes. For each plugin, we will need to add an incantation to its RST file, and this to the actual code (both of which are applied to `ChangeHistory` as part of this PR to illustrate its usage). Note that I am completely overwriting the plugin's module docstring here. Currently this is not an issue because the documentation does not use it, so instead I use it to auto-generate the config file.

Another alternative is to actually document the entire plugin completely within the module docstring (i.e., the intro, usage, screenshot, etc.). In that case, we *append* the cfg contents instead using `__doc__ += ...`. However, I didn't do it because it requires more work porting existing plugin documentation back to the code.

ejeschke: It might take a little time to update all the plugins--plugins without the incantation just show the documentation without the config file?

pllim: At its current state. This is what will happen:

Incantation on `.rst` | Incantation on `.py` | Result
---- | ----- | -----
Yes | Yes | Config file shown as expected
Yes | No | Shows actual module docstring instead (not desirable)
No | Yes | No docstring shows
No | No | No docstring shows

ejeschke: Thanks for the helpful matrix!  Merging.

Issue #302, opened 2016-03-24
Emit user facing error when mosaic falls out of cache


pllim:
When there is a mosaic in cache but `numImages` are so small that the mosaic falls out of cache when user visits/opens other images, the error message should be displayed explicitly in `Errors` plugin. Right now, it only goes to Ginga's logfile.

Issue #297, opened 2016-03-24
Mosaic should not error when active mosaic is deleted


pllim:
Currently if `Mosaic` plugin is active, and the current mosaic is deleted by user, and then user drags another file into mosaic, it will give the following error in the log:

```python
2016-03-23 20:07:35,445 | E | GwMain.py:86 (update_pending) | gui error: 'mosaic2'
2016-03-23 20:07:35,449 | E | GwMain.py:90 (update_pending) | Traceback:
  File ".../ginga/gw/GwMain.py", line 81, in update_pending
    res = future.thaw(suppress_exception=False)
  File ".../ginga/misc/Future.py", line 40, in thaw
    res = self.method(*self.args, **self.kwdargs)
  File ".../ginga/misc/plugins/Mosaic.py", line 314, in _inline
    iminfo = self.chinfo.get_image_info(self.img_mosaic.get('name'))
  File ".../ginga/Control.py", line 1378, in get_image_info
    return self.image_index[imname]
```

`Mosaic` should listen to `remove-image` callback and performs a clean up if the image being removed is the active mosaic.

PR #308, opened 2016-03-25
Graceful clean up when mosaic is deleted or falls out of cache

pllim:
Graceful clean up when mosaic is deleted or falls out of cache. @ejeschke , here is my idea but it is not fully tested. What do you think?

This attempts to address #302 and #297.

### DIFF
--- a/ginga/misc/plugins/Mosaic.py
+++ b/ginga/misc/plugins/Mosaic.py
@@ -4,12 +4,17 @@
 # This is open-source software licensed under a BSD license.
 # Please see the file LICENSE.txt for details.
 #
+
+# STDLIB
 import math
 import time
-import numpy
 import os.path
 import threading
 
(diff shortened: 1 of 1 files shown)


ginga/misc/plugins/Mosaic.py
```diff
@@ -391,6 +402,50 @@ def drop_cb(self, canvas, paths, *args):
...
+        if not self.gui_up:
+            return False
+
+        # Is the image being removed a mosaic?
+        self.fv.error_wrap(self._mos_check, name)
+
+    # TODO: Do we really need this?
+    def delete_channel_cb(self, viewer, channel):
+        """Clean up when the whole channel is deleted."""
+        self.img_mosaic = None
```
pllim: Do we need to explicitly do this? Or does Python automatically garbage collect?

PR #313, opened 2016-03-29
Applied Astropy affiliated package template and fixed API doc

### LINKED ISSUE #56: Behave more like an Astropy affiliated package
I think I mentioned this before ... it would be nice if the following things that work for other Astropy affiliated packages worked for Ginga (see https://github.com/astropy/package-template how it's implemented):

* `python setup.py test --coverage`
* `python setup.py build_sphinx -w`

(see uncommented lines in https://github.com/ejeschke/ginga/blob/master/.travis.yml )

pllim:
Applied Astropy affiliated package template (fixes #56). Added `__all__` to some modules so API doc would build correctly. Added extra API doc for `canvas` for cross-linking to work.

I did a similar change to `stginga` (spacetelescope/stginga#108) and thought I will do the same here, to be consistent between the two, since `stginga` depends on Ginga.


pllim:
p.s. This might look scary but most changes are according to http://astropy.readthedocs.org/en/latest/development/affiliated-packages.html?packagename=ginga#managing-the-template-files-manually ; i.e., copy and paste, followed by some customization based on existing stuff.

ejeschke:
`sunpy` should be an optional dependence.  As such, if the import fails then the tests using it should be skipped so that it doesn't flag the whole build as a fail.

pllim:
Thanks, @ejeschke . I didn't even know Ginga imports SunPy. I didn't touch any existing tests. I merely applied the test template from astropy helpers.

One thing I forgot to mention is that using the template, documentation will have Astropy theme, not RTD theme. Are you okay with this? Or should I overwrite the theme in `conf.py`?

pllim:
Two more things:

**Versioning using the template**

The version is now recorded in `setup.py`, as `ginga/version.py` is automatically generated during build and no longer under version control. I updated the version number in this PR to match your latest commit (bfb6e08).

**Failed tests**

I figured out what's wrong with the tests and submitted a fix to Astropy (astropy/astropy#4748). After the fix, here are the test results (ran locally on my machine using Python 2.7):

```
downtonabbey> python setup.py test
Freezing version number to ginga/version.py
running test
running build
running build_py
copying ginga/version.py -> build/lib.linux-x86_64-2.7/ginga
running build_scripts
============================= test session starts ==============================
platform linux2 -- Python 2.7.11, pytest-2.8.5, py-1.4.31, pluggy-0.3.1

Running tests with ginga version 2.5.20160330023527.
Running tests in lib.linux-x86_64-2.7/ginga doc.

Date: 2016-03-30T13:20:07

Platform: Linux-2.6.32-573.18.1.el6.x86_64-x86_64-with-redhat-6.7-Santiago

Executable: ~/anaconda/bin/python

Full Python Version: 
2.7.11 |Anaconda 2.3.0 (64-bit)| (default, Dec  6 2015, 18:08:32) 
[GCC 4.4.7 20120313 (Red Hat 4.4.7-1)]

encodings: sys: ascii, locale: UTF-8, filesystem: UTF-8, unicode bits: 20
byteorder: little
float info: dig: 15, mant_dig: 15

Numpy: 1.10.4
Scipy: 0.16.1
Matplotlib: 1.4.3
Pandas: 0.16.2
Astropy: 1.2.dev14847
scikit-image: 0.11.3

rootdir: /tmp/ginga-test-4QcSH3, inifile: setup.cfg
collected 99 items 

ginga/misc/tests/test_Callback.py .........................
ginga/misc/tests/test_Future.py ......................
ginga/misc/tests/test_Task.py ........
ginga/tests/test_ImageView.py ..
ginga/tests/test_cmap.py ........
ginga/tests/test_colors.py ..................
ginga/tests/test_imap.py ........
ginga/tests/test_wcs.py ........

================ 99 passed, 6 pytest-warnings in 19.52 seconds =================
```

ejeschke:
@pllim, I am ok with the astropy theme.

ejeschke:
> I figured out what's wrong with the tests and submitted a fix to Astropy (astropy/astropy#4748).

@pllim, I was thinking that the documentation should be moved into the package that is installed (as it presently is not).  Should we just move it into `.../ginga/docs`?

ginga/ImageView.py
```diff
@@ -23,6 +23,8 @@
 from ginga.canvas.types.layer import DrawingCanvas
 from ginga.util import io_rgb
 
+__all__ = ['ImageViewBase']
+
```
pllim: AFAIK `__all__` only affects `from module import *`, which I think is what was tripping the `automodsum` directive that comes with the package template, because without this, it was attempting to find random things like `zip`, `map`, functions from another module that were imported, etc. As long as you don't do `... import *` in your own modules when you want to subclass etc (which is not recommended anyway), you should be okay.

pllim: Related reading: http://stackoverflow.com/questions/44834/can-someone-explain-all-in-python

pllim:
@ejeschke , what is your motivation to install the documentation itself? `doc` is already listed in the manifest, so it will be distributed. According to the template, documentation does not live under `packagename`, where the actual codes live.

ejeschke:
> According to the template, documentation does not live under packagename, where the actual codes live.
Well that's my misunderstanding then.  If so, no need to move it.

ginga/ImageView.py
ejeschke: Ok, no worries then.

ejeschke:
> doc is already listed in the manifest, so it will be distributed.

Does this mean that the documentation is installed if the user installs via `pip`?  Into the package installation area?

pllim:
>  Into the package installation area?

Actually, no. It means `doc` will be distributed with the source. But not actually installed under `site-packages`. Does it have to be? User can look up the documents on RTD.

ejeschke:
It's nice to have the documentation installed.  Then it makes it easy to show the documentation inside the viewer.  Or sometimes users don't have internet access, like on a plane, odd locations, etc.  I've had one complaint about it not being installed with the package some time ago.

pllim:
> It's nice to have the documentation installed.

Do you mean the `*.rst` files, or the generated `*.html` files?

pllim:
If you meant for users to access it offline, you probably want the generated HTML files. However, the HTML files are only generated when someone `git clone` the repo, install Ginga, and then run `make html` under `doc` directory. 

If you want to actually install the HTML files, that probably means you intend to pre-generate them and include them as package data (hence under version control). This runs the risk of local HTML files update lagging behind RTD version, because someone who updates the code might not be aware that he/she also has to re-generate the HTML files and copy them under package data directory. Hmm...

pllim:
@ejeschke , the fix is in Astropy master now. Do you want me to update `.travis.yml` in this PR to use dev instead of stable Astropy?

pllim:
@ejeschke , regarding offline documentation, I think we should address that as a separate issue from this because it is not part of the Astropy template. Do you agree?

pllim:
p.s. Accidentally closed this. Reopened.

pllim:
All tests passed (using Astropy dev).

Success Kid

pllim:
@ejeschke , if possible, please review and merge this soon. Because everytime you bump release on `version.py`, there is a conflict. In the template, `version.py` is a *generated* file, thus not under version control.

ejeschke:
Hmm, not sure about maintaining the version/release number in `setup.py`.  That means I'll end up modifying that frequently just to make a release.  Does the template require this to be kept in the `setup.py`, or can I just move it out of `version.py` into something else and still have `version.py` auto-generated?

ejeschke:
@pllim, if the `astropy-helpers` package is updated, how complicated is the procedure to merge the changes into `ginga`, if we adopt it as implemented in this PR?

pllim:
Here are the instructions from the doc:
```
Updating to the latest template files
-------------------------------------

From time to time we will make changes to the package-template to fix bugs or
add functionality. Updating to the latest version is simple - simply check
the `TEMPLATE_CHANGES.md`_ file, which provides a changelog of the package
template. You can also re-copy over all the files listed in the above section
and see if any of the changes should be committed (some of the changes will
be reverting some of your edits, so do not include those!). Remember to
update the astropy-helpers sub-module to the latest stable version, and
update the corresponding ``ah_bootstrap.py`` file, for example::

    cd astropy_helpers
    git fetch origin
    git checkout v0.4.3
    cd ..
    cp astropy_helpers/ah_bootstrap.py .
    git add astropy_helpers ah_bootstrap.py
    git commit -m "Updated astropy-helpers to v0.4.3"

You can find out what the latest version of astropy-helpers is by checking the
`astropy-helpers <https://pypi.python.org/pypi/astropy-helpers/>`__ entry on
PyPI.
```

pllim:
Re: versioning

It won't be hard to avoid changing `setup.py` to bump version number. In fact, there is a PR to move versioning in the template to `setup.cfg` (see astropy/package-template#129). Will this work for you?

ejeschke:
I'll try to merge this in the next couple days.  I suppose this fixes the travis build issue with "build_sphinx"?

pllim:
@ejeschke 

>  I suppose this fixes the travis build issue with "build_sphinx"?

Interesting enough, Sphinx failed with the same error as my other PRs in my earlier commit today to this PR (5 commits ago as I wrote this). So, I updated `astropy_helpers` to the current commit on `master` (it points to that commit, not `master` itself for more stability), and ported some recent changes from `package-template` over (the biggest change among that is update to `.travis.yml`). On the template, it no longer forces `linkchecker` when `build_sphinx` test is run, so I removed it from here as well (did you added the `linkchecker` yourself?). The tests now use `astropy/ci-helpers`, so on the whole, the YAML looks cleaner. I am waiting to see if this fixes the failed test.

pllim:
Update: The tests all passed now. See https://travis-ci.org/ejeschke/ginga/builds/127823559

Note that I removed the `linkchecker` command. If it is https://wummel.github.io/linkchecker/ , note that its build is failing, so maybe that was the problem? Do you want me to add this back in Travis CI?

I also rebased this PR to `master` (as of this morning).

ejeschke:
Ok, here goes!

pllim:
:clap:

PR #314, opened 2016-04-05
Call update_value/update_html in Label, ContainerBase, and Box widgets

rdk1024:
In the web version of Widgets.py, we need to call the "do_operation" method and invoke the update_value/update_html operations to update the web page.


ejeschke:
Looks good

PR #318, opened 2016-04-07
Add SaveImage local plugin

### LINKED ISSUE #312: New plugin to save image(s)
This was brought up by @ejeschke in #292, *"... let's add a 'Save As FITS' menu item that starts the 'Save Image' plugin (which you should submit as a separate PR which we can merge first). That way after creating the mask they can either start the save image plugin from the Operations menu or they will discover it under the File menu."*

From the description above, it sounds like the idea is to keep `SaveImage` as a local plugin, as it currently is in `stginga`. But if it were to be started via `File` menu item, shouldn't it be a global plugin instead? Please clarify, @ejeschke .

Screenshot of current implementation:

saveimage_screenshot

pllim:
Added `SaveImage` local plugin and related doc. Also added text to `FBrowser` global plugin doc.

This fixes #312


ginga/misc/plugins/SaveImage.py
```diff
@@ -0,0 +1,477 @@
...
+        return
+
+    def save_images(self):
+        """Save selected images."""
+        res_dict = self.treeview.get_selected()
+        clobber = self.settings.get('clobber', False)
+        self.treeview.clear_selection()  # Automatically disables Save button
+
+        # If user chooses fitsio over astropy, cannot continue.
+        if (io_fits.fitsLoaderClass.__name__ !=
```
pllim: This is not strictly necessary, but it is better to alert user that their chosen `fitsio` is not used?

ginga/main.py
```diff
@@ -108,6 +108,7 @@
     Bunch(module='Operations', tab='_opns', ws='operations', start=True),
     Bunch(module='WBrowser', tab='Help', ws='channels', raisekey='?', start=False),
     Bunch(module='FBrowser', tab='Open File', ws='right', start=False),
+    Bunch(module='SaveImage', tab='Save File', ws='right', start=False),
```
pllim: This is slightly different from "Save as FITS" that you suggested but it is more consistent with "Open File".

ginga/misc/plugins/SaveImage.py
```diff
@@ -0,0 +1,477 @@
...
+
+            # Insert new keyword
+            if key not in hduhdr:
+                hduhdr[key] = (bnch.value, bnch.comment)
+
+            # Update existing keyword
+            elif hduhdr[key] != bnch.value:
+                hduhdr[key] = bnch.value
+
+    def _write_mosaic(self, key, outfile):
```
pllim: I thought about moving I/O related methods to `ginga.util.io_fits` module but decided against it because the functionality here also depends on other plugins (e.g., inserting history from `ChangeHistory` plugin), so things get a bit tricky to generalize.

ejeschke:
@pllim, does this only offer to save modified images?  Suppose I have just loaded a FITS file from a URL (dragged a URL to the window) and now I want to save it locally.  Can I do that with this plugin?

It says "nothing available for saving" at the bottom.

ginga/examples/configs/plugin_SaveImage.cfg
```diff
@@ -0,0 +1,22 @@
...
+
+# Default output parameters. Can also be changed in the GUI.
+output_directory = '.'
+output_suffix = 'ginga'
+
+# Clobber existing output files or not
+clobber = False
+
+# Only list modified images from the channel
+modified_only = True
```
pllim: @ejeschke , to answer your question, if you set `modified_only = False`, it lists *all* images. Does this work for you?

ejeschke: Sorry, didn't notice that, right off that bat.  Yes, it works, but wondering whether the general (non-editing) user will be confused if the default is not `False`.

pllim: Does user generally want to save *unmodified* files? Why don't they just do a system copy in that case? I was under the impression that `True` would make more sense, but I can certainly change it to `False` if you think otherwise. Let me know!

ginga/misc/plugins/SaveImage.py
```diff
@@ -0,0 +1,477 @@
...
+        self.chname = self.chinfo.name
+
+    def keys_from_datasrc(self):
+        """Yield back keys for image listing from Ginga's data cache."""
+        for key in self.chinfo.datasrc.sortedkeys:
+            yield key
+
+    def keys_from_contents(self):
+        """Yield back keys for image listing  from ``Contents`` global plugin.
+        """
```
ejeschke: I think you should be able to get all images by scanning the channel lists from `Control.py`.  That way if there is a configuration where the user doesn't use `Contents`, or has replaced it with their own, it would still work.  Code would be something like:
```python
chnames = self.fv.get_channelNames()
imnames = []
for chname in chnames:
    channel = self.fv.get_channelInfo(chname)
    imnames.extend(channel.get_image_names())
return imnames
```
You could use `yield` here if you want to write it that way, but since the list is being created, consumed and then dropped I'm not sure it really makes any difference except that the non-`yield` version should be faster.

ejeschke: BTW, would it make sense to show the channel name in the `SaveImage` tree?  Seems like that would be helpful. What do you think?

ginga/misc/plugins/SaveImage.py
ejeschke: Can't then load the image via `fitsio`?  It looks like you are importing `astropy.io.fits` in this plugin so that is a dependency to use the plugin (fine), but it shouldn't really matter how the image came in, right?

ginga/misc/plugins/SaveImage.py
pllim: @ejeschke 

Yes, it make sense to replace getting the names from `Control` instead of `Contents`. I'll add a commit to do that when I get the chance. Thanks!

As for showing channel name, I do not think it is necessary because `SaveImage` is a *local* plugin. There is already a `Channel: <channel name>` label at the top of the GUI to show which channel it is tied to.

ginga/misc/plugins/SaveImage.py
ejeschke: Just tested.  It works fine if the user has `fitsio` selected as their default in general preferences.  I think we could/should remove this test.

pllim: No, this method has nothing to do with loading images and it should not stop users from using `fitsio`. This error is meant to tell the user that they cannot use `fitsio` to save the images. So, basically, everything works except for the Save button.

ejeschke: Oh, and if we have a user that is opening lots of images with a new format like "asdf" or HDF5 or something, they probably will have that set as their general preference.  So we shouldn't discriminate how they got the image into ginga.

pllim: @ejeschke , so you don't think they will be confused nor care that their images are loaded with `fitsio` but saved with `astropy.io.fits`? If not, then I'll be happy to remove this test.

ejeschke: No, general user won't care.

pllim: Okay, I'll remove this in my next commit. Thanks!

ginga/misc/plugins/SaveImage.py
ejeschke: Ah, but if it is a local plugin then we should only really be showing images from the local channel, no?

pllim: @ejeschke , yes, it only shows contents from the local channel. In the next line, you will see
```python
self.list_plugin_obj.name_dict[self.chname]
```
that only extracts that one channel. So, my plan to is modify your example for `Control` above for local plugin usage.

ejeschke: It looks like you have designed the plugin to run either locally or globally.  If invoked locally, then it would make sense to show only the list of images in the channel it is associated with--user should not be confused if they understand the general difference between local/global. If invoked globally, or we are making the decision to allow any image to be saved then it might be good to show which channel the image came from.

One simple way would be to simply prepend a prefix when throwing the image name in the tree like:
`Image/u40x010hm_c0h.fits` or `Image: u40x010hm_c0h.fits`.

pllim: @ejeschke , when in the "global" mode, the `Channel: <channel name>` text label would change. I think I wrote that in the doc somewhere. Let me find it and tag you there.

doc/manual/plugins_global/saveimage.rst
```diff
@@ -0,0 +1,14 @@
+.. _sec-plugins-global-saveimage:
+
+SaveImage (Save File)
+=====================
+
+This brings up :ref:`local-plugin-saveimage` local plugin from the global
+Plugins menu for convenience.
+
+When done this way, the plugin is locked to the currently active image channel.
+To switch to another channel, follow these steps:
```
pllim: @ejeschke , perhaps this is not clear enough. But when user follows the steps below, the plugin will update the channel and list only images associated with that channel.

ginga/misc/plugins/SaveImage.py
ejeschke: I think I possibly understand now.  Did you write it so that it could be invoked as a global plugin in order to invoke it from the "File" menu?  It's possible to invoke a local plugin from a menu.

I think as long as we only ever run it as a local plugin we only need to show the per channel image names.

pllim: > Did you write it so that it could be invoked as a global plugin in order to invoke it from the "File" menu?

Yes, it is to be like `FBrowser` like you suggested a few weeks back. No, not the "File" menu, but the "Plugins" menu like `FBrowser`.

So, it is *always* a local plugin, but can be started from either the local or global menus.

pllim:
@ejeschke , I think I addressed all your comments? Please let me know if I missed anything.

ginga/misc/plugins/SaveImage.py
```diff
@@ -0,0 +1,462 @@
...
+            if os.path.exists(outfile) and not clobber:
+                self.logger.error('{0} already exists'.format(outfile))
+                continue
+
+            bnch = res_dict[infile]
+
+            if bnch.path is None or not os.path.isfile(bnch.path):
+                self._write_mosaic(f_pfx, outfile)
+            else:
+                shutil.copyfile(bnch.path, outfile)
```
ejeschke: Does this make a backup copy of the file if there is one already there?

pllim: It overwrites the existing file. But by the time it gets to this line, it is already very clear that user wants to clobber the old file, because there is a check for clobber setting on Line 401 above.

ejeschke:
@pllim, I'm being a little picky about this one because I can imagine that it could get lots of use for different scenarios and we should make it as intuitive as possible to use.

Can you checkout the branch "pllim-save-image" from [naojsoft](https://github.com/naojsoft/ginga.git) and give it a whirl?  In this version you have to invoke it as a global plugin.  LMK your thoughts.

pllim:
@ejeschke , thanks. My thoughts on your changes are as follow:

* :+1: on the new checkbox. Very convenient! 
* This was originally a *local* plugin because I don't want users to accidentally overwrite a modified image from one channel with an unmodified one of the same name from another. Now that it is global, I wonder if it would be safer to have the suffix change depending on the channel, in addition to the user provided string. For example, `<input_prefix>_ginga_Image0.fits`. Or if user is only going to work on a single channel, then perhaps this extra suffix can be disabled via user configuration to preserve the original `<input_prefix>_ginga.fits` convention.
* Since this would be exclusively a *global* plugin, it would be better off living under the "File" menu. I previously didn't do that because it didn't make sense to stick a *local* plugin there.
* By sticking in a combobox, it made the text entry boxes a lot shorter. Might have to put the combobox in a separate container to preserve text entry box lengths. But do not need this if... see the next item.
* Could be even more intuitive (maybe?) if we get rid of the "Channel" combobox and just have the plugin automatically switch to the current active channel. That is, if I focus on `Image0` channel, it automatically lists images for that channel. This can be accomplished by simply hooking up your new `update_channels()` method to the channel change callback.
* I understand that your branch is only a proof of concept, but I am :+1: with going ahead with your changes. So, the docs and screenshot need to be updated as well.

How do you wish to proceed? Do you want me to move your change to my PR? Or file a PR to your branch that is based on my PR? :smile:

ejeschke:
- I'm fine with changing the suffix based on channel, if that works for you. 
- I agree we should definitely add a menu item for it to the `File` menu.  It does have to be done in such a way that if the plugin is not loaded, then menu item is not added.
- We can move the widgets around to get a better layout.
- I tried a version where the plugin responds dynamically to changing channels--it just didn't feel quite right.  I don't think a less experienced user will understand the need to change channels.  By putting the combobox control right there it makes it easy to browse to the image that you want to save.

As to proceeding, we could:
1) Since that branch is based on your PR, I'll can just resubmit it as a new PR and it will have your commits in it.  OR, 
2) you can try to cherry pick the commits off of it onto this PR and update this PR.

I understand how to do option 1 easily.  Option 2 probably requires a little git-fu that you/I would need to look up.

pllim:
If you're more comfortable with Option 1, then please go ahead. I can always file a follow up PR after you merged it if necessary. Thank you!

pllim:
@ejeschke , just to confirm, you are going to merge *your* branch, right? That is why I haven't been touching this PR of mine.

ejeschke:
I will submit that one as a PR to be merged.  But I want to make a couple small tweaks we discussed in this thread, mostly the File menu item and changing the widget layout to look better.

PR #319, opened 2016-04-07
Fix matplotlib colormaps not loading properly

pllim:
I found a bug where matplotlib colormaps were not being loaded even if requested in Python 2.7. It is because the colormap names were compared to `str` when they are actually `unicode`.

While I am at it, I also made sure the hot new "viridis" colormap is present:

screenshot

Also added some much needed docstrings.

### DIFF
--- a/ginga/cmap.py
+++ b/ginga/cmap.py
@@ -11481,6 +11481,8 @@ def add_matplotlib_cmaps():
     names = list(plt.cm.datad.keys())
 
     # New default colormap for matplotlib 2.0, for those who cannot wait.
+    # Somehow Viridis is not part of "plt.cm.datad" keys above for
+    # "transition" versions of matplotlib (e.g., v1.5.1).
     # See http://matplotlib.org/style_changes.html
     if 'viridis' not in names:
         names.append('viridis')
@@ -11489,8 +11491,8 @@ def add_matplotlib_cmaps():
     for name in names:
         if not isinstance(name, six.string_types):
             continue
-        cm = plt.get_cmap(name)
         try:
+            cm = plt.get_cmap(name)
             add_matplotlib_cmap(cm, name=name)
         except Exception as e:
             print("Error adding colormap '%s': %s" % (name, str(e)))
--- a/ginga/cmap.py
+++ b/ginga/cmap.py
@@ -7,10 +7,16 @@
 # This is open-source software licensed under a BSD license.
 # Please see the file LICENSE.txt for details.
 #
-from __future__ import print_function
+from __future__ import absolute_import, print_function
+from .util import six
+from .util.six.moves import map, zip
 
 import numpy
-from ginga.util.six.moves import map, zip
+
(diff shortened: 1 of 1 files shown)


ginga/cmap.py
```diff
@@ -11421,55 +11427,76 @@
...
 
-    for name in plt.cm.datad.keys():
-        if not isinstance(name, str):
+    names = list(plt.cm.datad.keys())
+
+    # New default colormap for matplotlib 2.0, for those who cannot wait.
+    # See http://matplotlib.org/style_changes.html
+    if 'viridis' not in names:
+        names.append('viridis')
+        names.sort()
```
ejeschke: Can we be sure 'viridis' was loaded?  Does this work with e.g. older matplotlib versions.  I'm not sure how old we should support, but maybe something like 1.3.1?

pllim: Good point. Would changing
```python
cm = plt.get_cmap(name)
try:
    add_matplotlib_cmap(cm, name=name)
except Exception as e:
    ...
```
to
```python
try:
    add_matplotlib_cmap(cm, name=name)
    cm = plt.get_cmap(name)
except Exception as e:
    ...
```
be sufficient to address your concern?

ejeschke: Hmm, I guess it works because if it is not present it gets thrown out in the exception handler below.  @pllim, did you find it necessary to add this line to get 'viridis' to load with newer matplotlib versions?

pllim: I have 1.5.1, which I think is sufficiently new. I need that line to find Viridis. For whatever reason, it does not get included automatically.

pllim: You can double check to see if it is in `plt.cm.datad.keys()` in your own Matplotlib installation. If you find otherwise, please let me know, because then it could be something wrong with my Matplotlib installation.

ejeschke: I'm fine with leaving the line in, but maybe we should add a comment to the effect that the line is there because the colormap is somehow not picked up automatically with "transition" versions of matplotlib (or something to that effect).

pllim: Okay, I'll add that comment in this PR when I get the chance. Thanks!

pllim:
Anyways, I also opened [an issue with Matplotlib people about Viridis missing in the keys](matplotlib/matplotlib#6316).

pllim:
@ejeschke , I think I addressed your concerns in my third commit to this PR. Please let me know if it is sufficient. Thanks!

ejeschke:
Looks good!

ejeschke:
Regarding the "viridis" map, on matplotlib 1.3.1 I now get a prominent print message "Error adding colormap 'viridis': Colormap viridis is not recognized" when starting up the reference viewer.  I think that the code adding this name also needs to check whether this color map is actually available...

pllim:
Yikes, sorry for the trouble. I'll investigate a way to avoid a print out error tomorrow. One way is to only enable it for Matplotlib 1.5+ if I cannot find something more elegant.

PR #321, opened 2016-04-11
New TVMark and TVMask local plugins

pllim:
Added new `TVMark` and `TVMask` local plugins. Also added related docs.

This is a transfer from `stginga`, as requested during Python in Astronomy 2016 unconference in Seattle, WA.

c/c @r-owen


doc/manual/plugins_local/tvmask.rst
```diff
@@ -0,0 +1,53 @@
...
+To redraw the same masks with different color or alpha, press "Forget"
+and repeat the steps above, as necessary.
+
+If images of very different pointings/dimensions are displayed in the same
+channel, masks that belong to one image but fall outside another will not
+appear in the latter.
+
+To create a mask that this plugin can read, one can use results from
+:ref:`plugins-drawing` (press "Create Mask" after drawing and save the
+mask using ``SaveImage``), in addition to creating a FITS file
```
pllim: I added a mention of ``SaveImage`` in anticipation that #318 will be merged. If that one is merged first, I can even update this to point directly to the associated Sphinx doc ref.

ginga/misc/plugins/TVMask.py
```diff
@@ -0,0 +1,447 @@
...
+                        unicode_literals)
+from ginga.util.six import itervalues
+
+# STDLIB
+import re
+import os
+
+# THIRD-PARTY
+import numpy as np
+from astropy.io import fits
```
pllim: This bypasses `ginga.util.io_fits` but it is only used to load boolean masks for this plugin. Is this a problem for Ginga?

doc/manual/plugins_local/tvmask.rst
ejeschke: Let's merge `SaveImage` and then we can get to these and the mask creation addition.

ginga/misc/plugins/TVMask.py
ejeschke: It's no problem to have a plugin that depends on a package, especially `astropy`, which is for all intents and purposes a requirement for the reference viewer.  

In general it *is* nice if we can use the loading abstractions because it makes it easier 1) to test when there is a possible problem with one package (you can try the other), 2) to support new formats more easily with the same plugins.

pllim: @ejeschke , so what would be the best way to implement `fits.getdata(filename).astype(np.bool)` to satisfy such abstraction?

doc/manual/plugins_local/tvmask.rst
ejeschke: Merging, now that we merged `SaveImage`.

ginga/misc/plugins/TVMask.py
ejeschke: Let's look at that in a later PR, if necessary.

ejeschke:
Merging.

PR #322, opened 2016-04-19
Do not force loading Viridis colormap when it is not present

pllim:
Do not force loading Viridis colormap when it is not present (e.g., in Matplotlib v1.3 or earlier). The fix is to replace the colormap lookup with the correct dictionary that includes *all* the available colormaps. This is a follow-up of #319.

### DIFF
--- a/ginga/cmap.py
+++ b/ginga/cmap.py
@@ -11476,23 +11476,13 @@ def add_matplotlib_cmap(cm, name=None):
 
 def add_matplotlib_cmaps():
     """Add all matplotlib colormaps."""
-    import matplotlib.pyplot as plt
+    from matplotlib import cm as _cm
 
-    names = list(plt.cm.datad.keys())
-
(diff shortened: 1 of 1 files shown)


ginga/cmap.py
```diff
@@ -11476,23 +11476,15 @@ def add_matplotlib_cmap(cm, name=None):
 
 def add_matplotlib_cmaps():
     """Add all matplotlib colormaps."""
-    import matplotlib.pyplot as plt
+    from matplotlib import cm as _cm
```
pllim: Note: There is actually a *real* `matplotlib._cm` module that is not the same as `matplotlib.cm` but I renamed the import here as not to clash with your existing `cm` *variable* name. I don't think this is an issue because we will never need the actual `matplotlib._cm` module anyway.

ejeschke:
@pllim, thanks!

PR #325, opened 2016-04-26
Add SaveImage global plugin

ejeschke:
This subsumes the commits from #318, but turns it into a global plugin.


ejeschke:
@pllim, this still doesn't add the plugin to the `File` menu, but that can be added later.  I need to modify the `Menu` widget wrapper to allow insertion of menu items in addition to appending them.

ejeschke:
@pllim, can you test this PR?  If there are no major issues I think it is better to merge it now, because we have a couple other PRs waiting on this one.

doc/manual/plugins_local/saveimage.rst
```diff
@@ -0,0 +1,29 @@
+.. _local-plugin-saveimage:
+
+SaveImage
+=========
+
+.. image:: figures/saveimage_screenshot.png
```
pllim: Screenshot needs to be updated. Current image shows local plugin.

doc/manual/plugins_local/saveimage.rst
```diff
@@ -0,0 +1,29 @@
+.. _local-plugin-saveimage:
```
pllim: This documentation needs to move to `plugins_global/saveimage.rst`. This file needs to be deleted.

doc/manual/plugins_global/saveimage.rst
```diff
@@ -0,0 +1,14 @@
+.. _sec-plugins-global-saveimage:
+
+SaveImage (Save File)
+=====================
+
+This brings up :ref:`local-plugin-saveimage` local plugin from the global
```
pllim: This documentation should be updated and also include the relevant contents from `plugins_local/saveimage.rst`.

doc/manual/plugins.rst
```diff
@@ -80,5 +81,6 @@ is the graphical way to start an operation.
    plugins_local/mosaic
    plugins_local/drawing
    plugins_local/fbrowser
+   plugins_local/saveimage
```
pllim: Delete this line, as it does not apply to global plugin.

ginga/misc/plugins/SaveImage.py
```diff
@@ -0,0 +1,493 @@
...
+        self.tw.set_text("""Enter output directory and suffix, if different than default. Left click to select image name to save. Multiple images can be selected using click with Shift or CTRL key. Click Save to save the selected image(s).
+
+Output image will have the filename of <inputname>_<suffix>.fits.""")
+
+    def redo(self, *args):
+        """Generate listing of images that user can save."""
+        if not self.gui_up:
+            return
+
+        #mod_only = self.settings.get('modified_only', True)
```
pllim: The commented line can be deleted since we decided to keep the checkbox. Otherwise, it is just going to confuse people in the future.

ginga/misc/plugins/SaveImage.py
```diff
@@ -0,0 +1,493 @@
...
+        self.redo()
+
+    def select_channel_cb(self, w, idx):
+        self.chname = self.chnames[idx]
+        self.logger.debug("channel name changed to '%s'" % (self.chname))
+        self.redo()
+
+    def keys_from_datasrc(self):
+        """Yield back keys for image listing from Ginga's data cache."""
+        channel = self.fv.get_channelInfo()
```
pllim: Should we change this to `channel = self.fv.get_channelInfo(self.chname)` to be consistent with other parts of this plugin?

pllim: Alternately, if you want to pass in `channel` directly into this method to avoid duplicate codes, that is also possible but it has to return a list, not yield; And also the `func = blahblah` and using `func()` in the loop need to be changed.

ginga/misc/plugins/SaveImage.py
```diff
@@ -0,0 +1,493 @@
...
+            # Write history to PRIMARY
+            self._write_history(key, pf['PRIMARY'])
+
+    def toggle_save_cb(self, w, res_dict):
+        """Only enable saving if something is selected."""
+        if len(res_dict) > 0:
+            self.w.save.set_enabled(True)
+        else:
+            self.w.save.set_enabled(False)
+        return
```
pllim: The `return` can be removed as it does not really do anything. Probably I forgot to delete it.

ginga/misc/plugins/SaveImage.py
```diff
@@ -0,0 +1,493 @@
...
+
+        # Resize column widths
+        n_rows = len(treedict)
+        if n_rows == 0:
+            self.w.status.set_text('Nothing available for saving')
+        elif n_rows < self.settings.get('max_rows_for_col_resize', 5000):
+            self.treeview.set_optimal_column_widths()
+            self.logger.debug('Resized columns for {0} row(s)'.format(n_rows))
+
+        return
```
pllim: The `return` can be removed as it does not really do anything. Probably I forgot to delete it.

ginga/misc/plugins/SaveImage.py
```diff
@@ -0,0 +1,493 @@
...
+
+        # Process each selected file. Each can have multiple edited extensions.
+        for infile in res_dict:
+            f_pfx = os.path.splitext(infile)[0]  # prefix
+            f_ext = '.fits'  # Only FITS supported
+            oname = f_pfx + sfx + f_ext
+            outfile = os.path.join(self.outdir, oname)
+
+            s = 'Writing out {0} to {1} ...'.format(infile, oname)
+            self.w.status.set_text(s)
```
pllim: This can stretch out the plugin widget's width if filename is very long (which is probably true for all JWST images). Should we consider using `ginga.util.iohelper.shorten_name()` here for the informational text?

pllim:
@ejeschke , I left my comments. Also general comments below:

* Are the inclusion of toolkit-specific menu widget stuff intentional? They were not part of my original PR for this plugin.
* Commit history is a bit messy, but not sure if it matters.
* Travis CI failure might not be related to this PR.

ejeschke:
@pllim, thanks for the edits, good stuff!

I've updated the PR with most of these edits except for a couple things which I will comment on above.  

The commit history is messy because I think I picked up some unrelated commits. I'm a little nervous to try to squash commit at this point because I'm working on a PR based on a different PR involving multiple repos--probably it would have been better to just have you update your original PR with commits from *my* PR (which I could then delete) and keep working from your PR--I'll remember this for next time.

Travis failure is unrelated--it seems to time out occasionally trying to do debian installs of astropy--not sure what is going on there but I think that stuff will get sorted once we merge the astropy-helpers PR, since they will be sorting out the same issues.

doc/manual/plugins_local/saveimage.rst
ejeschke: I'm going to have you redo the screenshot, because I think you can put together a more interesting one based on a real workflow that you have.  Let's just handle this and any other final tweaks as a separate, new PR.

ginga/misc/plugins/SaveImage.py
ejeschke: I'll let you fix this up if you want.  Personally, I think it complicates reading code to use `yield` in cases where a not very large iterable is built.  And it may slow down the code as well slightly.  But I'm not overly concerned about it in cases where the code is not in a hotspot.

ginga/misc/plugins/SaveImage.py
ejeschke: I'll leave this up to you.

doc/manual/plugins_global/saveimage.rst
```diff
@@ -3,12 +3,27 @@
 SaveImage (Save File)
 =====================
 
-This brings up :ref:`local-plugin-saveimage` local plugin from the global
-Plugins menu for convenience.
+.. image:: figures/saveimage_screenshot.png
```
pllim: The screenshot needs to be moved to `manual/plugins_global/figures/` directory. Sphinx cannot find that image and it is causing Travis CI failure.

pllim:
@ejeschke , except the wrong directory for screenshot, you can certainly merge this (when the image is moved to the correct directory) and I will create a follow-up PR to address the remaining issues. Thank you!

ejeschke:
Yeah, just caught that from the travis build.  Updated.  Give it a quick spin with your workflow to make sure everything is working fine and if so I'll merge it.

ejeschke:
Ok, I will merge and let's sort out any new issues in a separate PR.

pllim:
I started working on follow-up fixes. Will submit PR soon. FYI.

Issue #330, opened 2016-04-29
SaveImage cannot find ChangeHistory


pllim:
In the midst of working on a follow-up PR for `SaveImage` (#331), I realized that it can no longer find `ChangeHistory` global plugin. To reproduce the problem:

1. Comment out the `try ... except ...` block in `SaveImage.__init__()` so that it calls `self.history_obj = self.fv.gpmon.getPlugin('ChangeHistory')` and reports failure as error to log.
2. Install this temporary version of Ginga and run it.
3. Start `ChangeHistory` global plugin from menu.
4. Start `SaveImage` global plugin from menu.

When I inserted `print(self.fv.gpmon.getNames())` in `SaveImage.__init__()`, I only see the following listed although I see `ChangeHistory` plugin active and working:

```python
['info', 'colorbar', 'contents', 'cursor', 'zoom', 'toolbar', 'header', 'thumbs', 'operations',
'fbrowser', 'wbrowser', 'pan']
```

This does not make sense, because `SaveImage` did not have this problem before (the last time I tested it maybe 1-2 months ago).

pllim:
@ejeschke , I see that you recently modified how things work behind the scenes. Do you see how that would affect `ChangeHistory` not showing up on `gpmon` but somehow still works?

ejeschke:
@pllim, I do not get this error.  It should be able to find the plugin even if you haven't started it.  Is this in stock Ginga reference viewer or `stginga`?

ejeschke:
@pllim, no major changes "behind the scenes" lately. Just a few tweaks to `freepan` handling and a few changes to pg widgets support.  Certainly nothing that should affect the loading of a plugin module.

pllim:
@ejeschke , I think it is just the way plugins interact with each other has changed now that `SaveImage` and `ChangeHistory` are both in Ginga and not loaded by default. I'll figure out a fix. Thanks!

pllim:
Fixed in e7f051b

PR #331, opened 2016-04-29
Clean up SaveImage

pllim:
This is the follow-up PR for #325 as promised.

~~**However**, it is held up by #330, which prevents me from fully testing the plugin features. I'll keep the "DO NOT MERGE" tag on the title until this issue is resolved.~~


pllim:
@ejeschke , this is ready for final review.

I am not sure why Sphinx test on Travis CI failed, as I don't have problem building the doc locally; Perhaps it is non-issue with the other changes in #313?

ginga/misc/plugins/SaveImage.py
```diff
@@ -169,14 +172,14 @@ def redo(self, *args):
 
         # Only list modified images for saving. Scanning Datasrc is enough.
         if mod_only:
-            func = self.keys_from_datasrc
+            all_keys = channel.datasrc.sortedkeys
```
ejeschke: Ok for now, but it would be good to use an API on the `Datasrc` since its internal implementation could easily change.

ginga/misc/plugins/TVMark.py
```diff
@@ -1,6 +1,5 @@
 """Non-interactive points marking local plugin for Ginga."""
-from __future__ import (absolute_import, division, print_function,
-                        unicode_literals)
+from __future__ import absolute_import, division, print_function
```
ejeschke: :+1: From what documentation I can find, either we should import `unicode_literals` everywhere or not at all.  I know that for some 3rd party python2 libraries it causes problems.  I had to track down an obscure bug with another module where this had been inserted.

I think with an eventual move to python3 only it will not be necessary anyway so I'm in favor of not using it if we can avoid it.

ginga/misc/plugins/SaveImage.py
pllim: @ejeschke , good point. I changed that to use `keys()` method now.

PR #338, opened 2016-05-08
Ginga Startup

wtgee:
* Couldn't get `ginga` to start without following modifications.  
  * Ubuntu 16.04
  * Python 3.5.1 :: Anaconda 4.0.0 (64-bit)

### DIFF
--- a/ginga/gw/GwMain.py
+++ b/ginga/gw/GwMain.py
@@ -14,8 +14,8 @@ import threading
 import logging
 import ginga.util.six as six
 if six.PY2:
-    import thread
-    import Queue
+    import thread as thread
+    import Queue as Queue
 else:
     import _thread as thread
     import queue as Queue


pllim:
Interesting, I had to hack around Python 2/3 compatibility like this for `Task` but my approach was slightly different. See d930a48a96ec.

wtgee:
@pllim Honestly, I'm not sure why my solution works as the `six.PY2` conditional is correct, which means I shouldn't even be loading those first modules, but it works with them and fails without. I'm not sure it solves the real problem.

ejeschke:
That sure seems strange!  And it would seem that importing it that way should result in the same name as if the "as xyz" part was omitted.  Did you try with python3, @wtgee ?

wtgee:
I'm only using python3 is the weird thing. `Python 3.5.1 :: Anaconda custom
(64-bit)`

I'll try with python 2.7

On Mon, May 9, 2016 at 8:21 AM, ejeschke <notifications@github.com> wrote:

> That sure seems strange! And it would seem that importing it that way
> should result in the same name as if the "as xyz" part was omitted. Did you
> try with python3, @wtgee <https://github.com/wtgee> ?
>
> —
> You are receiving this because you were mentioned.
> Reply to this email directly or view it on GitHub
> <https://github.com/ejeschke/ginga/pull/338#issuecomment-217946070>
>



-- 
~Wilfred Tyler Gee

ejeschke:
> I'm only using python3 is the weird thing. Python 3.5.1 :: Anaconda 4.0.0 (64-bit)

Ah, that might be something.  probably "python" resolves to "python3" on Ubuntu 16.04.  But  then it should not be executing the `six.PY2` clause, so something is failing in the detection of which python is running.

wtgee:
On Mon, May 9, 2016 at 8:56 AM, ejeschke <notifications@github.com> wrote:

> I'm only using python3 is the weird thing. Python 3.5.1 :: Anaconda 4.0.0
> (64-bit)
>
> Ah, that might be something. probably "python" resolves to "python3" on
> Ubuntu 16.04. But then it should not be executing the six.PY2 clause, so
> something is failing in the detection of which python is running.
>
That's what is weird is that it is *not* executing the six.PY2 clause. I
put print statements in both and it properly gets to the `else` for
python3. It just doesn't work unless I have the names for the PY2 clauses.
Very strange to me.




> —
> You are receiving this because you were mentioned.
> Reply to this email directly or view it on GitHub
> <https://github.com/ejeschke/ginga/pull/338#issuecomment-217956024>
>



-- 
~Wilfred Tyler Gee

pllim:
I'm curious. What if instead of this fix here, you do:

```python
if six.PY2:
    import thread
    import Queue
else
    import _thread
    import queue

    thread = _thread
    Queue = queue
```

Does this also fix your problem?

wtgee:
@pllim No, that doesn't work either (with edited code)

pllim:
@wtgee , which toolkit are you using? What are the error messages or tracebacks when Ginga refuses to start?

wtgee:
```
  File "/home/wtgee/anaconda3/bin/ginga", line 9, in <module>
    load_entry_point('ginga', 'console_scripts', 'ginga')()
  File "/home/wtgee/Projects/ginga/build/lib.linux-x86_64-3.5/ginga/main.py", line 560, in _main
    reference_viewer(sys.argv)
  File "/home/wtgee/Projects/ginga/build/lib.linux-x86_64-3.5/ginga/main.py", line 555, in reference_viewer
    viewer.main(options, args)
  File "/home/wtgee/Projects/ginga/build/lib.linux-x86_64-3.5/ginga/main.py", line 368, in main
    ev_quit=ev_quit)
  File "/home/wtgee/Projects/ginga/build/lib.linux-x86_64-3.5/ginga/main.py", line 312, in __init__
    GingaView.__init__(self, logger, ev_quit, thread_pool)
  File "/home/wtgee/Projects/ginga/build/lib.linux-x86_64-3.5/ginga/gw/GingaGw.py", line 46, in __init__
    app=self, thread_pool=thread_pool)
  File "/home/wtgee/Projects/ginga/build/lib.linux-x86_64-3.5/ginga/gw/GwMain.py", line 38, in __init__
    queue = queue.Queue()
AttributeError: 'NoneType' object has no attribute 'Queue'
```

pllim:
The code in the traceback does not match what I see in `GwMain.py` from the latest master. How old is your Ginga installation?

pllim:
Looking at 83b9e3270a when this file was added, that line should be
```python
queue = Queue.Queue()
```
not
```python
queue = queue.Queue()
```

wtgee:
On Mon, May 9, 2016 at 9:21 AM, P. L. Lim <notifications@github.com> wrote:

> Looking at 83b9e32
> <https://github.com/ejeschke/ginga/commit/83b9e3270a752085770db58d954d9b8ede7e11c8>
> when this file was added, that line should be
>
> queue = Queue.Queue()
>
>
Odd, this is what I actually see in my editor. The same file I change fixes
the error, it is just the reported error shows a different line.

I might have some old ginga installation that is hiding around somehow and
being used when run but this is a fresh laptop so not likely. I'll dig
around some more.



> not
>
> queue = queue.Queue()
>
> —
> You are receiving this because you were mentioned.
> Reply to this email directly or view it on GitHub
> <https://github.com/ejeschke/ginga/pull/338#issuecomment-217962449>
>



-- 
~Wilfred Tyler Gee

wtgee:
So I eliminated all my `build` directories and pulled down a fresh copy of master and can confirm that the output into my `home/wtgee/Projects/ginga/build/lib.linux-x86_64-3.5/ginga/gw/GwMain.py` file is different depending on my original edits. With the edits line 35 ends up (correctly) as `...Queue.Queue()` but without it ends up as `...queue.Queue`

I am using `python setup.py develop` to build but tried the same thing with `python setup.py install` yesterday.

pllim:
Ah, yes. I see that you are saying now. I think the culprit might actually be `use_2to3=True` in `setup.py`! I might have introduced it during #313. :scream:

pllim:
Please do one more test... Instead of your changes here, just set `use_2to3=False` in `setup.py`. Remove your old build and see if it works.

wtgee:
@pllim :+1: Yup, that did the trick too. I did two clean builds to confirm it failed again and then worked with the `use_2to3=False`. That also was much quicker and seems to make the whole app more responsive than my fix. 

I can close this merge request. Not sure if you need the `2to3` but I can let you fix that elsewhere.  Thanks!

pllim:
Yeah, I'll submit a fix ASAP, since I broke it in the first place...

pllim:
Okay, PR is #342.

wtgee:
Great, thanks for quick response!

On Mon, May 9, 2016 at 9:55 AM, P. L. Lim <notifications@github.com> wrote:

> Okay, PR is #342 <https://github.com/ejeschke/ginga/pull/342>.
>
> —
> You are receiving this because you modified the open/close state.
> Reply to this email directly or view it on GitHub
> <https://github.com/ejeschke/ginga/pull/338#issuecomment-217970963>
>



-- 
~Wilfred Tyler Gee

PR #339, opened 2016-05-09
Add file name to exec

olebole:
This allows to use the desktop file to handle the file types that are actually specified by the mime tye.


ejeschke:
Thank you, @olebole.

PR #342, opened 2016-05-09
Disable 2to3 in setup.py

pllim:
Set `use_2to3=False` in `setup.py`, as discussed in #338.

Thanks for investigating, @wtgee! Sorry for this oversight, @ejeschke !


ejeschke:
No worries.  Glad we caught this quickly.  Yeah, originally ginga did use on-the-fly 2to3 conversion.  But then @cdeil convinced me that `six` was the way to go and it works well.

PR #345, opened 2016-05-20
Rotate scale improvements

ejeschke:
This PR contains a number of small fixes for python 3, Qt 5.  It also cleans up and improves the support for plotting based on WCS coordinates. Additionally new rotate and scale edit control points are provided in the editing gui for shapes.  Some fixes for shared canvases are also included.

### DIFF
--- a/ginga/Control.py
+++ b/ginga/Control.py
@@ -1507,8 +1507,7 @@ class Channel(Callback.Callbacks):
             curimage = self.fitsimage.get_image()
             if curimage != image:
                 self.logger.debug("Setting image...")
-                self.fitsimage.set_image(image,
-                                         raise_initialize_errors=False)
+                self.fitsimage.set_image(image)
 
                 # update cursor to match image
                 imname = image.get('name')
--- a/ginga/ImageView.py
+++ b/ginga/ImageView.py
@@ -615,7 +615,7 @@ class ImageViewBase(Callback.Callbacks):
             Image object.
 
         """
-        if not (self._imgobj is None):
+        if self._imgobj is not None:
             # quick optomization
             return self._imgobj.get_image()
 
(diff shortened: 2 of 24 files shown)


pllim:
This looks like the PR that we discussed yesterday. I'll test it next week. Thanks!

ejeschke:
@pllim, this is the PR I was talking to you about on hangouts. Can you give it a whirl with your plugins? Tnx

ejeschke:
Great thanks!

pllim:
I think this might have fixed #291!

Note that the generated mask does not follow distortion, i.e., you will see slight mismatch between the box drawn in WCS space and the mask generated from it. @ejeschke , you mentioned that you might have a solution for that?

pllim:
@ejeschke , this seems to have broken my [BackgroundSub plugin](http://stginga.readthedocs.io/en/latest/stginga/plugins_manual/backgroundsub.html) in `stginga`. It uses the "old style" drawing, i.e., draw with right-click. Here is the full session log using `--loglevel=10`: 
[ginga_log.txt](https://github.com/ejeschke/ginga/files/279995/ginga_log.txt)

ejeschke:
Hi @pllim, it looks like this is because the `self.coord` value of the object being drawn is `None`. I've added a fix for this and updated the PR.  Can you give it another try?

ejeschke:
And yes, in my tests you can make masks from objects drawn with WCS coordinates.

If you want to make a mask from a shape, accounting for distortion, the best way that I know is using a polygon. So for example, make  a polygon shaped like a circle based on a sampling of wcs points spaced evenly around the origin by radius `r_deg`. Obviously the more points, the better the "shape" of the distorted circle.  But this will be true to the distortion and of course you can make a mask from it.

pllim:
@ejeschke , thanks! I think the PR is good, but maybe with a doc addition, as stated below?

It would be good to note in [Drawing doc](http://ginga.readthedocs.io/en/latest/manual/plugins_local/drawing.html) that polygon is the recommended shape if user wants a more accurate mask drawn in WCS space. It see that the doc is currently empty. Since your work here greatly improves `Drawing`, maybe it is time to write something there?

ejeschke:
Did that fix your plugin then?

Yes, definitely past time to write those missing plugin documentation. However, I think that should be a different PR/update. As this PR fixes a number of bugs, including broken WCS drawing, I don't want to delay it too long.

pllim:
> Did that fix your plugin then?

Yes. Thanks, @ejeschke !

PR #346, opened 2016-06-09
Add new test for ColorDist module

rirze:
This is a first attempt at writing a unit test for the ColorDist module. Any feedback is appreciated. 
- [x]  Implement a test result for the 'histogram equalization scale function'

### DIFF
--- a/doc/quickref.rst
+++ b/doc/quickref.rst
@@ -22,6 +22,7 @@ In a mode, there are usually some special key, cursor and scroll bindings
 which will override *some* of the default ones.
 
 Modes additionally have a *mode type* which can be set to one of:
+
 * `held`: mode is active while the activating key is held down
 * `oneshot`: mode is released by initiating and finishing a cursor drag
 * `locked`: mode is locked until the mode key is pressed again (or `Esc`)
@@ -72,8 +73,9 @@ Panning and Zooming commands
 |                      |   right-drag zooms, middle-click sets pan        |
 |                      |   position                                       | 
 +----------------------+--------------------------------------------------+
-| w                    | Free pan mode: Left-click zooms in, right-click  |
-|                      |   zooms out.  Middle-drag pans freely.           |
+| w                    | Free pan mode: Left-click pans and zooms in,     |
+|                      |   right-click pans and zooms out.  Middle-drag   |
+|                      |   pans freely.                                   |
 +----------------------+--------------------------------------------------+
 | Ctrl + Left drag     | Proportional pan (press and drag left mouse      |
 |                      |     button                                       |
@@ -86,7 +88,7 @@ Panning and Zooming commands
 +----------------------+--------------------------------------------------+
 | doublequote (")      | Toggle autozoom for new images to *on* or *off*  |
 +----------------------+--------------------------------------------------+
-| Ctrl + Scroll wheel  | Adjust zoom by intermediate coarse steps         |
+| Ctrl + Scroll wheel  | Adjust zoom by intermediate fine steps           |
 |   turned             |                                                  | 
 +----------------------+--------------------------------------------------+
 | Shift + Scroll wheel | Adjust pan up/down (and left/right if you have   |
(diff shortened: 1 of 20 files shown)


ginga/tests/test_colordist.py
```diff
@@ -0,0 +1,92 @@
...
+            out = dist.hash_array(data)
+            y = out[0]
+            expected_y = dists[dist_name]
+        
+            try:
+                assert np.array_equal(y, expected_y)
+            except:
+                print y
+                print expected_y
+                print AssertionError("Assertion failed for " + dist_name)
```
nabobalis: These three print statements need () so we can maintain python 3 compatibility and to add: from __future__ import print_function at the top of the code.

ejeschke:
Nice use of a decorator.  Do the tests pass?

pllim:
Travis CI reported test failures for Python 3.

rirze:
A quick question; should testing each distribution be its own test or is it acceptable to have one test go through testing all the distributions. As it is now, there is one function/test that tests every distribution and reports the failed tests. However even if there multiple failed distribution tests, it will only count as one failure. Wondering if this should be the default behavior.

nabobalis:
I would say that it should be one test and Travis will check each distribution. But the code should be written to work under both python 2 and python 3

ejeschke:
@rirze, it would be preferable to have each one be its own test, because reporting is generally done by tests.  So if the "test_asinh" fails then we know that only that distribution might have a problem and that the others are ok.

nabobalis:
Sorry about that, I clearly misinterpreted when you said distribution.

rirze:
Could I get a rebuild on the one check that failed? It seems to have hung up on installing the astropy package and reports a failed check. It doesn't seem to have failed due to my changes.

rirze:
Also, I have problems running `python setup.py test` with ginga. It throws me this output:
```
running test
running build
running build_py
running build_scripts
  File "setup.py", line 116, in <module>
    **package_info
  File "/usr/lib/python2.7/distutils/core.py", line 151, in setup
    dist.run_commands()
  File "/usr/lib/python2.7/distutils/dist.py", line 953, in run_commands
    self.run_command(cmd)
  File "/usr/lib/python2.7/distutils/dist.py", line 972, in run_command
    cmd_obj.run()
  File "/usr/local/lib/python2.7/dist-packages/astropy-1.2.dev14797-py2.7-linux-x86_64.egg/astropy/tests/command.py", line 158, in run
    self._build_temp_install()
  File "/usr/local/lib/python2.7/dist-packages/astropy-1.2.dev14797-py2.7-linux-x86_64.egg/astropy/tests/command.py", line 213, in _build_temp_install
    os.path.basename(self.docs_path))
  File "/usr/lib/python2.7/posixpath.py", line 114, in basename
    i = p.rfind('/') + 1
AttributeError: 'NoneType' object has no attribute 'rfind'
```
I'm currently using `py.test ginga` to run the tests with no problem, but was wondering if there was something on my end I could do to successfully run `python setup.py test` instead. 

I found a issue [here](https://github.com/sunpy/sunpy/issues/1622) with similar output, but am not sure if this is the same problem that I am having.

pllim:
@rirze, with regards to `'rfind'` failure, pretty sure it was fixed in the latest dev, or even 1.2rc1.

pllim:
I think only @ejeschke has admin access to restart the Travis build. In the meantime, you can also hook up Travis CI to your own fork and run the tests there; that way, you have full control on starting/stopping tests for your branch.

rirze:
@pllim Thanks! Pulling the latest Astropy dev fixed the issue. I can run `python setup.py test` without error now.

pllim:
As mentioned during tag-up today, the tests are not running on Travis CI. Maybe this link would help -- http://docs.astropy.org/en/stable/development/testguide.html

pllim:
Thanks for your contribution, @rirze . I have folded in your tests re-written in `pytest` style (and simplified a little) as part of #558. Your PR here has saved me a lot of work figuring out how to test the module, so thanks again!

PR #356, opened 2016-07-06
Typo fix and adding optional keyword

pllim:
Fixed typo in `SaveImage` doc. Make `cfgpath` optional for `generate_cfg_example()`.

@ejeschke , nothing controversial here. I need `cfgpath` to be optional for one of my projects using that function.


ejeschke:
Sounds good!

pllim:
Thanks!

PR #358, opened 2016-07-20
Fixed a bug in canvas.types.astro.Compass.draw

jkunimune15:
When `util.wcsmod.AstropyWCS.radectopix` returns `nan` (caused by some kind of faulty FITS file), `Compass.draw` throws an error that causes the viewer at the top of the Info tab to stop working properly. I'm not sure what caused the error in `radectopix`, but the FITS files I'm using have some bizarre transformations on them from IRAF, so I guess the coordinates are all messed up. I set it to print `"BAD WCS"` like `Ruler` does when it can't resolve coordinates.

### DIFF
--- a/ginga/canvas/types/astro.py
+++ b/ginga/canvas/types/astro.py
@@ -308,9 +308,14 @@ class Compass(OnePointOneRadiusMixin, CanvasObjectBase):
                                   self.cap_radius)
 
     def draw(self, viewer):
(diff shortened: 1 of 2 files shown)


ejeschke:
@jkunimune15, this looks good.  I'll merge it when I have a chance to test it when I get back from vacation in a few days.

ejeschke:
Merging.

PR #359, opened 2016-07-27
Allowed errors to print out from callbacks without loggers

jkunimune15:
Previously, when an error was thrown by a callback from an object without a `logger` attribute, the error remained silent. I set it to print errors to stderr when there is no available `Logger`, instead. That way, errors thrown from callbacks to objects without a `logger`, like `Button`, don't get suppressed.
In doing this, I discovered an error in gw/Desktop.py, where `Desktop` tries to log debug messages, but does not have a `logger`. I fixed that by assigning it a `logger` in its `__init__`.

### DIFF
--- a/ginga/gw/Desktop.py
+++ b/ginga/gw/Desktop.py
@@ -19,6 +19,7 @@ class Desktop(Callback.Callbacks):
         super(Desktop, self).__init__()
 
         self.app = app
+        self.logger = app.logger
         # for tabs
         self.tab = Bunch.caselessDict()
         self.tabcount = 0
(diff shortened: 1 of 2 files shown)


ejeschke:
Looks good, I will merge.  In the future, @jkunimune15, can you please make a branch for your fix and then push your branch to your github repo and generate the PR off of that?  The main reason is so that your `master` remains true to the upstream repo--if you don't, rebasing becomes difficult and you end up making odd merges instead of fast-forwards.  This is standard `git` advice and the default protocol for PRs on github. Tnx.

Issue #357, opened 2016-07-19
No method found matching xxxx


pllim:
I get the following printouts on my terminal when I start Ginga:

```
No method found matching 'gs_swipe'
No method found matching 'ms_draw'
No method found matching 'kp_poly_add'
No method found matching 'kp_edit_del'
No method found matching 'ms_cursor'
No method found matching 'ms_none'
No method found matching 'kp_poly_del'
No method found matching 'ms_wheel'
No method found matching 'gs_tap'
```

It is emitted by `ImageViewBindings.setup_settings_events()` method in `Bindings.py`, in the following logic block:

```python
            # Register for this symbolic event if we have a handler for it
            try:
                cb_method = getattr(self, name)

            except AttributeError:
                self.logger.warning("No method found matching '%s'" % (name))
                cb_method = None
```

And then farther down the codes, I see one of the offending callback name is explicitly commented out:

```python
    ## def ms_draw(self, viewer, event, data_x, data_y):
    ##     return False
```

So, I gather there are no methods defined for the callbacks in the message but they are done on purpose. And there is nothing a user can do about it, so why emit such a warning?

ejeschke:
@pllim, I think at least in some of these cases there is no method in `Bindings` class, but we still want the event to be registered, because some plugins will subscribe for the event later.  I do agree that these messages are annoying and should be eliminated, either by supplying a method that does nothing or by simply eliminating the warning.

I'll take a closer look at it when I get back from vacation in a few days, but if you want to propose something with a PR please feel free.

pllim:
No problem, it's not urgent.

p.s. Does vacation mean you walk 10 feet down to the beautiful Hawaiian beach? :smile:

ejeschke:
> p.s. Does vacation mean you walk 10 feet down to the beautiful Hawaiian beach?
No that's just called "after work"

ejeschke:
@pllim, this should be addressed in #360

ejeschke:
@pllim, is this solved for you by #360?  Can I close it?

pllim:
Yes. Thanks!

PR #360, opened 2016-07-28
Updates to mode handling

ejeschke:
This modifies the mode handling of the basic ginga widget slightly to
fix an issue with overlapping names for triggers for events (e.g.
left arrow key vs. left mouse button).  It also adds the following
capabilities:

- when in "dist", "cmap" or "cuts" mode, you now have the choice of
  scrolling or using the up/down arrow keys, or the b/n keys when
  setting the preference
- intensity map can be set in "cmap" mode using the k/l keys, and
  restored with "i"
- when in "cmap" mode you can toggle a color bar overlay with "c"
- when in "pan" or "freepan" mode, you can use the arrow keys to pan;
  holding down ctrl, shift or both will adjust the amount of the pan
- when in "pan" or "freepan" mode there is a new feature where you can
  save the current zoom setting by pressing "z", and then use the
  backquote key to zoom out to the whole image, and then the "1" key
  to pan to the position under the cursor and restore the zoom level

This PR also fixes two issues:
- as per #357 this suppresses the warning messages for non-existent
    methods in Bindings
- there is an update to ImageViewTk for recent changes to PIL


ejeschke:
@pllim, can you please try this PR with `stginga`?  Although I don't expect any problems unless you are making explicit calls to the 'Bindmap` object.  This adds some nice features to several modes in the basic widget, all of which are documented in the quickref.

This also silences the messages referenced in #357

ejeschke:
Updated with slight modifications to mode lock handling.

pllim:
Thanks, @ejeschke ! I'll take your PRs for a spin on Monday.

pllim:
@ejeschke , I tested this and #361 together and didn't find any problem on my end. Thank you! :+1:

PR #361, opened 2016-07-29
Replace widget native versions of ColorBar

ejeschke:
This removes the widget native (qt/gtk) implementations of the ColorBar
widget and replaces it with a common version based on a ginga widget.
All the functionality is preserved.

Additionally, the Readout class is changed from using a wrapped Label
widget to using a wrapped ginga widget.  This solves a problem with the
Qt label widget limiting/forcing resize of the reference viewer main
screen.


ejeschke:
@pllim, here's another one for you to test, when you get the chance.  Under Qt I find the color bar looks a little better with this one.  And it eliminates a bunch of backend-specific code.  Shouldn't affect anything for `stginga`, I suppose.

pllim:
:+1:  Tested with #360 and saw no problem.

PR #362, opened 2016-08-01
Fix for PIL and Tk backend

ejeschke:
This fixes an issue where the PIL backend doesn't have the correct
surface size configured for the window and a render happens, causing
error messages for the first redraw.  Since the Tk backend frequently
relies on PIL to do the rendering, this fixes the Tk backend in those
cases.

-


ejeschke:
merging

PR #363, opened 2016-08-01
Provide gtk3 backend

ejeschke:
This provides a gtk3 backend to Ginga.  In this commit the basic widget
works well.  Two examples are provided in the examples folder.

The reference viewer starts up.  It is not completely functional yet and may have issues.


ejeschke:
Reference viewer is still somewhat broken with gtk3, but I think it is best to merge this now, as the basic widget is working well.

PR #365, opened 2016-08-05
Fixed gtkw.SpinBox.get_value return type

jkunimune15:
Although it takes a dtype keyword parameter, the `gtk` `SpinBox` widget always returns `float` types. I added a `dtype` attribute to `SpinBox`, and a cast in `SpinBox.get_value` such that it will always return a value of the type that was specified at the object's creation. This makes the class more intuitive to use, and makes its behaviour more similar to `qtw.SpinBox`.


ejeschke:
Makes sense.  We should probably add this to `gtk3w.Widgets` as well.

ejeschke:
Also the callback in `_cb_redirect` should be updated to pass the correct type.

Issue #348, opened 2016-06-10
Make use of QtPy instead of QHelp


astrofrog:
The [QtPy](https://pypi.python.org/pypi/QtPy/) package now provides a common way to deal with the PyQt4/5/PySide differences, and it would make sense for as many projects as possible to use this.

This issue is just to suggest that it might be good for ginga to also migrate to using QtPy?

(I can provide advice if needed, since I'm going through the migration myself for glue)

ejeschke:
@astrofrog, thanks for pointing that out!  I'm very interested. Does it seem to work well and is stable?

ejeschke:
BTW, is it in conda?

astrofrog:
@ejeschke - it is in conda by default I believe, and it seems pretty stable. I've been making a few PRs to port over some of the patches I had in qt-helpers. For the record, here's the PR to change it for glue: https://github.com/glue-viz/glue/pull/1018

Essentially you have to make sure you treat QtPy as if it was PyQt5 (in glue we did the opposite and assumed the PyQt4 API, which is why there was a fair bit to update)

astrofrog:
@ejeschke - just for info, I've now switched over to it in glue: https://github.com/glue-viz/glue/pull/1018#issuecomment-238117149

The latest version can be accessed in the ``conda-forge`` channel in conda:

```
conda install -c conda-forge qtpy
```

But there is also a (slightly older) version available in the default channel.

ejeschke:
Closing this issue, as qtpy is now used via #367

PR #367, opened 2016-08-09
Initial simple changes to use qtpy package for Qt version abstraction

ejeschke:
This PR addresses issue #348.  There is more to be done, including removing code-specific tests for qt5/qt4, but this one change implements most of the work.


ejeschke:
@astrofrog, is it true that you can basically ignore any qt4/qt5 distinctions and write straight to the qt5 api?  I'm wondering also whether `qtpy` automatically will try to use the qt4 v2 api when it is available (I hope so)?

ejeschke:
@pllim, although this is not quite ready for merging, it does remove any direct imports of qt4/qt5/pyside in favor of the `qtpy` module.  So far I don't see any problem for our usual plugins.  You might give this a spin and see if everything is working reasonably for yours.

astrofrog:
> is it true that you can basically ignore any qt4/qt5 distinctions and write straight to the qt5 api?

Yep!

> I'm wondering also whether qtpy automatically will try to use the qt4 v2 api when it is available (I hope so)?

Indeed it will do that:

https://github.com/spyder-ide/qtpy/blob/master/qtpy/__init__.py#L92

pllim:
I get the same error below no matter if I have `qtpy` installed or not:

```
$  ginga --toolkit=qt --loglevel=40 --stderr
  File ".../bin/ginga", line 9, in <module>
    load_entry_point('ginga==2.5.20160805082000', 'console_scripts', 'ginga')()
  File ".../ginga/main.py", line 574, in _main
    reference_viewer(sys.argv)
  File ".../ginga/main.py", line 569, in reference_viewer
    viewer.main(options, args)
  File ".../ginga/main.py", line 308, in main
    from ginga.gw.GingaGw import GingaView
  File ".../ginga/gw/GingaGw.py", line 15, in <module>
    from ginga.gw import GwHelp, GwMain, PluginManager
  File ".../ginga/gw/GwHelp.py", line 8, in <module>
    from ginga.qtw.QtHelp import *
  File ".../ginga/qtw/QtHelp.py", line 65, in <module>
    raise ImportError("Failed to configure qt4, qt5 or pyside. Is the 'qtpy' package installed?")
ImportError: Failed to configure qt4, qt5 or pyside. Is the 'qtpy' package installed?
```

When looking at `QtHelp.py`, I notice that on my machine, `from qtpy.QtCore import QItemSelectionModel` gives me `ImportError`.

astrofrog:
@pllim - I think you are using an old version of QtPy. Can you pip install QtPy 1.1 or later, or install it from conda-forge? Unfortunately in that old version, ``__version__`` was the version of PyQt not QtPy... (but this has changed since).

pllim:
@astrofrog , so it is not on the "default" Anaconda channel? All I did was "conda install qtpy" this morning.

pllim:
So, I think Ginga needs to point out in some way that it requires `qtpy>=1.1` and not just any `qtpy`. I think anyone installing from the default `conda` channel is going to run into this problem.

astrofrog:
@pllim - it's not up to date yet in the defaults channel.

pllim:
I did the following and Ginga appears to work fine for a few GUI operations that I tested. :+1: 

```
$ conda uninstall qtpy
The following packages will be REMOVED:
    qtpy: 1.0.2-py35_0

$ pip install qtpy
Successfully installed qtpy-1.1.2
```

ejeschke:
I think the only remaining question to merging this has to do with installation requirements in the `setup.py`. Currently `ginga` does not specify any special backend as a "requirement", since the user may be building any kind of viewer with the widget.  Requiring `qtpy` would probably force installation of a `Qt` toolkit (?), which may not be what the user wants or expects.

This might simply be handled by an error message that specifies something like: "No available back ends found.  Please install a supported python widget set interface.  If using Qt/PySide, please install package `qtpy`.

pllim:
@ejeschke , specifically it needs `qtpy>=1.1` (which is not yet in default Anaconda distribution).

astrofrog:
> Requiring qtpy would probably force installation of a Qt toolkit (?), which may not be what the user wants or expects.

I don't think that's the case - qtpy doesn't have any dependencies (i.e. one can install it without any qt binding being present, it just won't work at runtime)

astrofrog:
Note that if you install QtPy and PyQt/PySide is not installed, it will tell you. So I think it's ok to put QtPy in the dependencies - as I mentioned above, it doesn't actually have any dependencies as a package.

ejeschke:
I'm planning to merge this when I get back from vacation.  I need to rebase against master and add an installation dependency for `qtpy` in `setup.cfg`.

ejeschke:
Hmm, Travis failure looks unrelated.

ejeschke:
Merging.

PR #369, opened 2016-09-15
New TableViewer plugin

### LINKED ISSUE #368: FITS table viewer
Someone was looking for something that can replace [fv FITS Viewer](http://heasarc.gsfc.nasa.gov/ftools/fv/), which has a table viewer. I'm just wondering if Ginga is the right package for such a thing, or is it out of scope?

I did a test with a sample FITS file that has the following structure (in reality, it can be a lot more extensions with multiple images and tables, and the tables can have many different columns each):
```
Filename: simple_im_tab.fits
No.    Name         Type      Cards   Dimensions   Format
0    PRIMARY     PrimaryHDU       4   ()              
1                ImageHDU         7   (100, 100)   float64   
2                BinTableHDU     12   3R x 2C      [20A, E]   
```

When I use `MultiDim` and select the table extension, I get the following errors from Ginga:
```
2016-09-03 08:33:03,961 | E | MultiDim.py:446 (set_hdu) | Error loading FITS HDU #2:
    too many indices for array
```

pllim:
Fixes #368.

* Added new `TableViewer` global plugin to display FITS table extension.
* Hooked up `MultiDim` to start `TableViewer` when needed.
* Minor bug fix and clean up for `MultiDim`.
    * Removed unused imports.
    * Removed `get_name()` method that is no longer necessary.
    * Fixed a bug of multiple thumbnails created for the same image (e.g., `image[SCI,1]` and `image[1]`).
    * Minor PEP8 clean-ups (I was tempted to clean more but I didn't).
* Added doc.

What this does **not** do:

* Does not include displayed table under `Thumbs` and `Contents`.
* Does not have a way for user to manually remove table from `Datasrc` cache.
* Does not have any callbacks for table.
* Does not handle table that cannot be read by Astropy.
* Does not handle very large table that exceeds available memory.
* Does not have fancy table manipulation (sorting by column is pretty much the only thing you *can* do).


ejeschke:
@pllim, I am seeing the problem of having two thumbnails show up: one with the [primary, 1] designation (this happens when I visit the ascii table in HDU 1 via `MultiDim`) and one with the [0] designation (which is for the main image in the primary HDU).  Any idea what might be going on?  I can probably send an example FITS file if that would be helpful.

ejeschke:
The table viewer looks great.  Can you open more than one table at a time? It seems like it would be nice to be able to have them open in different tabs.

ejeschke:
Thinking about this a little more, it seems like if we were to be consistent with the way channels are managed, if you opened up a table from a file in a certain channel, the viewer in that channel would change to a tableviewer to view the table, and back to an image viewer if you opened up an image.

Wouldn't that seem logical?

pllim:
>  two thumbnails show up: one with the [primary, 1] designation (this happens when I visit the ascii table in HDU 1 via MultiDim) and one with the [0] designation (which is for the main image in the primary HDU)

@ejeschke , I cannot reproduce it. Were you using this PR or `master`? Yes, providing me with the file will help since it is probably specific to how that file is formatted.

pllim:
>  Can you open more than one table at a time?

No. Ideally, if there is a way to switch between tables via `Thumbs` or `Contents`, this won't be a problem. However, to achieve that, I think the proper way is to create a new `AstroTable` class, which is like `AstroImage` but for tables. That is a lot more work and I am not sure if the current need justifies the extra work. What do you think?

> the viewer in that channel would change to a tableviewer to view the table, and back to an image viewer if you opened up an image

Do you mean to simply bring that viewer into focus? Or actually hide/close the other viewer? I prefer the former, but what would be a proper global plugin method to use for that?

ejeschke:
> No. Ideally, if there is a way to switch between tables via Thumbs or Contents, this won't be a problem. However, to achieve that, I think the proper way is to create a new AstroTable class, which is like AstroImage but for tables. That is a lot more work and I am not sure if the current need justifies the extra work. What do you think?

Yes, exactly the idea. And I don't think it is actually a lot more work. We just need `AstroTable` to share some minimal functionality with `AstroImage`.

> Do you mean to simply bring that viewer into focus? Or actually hide/close the other viewer? I prefer the former, but what would be a proper global plugin method to use for that?

Ok, so I borrowed some of your PR and threw together an example in an alternate PR #370.  If you can check this out and give it a whirl you'll get the idea of what I mean.  The thumbnails are currently meaningless black, but that could be improved, maybe.  But the basic navigation within the reference viewer is consistent, without having to open up a special global plugin to view the tables.

pllim:
I ported over my relevant changes from here to naojsoft/ginga#2, so I am closing this one.

PR #370, opened 2016-09-19
Adding a table viewer to Ginga reference viewer

ejeschke:
see github.com issue #368


ginga/table/TableView.py
```diff
@@ -0,0 +1,138 @@
...
+
+    def __init__(self, logger=None, settings=None):
+        super(TableViewGw, self).__init__(logger=logger, settings=settings)
+
+        self.add_callback('table-set', self.set_table_cb)
+
+        from ginga.gw import Widgets
+
+        # Create the viewer as a Treeview widget
+        color_alternate = self.settings.get('color_alternate_rows', True)
```
pllim: It is not immediately clear how user can pass in such a setting to the table viewer. I mean, there is the keyword, but where is the `.cfg` file to use?

ejeschke:
@pllim, can you try out this PR?  This illustrates the idea of keeping the channel organization consistent, and just changing the viewer type within the channel.  I borrowed some of your code from PR #369 to put this together.  It unfortunately doesn't yet have the fix(es?) for the multiple name issue that you fixed in your PR, but if you like this idea we can add the rest of your fixes here.

ginga/table/AstroTable.py
```diff
@@ -0,0 +1,145 @@
...
+
+        try:
+            tbl = Table.read(hdu, format='fits')
+            self._data = tbl
+
+        except Exception as e:
+            self.logger.error("Error reading table from hdu: {0}".format(
+                str(e)))
+
+    def load_file(self, filepath, numhdu=None,
```
pllim: Not sure if you ever plan to support anything other than FITS table (e.g., DAOPHOT, ECSV, general ASCII, etc), but if you do, `load_file()`'s API should be flexible enough to support other formats in the future.

ginga/table/TableView.py
ejeschke: I think this is sharing the basic channel settings (e.g. `channel_Image.cfg`).

pllim:
@ejeschke , I am unable to use this PR to open my test file. It is the same file as given in #368.

```
2016-09-20 10:06:27,571 | E | MultiDim.py:456 (set_hdu) | Error loading FITS HDU #2: too many indices for array
```

pllim:
@ejeschke , I added stuff to this PR via naojsoft/ginga#2

ejeschke:
@pllim, I took the opportunity to do a fair amount of PEP-8 clean up in the plugins, as well as some other minor changes to take advantage of the revamped channel structure that we put in a few months ago for the `ChangeHistory` stuff.  Can you take this for a spin, especially via `stginga` and let me know any issues?

travis is [complaining heavily](https://travis-ci.org/ejeschke/ginga/builds/162074435) about Qt versions.  I didn't change anything regarding this.  Is this related to the recent turbulence with `astropy-helpers`?

ejeschke:
Also, I rebased against master, so the history of the PR has changed.

pllim:
@ejeschke , thanks. I'll have a look when I get the chance, which I might not get until next week.

As for the test failures, it appears Anaconda changed things such that pinning Qt now is the *wrong* thing to do. We just have to keep up by un-pinning Qt. See astropy/astropy#5361 for example.

pllim:
@ejeschke , [this comment on Astropy](https://github.com/astropy/astropy/issues/5360#issuecomment-249045035) also explains the current Travis CI failures.

ejeschke:
Thanks for the CI fix, @pllim.  Looks like Travis is happy again.

ginga/gw/GingaGw.py
```diff
@@ -806,4 +815,11 @@ def page_closed_cb(self, widget, child, wsname):
             self.gui_delete_channel(channel.name)
 
 
+    ########################################################
+    ### NON-PEP8 PREDECESSORS: TO BE DEPRECATED
+
+    getFont = get_font
```
pllim: Maybe use [functions in astropy.utils](http://docs.astropy.org/en/stable/utils/index.html#id3) like `deprecated` or `deprecated_attribute` (where appropriate) throughout Ginga, so it is clear that when the old-style is used, it should be updated?

ginga/table/AstroTable.py
```diff
@@ -23,7 +23,7 @@ class AstroTableHeader(Header):
     pass
 
 
-class AstroTable(Callback.Callbacks):
+class AstroTable(ViewerObjectBase):
```
pllim: :+1:  I thought about doing something like this but I wasn't sure what to name the base class. I am glad you also came to the same conclusion.

pllim:
I also thought about a second header row to display column units (some FITS tables have `TUNITx` keywords to indicate unit strings for the associated columns), but maybe that can be a future enhancement?

pllim:
@ejeschke , `stginga` works as-is with this PR. However, please let me know if you plan to remove deprecated methods so I have time to update `stginga` accordingly, as I am pretty sure I still use the old methods here and there in my plugins.

ginga/table/AstroTable.py
ejeschke: I think `BaseImage` needs to be reserved for things that are fundamentally image like.  But now that we have two kinds of viewers it stands to reason that a "viewable base class" should exist.

ginga/gw/GingaGw.py
ejeschke: Yes, but maybe not immediately.  See comment below.

ejeschke:
> However, please let me know if you plan to remove deprecated methods so I have time to update stginga accordingly, as I am pretty sure I still use the old methods here and there in my plugins.

We have the same situation, so no plans to remove them without ample warning.  At some point in the future I would like to add deprecation warnings, but I want to let a little time pass and clean up a few more classes containing these before we start that.

ejeschke:
@pllim, ok with you to merge?

pllim:
Hold your horses... I am trying to see if I can fit a unit display in as secondary header. Almost done...

pllim:
@ejeschke -- please see naojsoft/ginga#3

PR #371, opened 2016-09-20
Update Travis CI

pllim:
Update Travis CI:

* Pin Qt4 to fix Sphinx build failure.
* Remove Python 3.4 from test (2.7 and 3.5 alone should be sufficient).
* Remove Numpy 1.7 and 1.8 tests, and replaced them with Numpy 1.9 and 1.10 tests with Python 3.
* Now that Astropy 1.2 is released, added stable Astropy back to tests.
* Added conda channels to accommodate for an upcoming `ci-helpers` change.
* Added PEP8 test and allow it to fail. This is useful if you want to clean up the code in the future.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,7 +2,6 @@ language: python
 
 python:
     - 2.7
-    - 3.4
     - 3.5
 
 # Setting sudo to false opts in to Travis-CI container-based builds.
@@ -26,15 +25,17 @@ env:
         # overidden underneath. They are defined here in order to save having
         # to repeat them for all configurations.
         - NUMPY_VERSION=stable
+        - ASTROPY_VERSION=stable
         - SETUP_CMD='test'
         - PIP_DEPENDENCIES=''
+        - MAIN_CMD='python setup.py'
         # *** TODO: We should test the various GUI toolkits that ginga supports
         # on travis-ci ... probably one build for Python 2 / 3 and each toolkit
         # https://ginga.readthedocs.io/en/latest/install.html#dependences
-        - CONDA_DEPENDENCIES='pyqt'
-        # Use dev version until astropy/astropy#4748 is in stable version (1.2)
-        #- ASTROPY_VERSION=stable
-        - ASTROPY_VERSION=development
+        - CONDA_DEPENDENCIES='pyqt qt=4'
+        # Conda channels now need to be explicitly listed, ci-helpers won't
+        # add astropy and astropy-ci-extras by default.
+        - CONDA_CHANNELS='astropy-ci-extras astropy'
 
     matrix:
         # Make sure that egg_info works without dependencies
@@ -43,6 +44,10 @@ env:
         - SETUP_CMD='test'
 
 matrix:
+
+    # Don't wait for allowed failures
+    fast_finish: true
+
     include:
 
         # Do a coverage test in Python 2.
@@ -59,10 +64,12 @@ matrix:
 
         # Try Astropy development version
         # NOTE: Uncomment this if global version is stable, not dev
-        #- python: 2.7
-        #  env: ASTROPY_VERSION=development
-        #- python: 3.5
-        #  env: ASTROPY_VERSION=development
+        - python: 2.7
+          env: ASTROPY_VERSION=development
+        - python: 3.5
+          env: ASTROPY_VERSION=development
+
+        # Try Astropy LTS version (disabled for now)
         #- python: 2.7
         #  env: ASTROPY_VERSION=lts
         #- python: 3.5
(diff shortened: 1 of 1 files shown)


pllim:
@ejeschke , did you disable the Travis CI hook? I don't see the CI kicking off on some recent PRs.

pllim:
Anyway, the build passed in my own branch, see https://travis-ci.org/pllim/ginga/builds/161346137 .

ejeschke:
@pllim, this looks like a really useful set of fixes, thanks!  Question though: are you removing the numpy 1.7 and 1.8 tests just because those versions are "long in the tooth"?  Ginga suggests a minimum version of 1.7 for numpy, but of course I'm happy to raise the required version if there are important efficiency improvements or bug fixes.

pllim:
@ejeschke , removing the versions does not mean they are no longer supported. It simply means they are not tested. I thought comparing 1.9 and 1.10 against both Python 2 and 3 is more useful than comparing 1.7-1.10 for Python 2 only. I can certainly revert that change if you want.

ejeschke:
I see.  numpy does not have python 3 support for any version less than 1.9?  Then that makes a lot of sense.  There are too many combinations to spend a lot of time testing/supporting old versions.  I wonder if I should bump up the requirement to 1.9?

ejeschke:
Merging.

pllim:
@ejeschke , numpy has been supporting Python 3 since 1.5 (according to https://www.scipy.org/scipylib/faq.html#do-numpy-and-scipy-support-python-3-x). It was just my attempt keep the same amount of tests.

pllim:
> I wonder if I should bump up the requirement to 1.9?

IMHO, not until Astropy does, since Ginga is an affiliated package. Otherwise, people will ask why they cannot run Ginga anymore under older Numpy version while Astropy works fine.

ejeschke:
ok, good to know.  There is a bit if a trade off in raising the minimum requirements. On one hand, ginga works perfectly fine with numpy 1.7 and there are some users with "old environments"--implementations they can't upgrade because of other dependencies that are pinned. And then there are the users that are semi-frequently updating to the newer versions.  Sort of want to be able to support both.

I guess I will keep the requirement at >=1.7 for now.

ejeschke:
I just logged in to readthedocs.  It looks like builds are kicking off regularly.  Try forcing a reload in the browser.

pllim:
@ejeschke , Travis CI hook is not the same as RTD hook. When someone submits a PR or a new commit to an existing PR, Travis CI for Ginga is supposed to run the test suite and report success/failure. I don't see that happening anymore.

ejeschke:
Yay, Travis is fixed!

ejeschke:
> When someone submits a PR or a new commit to an existing PR, Travis CI for Ginga is supposed to run the test suite and report success/failure. I don't see that happening anymore.

Hmm, well I am certainly getting reports from Travis.  That's how I knew that the build was failing after merging the upgrade to astropy-helpers.

ejeschke:
And just now it mailed me that the build is fixed.

pllim:
Maybe you disabled Travis CI from running for pull requests? I don't see any Travis CI report for pull requests lately. Used to work.

ejeschke:
I don't think so, unless the defaults have changed with the new Github upgrades.  Do you happen to know where that setting is?

pllim:
Should be `Settings -> Integrations & Services -> Services -> Travis CI -> Edit`

pllim:
Or I think you should be able to access the settings directly at https://travis-ci.org/ejeschke/ginga/settings

pllim:
"Build pushes" and "Build pull requests" should be ON.

ejeschke:
Thanks, they are both on.

ejeschke:
As is were on, and are still on.

pllim:
That is weird. If you look at Ginga PRs that are active, anything after #367 has no "check mark" next to it.

ejeschke:
@pllim, question about the PEP-8 check: as you know, Ginga does include support for Qt, which has a camel case (non PEP8) API.  Is it possible to configure the check to avoid certain files?  I think there may be additional libraries that Ginga uses that may also have this issue.  I don't have any problem fixing up Ginga's methods, although I want to provide the old bindings (clearly marked for future deprecation) for calls that users are very possibly or likely to have included in their plugins or other code.

ejeschke:
Hmm, when I look [here](https://travis-ci.org/ejeschke/ginga/pull_requests) it shows check marks next to other people's builds (if I understand correctly).

pllim:
@ejeschke 

Very odd, why does Travis CI not report the result back in the actual PR?

Also according to http://pep8.readthedocs.io/en/latest/advanced.html, you can ignore certain error codes. But I am not sure about skipping certain files... Maybe @bsipocz knows how?

ejeschke:
> Very odd, why does Travis CI not report the result back in the actual PR?

Maybe that is part of the new changes?  They did make some significant changes to PR review.

pllim:
I don't see such a change over at Astropy. If you look at https://github.com/astropy/astropy/pulls , even the newest PR has the result reported as check marks or crosses.

ejeschke:
The mystery deepens...

pllim:
@ejeschke , when you look at `Settings -> Integrations & Services -> Services -> Travis CI -> Edit`, is the box for "Active" (at the bottom) checked?

ejeschke:
Yes, it shows "Active' (checked) "We will run this service when an event is triggered."

ejeschke:
And I think the page I linked to above shows that indeed the builds are happening for PRs.  The question is why they don't show the status in the PR page.  Maybe there is a setting for that in the PR settings?

pllim:
I am never aware of a specific setting to *disable* reporting of test status on a PR. Perhaps it is time to contact Travis CI support...

pllim:
@ejeschke , I think the problem fixed itself! I can see the test kicking off on PR page for your latest commits into #370. :tada:

ejeschke:
Hmm, but that was me merging, right?  Shouldn't you see something when you post a PR?  I wonder if the build breaks, maybe they disable reporting on the PR page until the build is successful again for a few cycles?

pllim:
@ejeschke not sure. Might just be a temporary glitch from the service provider. We'll see.

Issue #257, opened 2015-12-17
Sorting by numbers with TreeView


pllim:
`TreeView` does not sort numeric column properly because it internally stores the numbers as strings. I tried passing in numbers in `set_tree()` but that didn't work. Also tried padding them with leading spaces or filler characters, also didn't work.

How do I fix this?

![screenshot](https://cloud.githubusercontent.com/assets/2090236/11883963/6fd42eb8-a4e4-11e5-9637-387221d6e1ad.png)

ejeschke:
I discovered that also just a couple days ago.  And I really need numerical sorting.  Currently looking into this as a priority issue as it downgraded one of our applications.  Will keep you posted here.

pllim:
Okay, thank you!

pllim:
@ejeschke , any update on this? Just wondering.

ejeschke:
Not yet.  Are you going to PiA2016?  Might be one of the sprints to fix it.

letyrodridc:
Hi, I'm starting my GSoC PR. Can I take a look to this?

ejeschke:
@letyrodridc, sounds good!

Pointers:
1. You are going to want to fix the `ginga.qtw.Widgets` module (and then the other widgets can follow the same solution--see below)--look for the `TreeView` class
2. The key to solving this for all widget sets is changing the `shadow` type from a `dict` to an `OrderedDict`, and being able to sort the ordered dict, which can be arbitrarily nested.
3. You then change the click events to sort the shadow dict and simply rebuild the table.

[Here](http://stackoverflow.com/questions/4110665/sort-nested-dictionary-by-value-and-remainder-by-another-value-in-python) and [here](http://stackoverflow.com/questions/11753660/sorting-python-dictionary-based-on-nested-dictionary-values) are a couple of links on sorting dictionaries that may give you a start.

I'd suggest starting by solving the problem of sorting an arbitrarily nested ordered dictionary, which is a nice little self-contained problem.

ejeschke:
@pllim, jump in if you have other suggestions!

pllim:
[...]

PR #378, opened 2016-09-27
Move table units to main header

pllim:
Moved table units to main header. Updated screenshot and doc. This is a follow-up of #370. @ejeschke , what do you think? I think this is a cleaner way to display column units.

Side note: Table sorting in general is still affected by #257 as it uses `TreeView` for display. For example, sorting by flux column does not really sort the flux values properly.

New display (also in doc):
multidim-plugin-table

When table has masked data (this is explained in doc as text):
screenshot

Note: The reason masked data is "--" instead of 999999 is reported at astropy/astropy#5363.


pllim:
The second commit replaces the "--" data with 999999 to be consistent with how FITS works:
screenshot-1

ejeschke:
@pllim, I think this is cleaner, and less likely to confuse the header with data.

ejeschke:
Do you want to work on this further, or shall I merge?  The table viewer is your initiative, really.

pllim:
You can merge it. Thanks!

PR #380, opened 2016-09-29
New PlotTable plugin to plot table data

### LINKED ISSUE #379: Plot table data
Now that we have a table viewer (#370), the next logical thing to do is to enable some sort of basic plotting. For example, user starts a plugin for such a plot, select a column for X axis, then another for Y axis, and a plot would show up in the plugin. This would be useful to visualize spectrum data (wavelength vs flux).

This is just an idea. Personally, I just read the table in and plot it with `matplotlib` from within a Python session. But I can see that some users who are not Python-savvy might like such a thing.

If this is out of scope, feel free to close.

pllim:
Added a new `PlotTable` plugin. Also, some minor PEP8 clean up of `main.py`.

Fixes #379.


pllim:
@ejeschke , this is now ready for review. I tested it locally and seems to work.

ejeschke:
Taking a look at this today

ejeschke:
This looks great!  Useful right out of the box for quick look type plots.

ginga/main.py
```diff
@@ -292,10 +297,10 @@ def main(self, options, args):
 
         if toolkit == 'choose':
             try:
-                from ginga.qtw import QtHelp
+                from ginga.qtw import QtHelp  # noqa
```
ejeschke: what is "noqa" ?

ejeschke: Looks good to me.

ejeschke:
Shall I merge, or do you want to add anything else right now?

ginga/main.py
pllim: It's just an incantation to tell the PEP8 checker that I know it's not PEP8 compliant and stop complaining. :)

pllim: Technically it's really flake8 that's complaining. In this case, about unused import.

pllim:
@ejeschke, I am done with this PR. Thanks!

PR #383, opened 2016-10-04
Change HTML theme back to RTD and added a logo on sidebar.

pllim:
As part of #313, the HTML theme was changed to Astropy theme, but I find its sidebar not very useful for how Ginga's doc is nested. So, this PR changes the theme back to RTD.

While I am at it, I shrunk the Ginga splash image and use it as a sidebar logo. Of course, this image can be easily replaced in the future when Ginga has an actual logo.


ejeschke:
I can generate a cleaner version of the icon, possibly, but it's not urgent.

ejeschke:
Merging.

PR #389, opened 2016-10-23
Load FITS table automatically when there is no image

### LINKED ISSUE #387: Cannot load table data outside of MultiDim
When I attempt to load FITS table data directly either using `File -> Load Image` or `FBrowser` by specifying the desired extension directly in the filename, I get the following traceback:

```
2016-10-17 11:04:53,409 | E | Control.py:532 (load_image) |
    Failed to load file 'j2o1506cj_osc.fits': too many indices for array
Error making callback 'activated': local variable 'image' referenced before assignment
Traceback:
  File ".../ginga/misc/Callback.py", line 98, in make_callback

  File ".../ginga/gw/GingaGw.py", line 144, in <lambda>
    item.add_callback('activated', lambda *args: self.gui_load_file())

  File ".../ginga/gw/GingaGw.py", line 555, in gui_load_file
    initialdir=initialdir)

  File ".../ginga/qtw/QtHelp.py", line 223, in popup
    self.cb(path)

  File ".../ginga/Control.py", line 653, in load_file
    image.set(loader=image_loader, image_future=future)
```

Perhaps the solution for this will also indirectly address #386?

pllim:
Fixes #387

Notes:
- The logic can probably be cleaner but maybe that's for the future when refactoring happens in general.
- Not sure if this works on FITSIO. I only use `astropy.io.fits` (formerly PyFITS).

### DIFF
--- a/ginga/Control.py
+++ b/ginga/Control.py
@@ -649,6 +649,7 @@ class GingaControl(Callback.Callbacks):
         except Exception as e:
             errmsg = "Failed to load '%s': %s" % (filepath, str(e))
             self.gui_do(self.show_error, errmsg)
+            return
 
         future = Future.Future()
         future.freeze(image_loader, filepath, **kwdargs)
--- a/ginga/Control.py
+++ b/ginga/Control.py
@@ -5,23 +5,28 @@
 # Please see the file LICENSE.txt for details.
 #
 # stdlib imports
-import sys, os
+import sys
+import os
 import traceback
-import re, time
+import time
 import tempfile
-import ginga.util.six as six
-if six.PY2:
-    import thread
-    import Queue
-else:
-    import _thread as thread
-    import queue as Queue
 import threading
 import logging
 import mimetypes
 from collections import deque
-import atexit, shutil
+import atexit
+import shutil
 from datetime import datetime
+
+# Local application imports
+from ginga import cmap, imap
+from ginga import AstroImage, RGBImage, BaseImage
+from ginga.misc import Bunch, Datasrc, Callback, Timer, Future
+from ginga.util import catalog, iohelper
+from ginga.canvas.CanvasObject import drawCatalog
+
+# Version
+from ginga import __version__
 
 magic_tester = None
 try:
@@ -37,17 +42,7 @@ try:
 except (ImportError, Exception):
     have_magic = False
 
-# Local application imports
-from ginga import cmap, imap
-from ginga import AstroImage, RGBImage, BaseImage
-from ginga.misc import Bunch, Datasrc, Callback, Timer, Future
-from ginga.util import catalog, iohelper
-from ginga.canvas.CanvasObject import drawCatalog
-
-# Version
-from ginga import __version__
-
-#pluginconfpfx = 'plugins'
+# pluginconfpfx = 'plugins'
 pluginconfpfx = None
 
 packageHome = os.path.split(sys.modules[__name__].__file__)[0]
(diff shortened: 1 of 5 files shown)


ejeschke:
I'll take a look at my first opportunity.

pllim:
@ejeschke , I got this to work. Please review when you get the chance. Also, Travis failure is caused by astropy/astropy#5437

ginga/Control.py
```diff
@@ -517,29 +514,37 @@ def load_image(self, filepath, idx=None):
...
             image.load_file(filepath, **kwdargs)
 
-        except Exception as e:
-            errmsg = "Failed to load file '%s': %s" % (
-                filepath, str(e))
-            self.logger.error(errmsg)
+        except Exception:
+            # TODO: Refactor code to be less hacky.
+            # Maybe it has a table on EXT 1.
             try:
```
ejeschke: My suggestion would be: once we have determined the file is a FITS file (or we have determined that we don't know what it is, so we are assuming it is FITS), we ask `ginga.util.io_fits` to open it and magically return us the right kind of thing (`AstroImage` or `AstroTable`).  We can assume that module has more detailed info than we have here; for example, once it opens the file and takes a look at the HDU it will know what kind of thing to create.

ginga/Control.py
```diff
@@ -576,8 +581,8 @@ def  _dl_indicator(count, blksize, totalsize):
...
                 dl = catalog.URLServer(self.logger, "downloader", "dl",
                                        info.url, "")
-                filepath = dl.retrieve(info.url, filepath=info.filepath,
-                                       cb_fn=_dl_indicator)
+                dl.retrieve(info.url, filepath=info.filepath,
+                            cb_fn=_dl_indicator)
```
ejeschke: I think we need `filepath` below.  So we shouldn't get rid of the assignment here.

ginga/Control.py
```diff
@@ -644,6 +649,7 @@ def load_file(self, filepath, chname=None, wait=True,
         except Exception as e:
             errmsg = "Failed to load '%s': %s" % (filepath, str(e))
             self.gui_do(self.show_error, errmsg)
+            return
 
         future = Future.Future()
         future.freeze(image_loader, filepath, **kwdargs)
```
ejeschke: Here's where we need `filepath`

ginga/Control.py
ejeschke: Probably we should add a new method to the fits loader classes in `io_fits` that looks a bit like `load_file`, but instead of returning a tuple with the data (and filling in a passed header) it creates the right kind of `AstroImage` or `AstroTable` and then calls the right methods to load it and return it.

pllim: I have two concerns here:
* I am not familiar with FITSIO, so I can only fulfil this implementation for `astropy.io.fits`.
* Circular imports.

ginga/Control.py
pllim: My `flake8` checker is complaining that `filepath` is defined by unused, that's why I deleted it. I don't see how it can be used in Line 649 (655 in this PR). See my next reply below.

ginga/Control.py
```diff
@@ -1466,8 +1477,8 @@ def get_loaded_image(self, imname):
     def add_image(self, image, silent=False, bulk_add=False):
 
         imname = image.get('name', None)
-        assert imname is not None, \
-               ValueError("image has no name")
+        if imname is None:
+            raise ValueError("image has no name")
```
pllim: I think `assert` is fine if you want `AssertionError`. But here, you are reraising the error as `ValueError`, so it is quite confusing to read. But I guess it comes down to personal preference. Do you want me to change it back?

pllim:
Superseded by #397.

PR #390, opened 2016-11-03
Refactoring io_fits

ejeschke:
- this is an alternate implementation of PR389 (Load FITS table
  automatically when there is no image)
- Refactors io_fits so that the decision about what kind of object to
  make (AstroImage or AstroTable) is made in the context of the FITS
  loader, which has the information at hand about the HDUs and their
  types
- Still allows AstroImage and AstroTable objects to be created and use
  load_file() or load_hdu(), but code is reduced overall by centralizing
  the handling somewhat

- still need a little more refactoring to provide support for fitsio
  loader, but this shouldn't be difficult

### DIFF
--- a/ginga/AstroImage.py
+++ b/ginga/AstroImage.py
@@ -11,7 +11,7 @@ import logging
 import time
 import traceback
 
-import numpy, numpy.ma
+import numpy
 
 from ginga.util import wcsmod, io_fits, iohelper
 from ginga.util import wcs, iqcalc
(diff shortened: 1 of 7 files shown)


ejeschke:
@pllim, I took a few of your edits and created this alternate PR for #389.  Can you take a look and comment and try it out?  I think this explains a little more clearly what was trying to articulate in the comments on #389.  

If you agree this is a reasonable alternate, we can merge some more of your PEP-8 clean up from the other PR.  Plus I want to add support for `fitsio` yet, which should be pretty simple.

pllim:
@ejeschke , thanks for doing this! Maybe `MultiDim` can benefit from this as well. It also uses some logic for table HDU.

pllim:
I will try this out tomorrow. Thanks again!

ejeschke:
Yes, I think `MultiDim` is definitely due for some refactoring, and probably should tie into these changes.

pllim:
Seems to work for my test data. :+1:

ejeschke:
@pllim, image names seem ok?  That is the main thing I am wondering about.

pllim:
I opened one with mixed image and table extension. It opened up the image extension first with no error.

pllim:
I guess this can be closed now that it is superseded by #397.

ejeschke:
This PR was subsumed and merged in another PR

PR #393, opened 2016-11-09
Quick fix for scroll crash

stscieisenhamer:
The debug message was trying to format None as a float.
Made formatting generic.


ejeschke:
Thank you, @stscieisenhamer!  That was quick!

PR #401, opened 2016-11-14
MultiDim to use new FITS I/O interface

### LINKED ISSUE #399: Update MultiDim to use new I/O interface
This is a follow-up of #397, since now `MultiDim` can use some clean-up to take advantage of the cleaner FITS I/O behind the scene.

pllim:
Fixes #399 . Also see #397. Also some PEP8 clean-ups on `MultiDim` and `AstroImage`.

### DIFF
--- a/ginga/AstroImage.py
+++ b/ginga/AstroImage.py
@@ -6,7 +6,6 @@
 #
 import sys
 import math
-#import time
 import traceback
 
 import numpy
@@ -640,7 +639,6 @@ class AstroImage(BaseImage):
         ra_lbl, dec_lbl = six.unichr(945), six.unichr(948)
 
         # Calculate WCS coords, if available
-        #ts = time.time()
         try:
             if self.wcs is None:
                 self.logger.debug("No WCS for this image")
@@ -652,7 +650,6 @@ class AstroImage(BaseImage):
 
             elif self.wcs.coordsys == 'pixel':
                 args = [data_x, data_y] + self.revnaxis
-                # args = (data_x, data_y)
                 x, y = self.wcs.pixtosystem(args, system=system, coords='data')
                 ra_txt = "%+.3f" % (x)
                 dec_txt = "%+.3f" % (y)
@@ -661,7 +658,6 @@ class AstroImage(BaseImage):
             else:
                 args = [data_x, data_y] + self.revnaxis
 
-                # args = (data_x, data_y)
                 lon_deg, lat_deg = self.wcs.pixtosystem(
                     args, system=system, coords='data')
 
@@ -707,7 +703,6 @@ class AstroImage(BaseImage):
                 tb_str = "Traceback information unavailable."
                 self.logger.error(tb_str)
 
-        #te = time.time() - ts
         info = Bunch.Bunch(itype='astro', data_x=data_x, data_y=data_y,
                            x=data_x, y=data_y,
                            ra_txt=ra_txt, dec_txt=dec_txt,
--- a/ginga/AstroImage.py
+++ b/ginga/AstroImage.py
@@ -420,7 +420,6 @@ class AstroImage(BaseImage):
         header = self.get_header()
         ((xrot_ref, yrot_ref),
          (cdelt1_ref, cdelt2_ref)) = wcs.get_xy_rotation_and_scale(header)
-        #ref_rot = yrot_ref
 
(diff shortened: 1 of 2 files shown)


ginga/misc/plugins/MultiDim.py
```diff
@@ -316,38 +318,8 @@ def close(self):
...
-        if not auto_start:
-            return False
-
-        # Start ourselves if GUI is not up yet
-        if not self.gui_up:
-            self.fv.start_local_plugin(self.chname, str(self), None)
-
-        return True
-
-
```
ejeschke: @pllim, why remove `redo()`?  It means that `MultiDim` does not reinitialize properly if a new image is set in the channel.  It will display incorrect information, and it will contain the previous image's open `fits_f`.

ginga/misc/plugins/MultiDim.py
```diff
@@ -399,24 +371,14 @@ def set_hdu(self, idx):
...
-        else:
-            # inherit from primary header?
-            inherit_prihdr = self.fv.settings.get('inherit_primary_header',
-                                                  False)
-
-            image = AstroImage.AstroImage(logger=self.logger,
-                                          inherit_primary_header=inherit_prihdr)
-
-        self.image = image
```
ejeschke: I like how this collapses nicely and the plugin no longer has to figure out what kind of object to create.

ginga/AstroImage.py
```diff
@@ -643,7 +640,7 @@ def info_xy(self, data_x, data_y, settings):
         ra_lbl, dec_lbl = six.unichr(945), six.unichr(948)
 
         # Calculate WCS coords, if available
-        ts = time.time()
+        #ts = time.time()
```
ejeschke: we can probably remove these timing calls.  They were there to measure the slowness of astropy coordinate conversions.  But now that problem is handled elsewhere by just not calling this function as frequently.

ginga/AstroImage.py
```diff
@@ -655,22 +652,23 @@ def info_xy(self, data_x, data_y, settings):
 
             elif self.wcs.coordsys == 'pixel':
                 args = [data_x, data_y] + self.revnaxis
-                x, y = self.wcs.pixtosystem(#(data_x, data_y),
-                    args, system=system, coords='data')
+                # args = (data_x, data_y)
```
ejeschke: I think this commented line can be removed as well.

ginga/misc/plugins/MultiDim.py
pllim: The `redo()` that I removed is a **duplicate**. In fact, my Emacs was complaining that it had be superseded by the second `redo()` method farther down.

ginga/AstroImage.py
pllim: Done.

ginga/AstroImage.py
pllim: Done.

ginga/misc/plugins/MultiDim.py
ejeschke: Good grief.  Ok, well that's shrinks the code even more and that's good too.

pllim:
@ejeschke , so is this good to merge now?

ginga/AstroImage.py
```diff
@@ -425,7 +420,7 @@ def mosaic_inline(self, imagelist, bg_ref=None, trim_px=None,
         header = self.get_header()
         ((xrot_ref, yrot_ref),
          (cdelt1_ref, cdelt2_ref)) = wcs.get_xy_rotation_and_scale(header)
-        ref_rot = yrot_ref
+        #ref_rot = yrot_ref
```
ejeschke: I think we can remove this comment if not used in any commented code below

ejeschke:
@pllim, you can merge it if you've tested thoroughly.  Otherwise I'll test it and merge a little later today.

ginga/AstroImage.py
pllim: I couldn't find any usage of `ref_rot`. It is deleted.

pllim:
@ejeschke , extra tests wouldn't hurt. I don't have Subaru data to test with. So I'll wait for you to merge then. There is no real rush on my side. Thanks!

ejeschke:
Merged.

pllim:
Thanks!

PR #407, opened 2016-11-23
Refactor reference viewer

ejeschke:
This plugin reorganizes and consolidates the reference viewer code under `ginga.rv`.

This will aid in documenting and understanding how the reference viewer is put together and separates it a bit from the rest of the ginga toolkit.  As part of this, I have also merged two classes that used to be combined at run-time to produce the reference viewer shell.  For a long time now this has not been necessary and so it makes sense to wrap that all up together in one class for clarity.  I've also moved the `Channel` class into it's own file for the same reason.

The plugins are reorganized into one subfolder `ginga.rv.plugins`. This mimics the design of custom reference viewers that do not subclass ginga's reference viewer.

This reorganization should not affect any users except those creating a custom reference viewer by subclassing ginga's, or are subclassing an existing plugin.  For these users a simple path change importing the module is all that is needed.


ejeschke:
@pllim, take a look at this PR when you get a chance.  I've filed a PR for `stginga` that fixes the appropriate imports if we merge it. Thanks!

pllim:
Thanks, I'll review after Thanksgiving holidays. Happy Thanksgiving!

ejeschke:
Sounds good.

pllim:
@ejeschke , this PR is reporting conflict. Could you please resolve the conflict first, so I can be sure that I am actually testing the version that is going to be merged? Thanks!

ejeschke:
@pllim, rebased against master.

pllim:
@ejeschke , I been testing it (or so I thought) and then I realized I might had been just using an older version by accident. Because without naojsoft/ginga#4, I would not be able to start Ginga in the first place. I did find some issues but I will open them as separate ones since I don't think they are related to this PR.

pllim:
:+1: to merge from me. Thanks!

ejeschke:
Great, Thanks!

pllim:
@ejeschke , are you going to make a release of this refactoring soon?

ejeschke:
Yes.  But I want to merge a couple of other smaller mods first.  And if #411 is actually an issue we should fix it before making a release.

PR #408, opened 2016-11-23
Increase play timer internal to avoid viewer crashing

### LINKED ISSUE #406: MultiDim: Play button broken for playing through image cubes
The play button in the `MultiDim` plugin appears to be broken for playing through image cubes.  All the other controls seem to work: First, Last, Next, Prev, etc.

The symptom is either slow playing or no playing, followed by a viewer freeze.

pllim:
Increase play timer internal to avoid viewer crashing. Fixes #406. At least on my machine, it works when I increase the interval. With the old value, I could reproduce the crashing.

### DIFF
--- a/ginga/misc/plugins/MultiDim.py
+++ b/ginga/misc/plugins/MultiDim.py
@@ -52,7 +52,7 @@ class MultiDim(GingaPlugin.LocalPlugin):
         self.play_axis = 2
         self.play_idx = 1
         self.play_max = 1
-        self.play_int_sec = 0.1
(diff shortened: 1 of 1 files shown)


pllim:
Note: Looks like the Play button was added in b6640b5d73dd483b and has not changed since then. Pretty sure it once worked, so I cannot explain why it does not anymore and needs this fix.

ejeschke:
@pllim, I think I have a fix that allows very fast playback.  Should have it merged shortly.

Issue #310, opened 2016-03-25
Operations menu to handle a lot of local plugins


pllim:
The number of plugins are getting to the point that current `Operations` dialog is not going to be sufficient. Some ideas below:
- Some kind of scrolling mechanism
- Separating them into categories -- But what if a plugin can fall into multiple ones? How do users instinctively know which category to look for a certain plugin?
- Multi-column menu
- Anything else?

ejeschke:
@pllim this issue is partially addressed in commit 3364b3206dc8f64a945b3567d8062ce3d39fe0dd.

We can now specify categories for plugins and it will make nested menus under those categories.  I'm not suggesting that as a full solution to the issue, but it's a capability we could use.  See [main](https://github.com/ejeschke/ginga/blob/master/ginga/rv/main.py) spec entries for plugins for examples of how the categories are set.  If no category is specified for a local plugin it gets put in at top-level; if none is specified for a global it goes in under "Global".

If you put a "." in the category, it creates a sub-category.  When you get a chance, play with it a bit and see if that seems like something we could use to wrangle some order into the increasingly large list of plugins.

PR #426, opened 2017-01-14
Changes to allow separately distributed plugins

ejeschke:
This is a refresh on issue #207.  It uses the entry points method as suggested in the comments there to provide the linkage between a separately distributed and installed plugin and Ginga.

Note that this does not change all the existing ways that a plugin can be registered (e.g. from the command line, or the user's ginga configuration, etc.  It just adds an additional way.

The template for writing such an externally distributed plugin [can be found here](https://github.com/ejeschke/ginga-plugin-template).


ejeschke:
@pllim, have a look at this when you get a chance.

pllim:
Would be useful to add some user-facing documentation about this under "developer" section and mention the template.

Personally, I don't like having hyphen in directory names. Why can't "reference-viewer" simply be renamed "rv" for consistency with the rest of the code base?

So, I implemented a proof-of-concept at https://github.com/pllim/ginga_plugin_exp and installed it. When I run `python util/test_registration.py`, I got:

```
{'klass': 'BackgroundSub', 'path': '.../lib/python3.5/site-packages/ginga_plugin_exp-0.1.dev0-py3.5.egg/plugins/BackgroundSub.py', 'module': 'BackgroundSub', 'workspace': 'dialogs'}
{'path': '.../lib/python3.5/site-packages/ginga_plugin_exp-0.1.dev0-py3.5.egg/plugins/MosaicAuto.py', 'module': 'MosaicAuto', 'tab': 'MosaicAuto', 'workspace': 'right', 'klass': 'MosaicAuto', 'start': False}
```

But when I start Ginga (2.6.2.dev1326), I get no error message but I don't see my plugins either. I used `ginga --loglevel=40 --stderr` command to start Ginga. Am I doing something wrong here?

ejeschke:
> But when I start Ginga (2.6.2.dev1326), I get no error message but I don't see my plugins either. I used ginga --loglevel=40 --stderr command to start Ginga. Am I doing something wrong here?

First, are you sure you were running the version of ginga in this PR (it was installed etc)?

I believe you should see an error message about not being able to load the plugin if there was an issue with it.  I suppose `--loglevel=40` should show it. I usually use `--loglevel=20 --stderr` and then just scroll back to the beginning of the output after the command line.  If there was no problem then there should be an INFO  level line at least saying that the plugin was loaded.

ejeschke:
Could you try to just install the template as-is?  If we see those plugins then we should be able to debug why your custom ones don't show up.

ejeschke:
> Personally, I don't like having hyphen in directory names. Why can't "reference-viewer" simply be renamed "rv" for consistency with the rest of the code base?

Do you mean the name under the `examples` folder?  I guess I'd be ok with that, but all the other names under there are for toolkits.  I thought "rv" under there might seem like a toolkit name.  But I don't have a strong opinion about it.

pllim:
> are you sure you were running the version of ginga in this PR

Ops!!! Rookie mistake. Will try again.

> Do you mean the name under the examples folder? 

Yes.

pllim:
So, by actually using Ginga from this PR, my local plugin worked but not my global one:

```
2017-01-25 16:32:08,671 | E | PluginManager.py:70 (load_plugin) | Failed to load plugin 'MosaicAuto': __init__() missing 1 required positional argument: 'fitsimage'
2017-01-25 16:32:08,672 | E | PluginManager.py:74 (load_plugin) | Traceback:
  File ".../python3.5/site-packages/ginga-2.6.2.dev1319-py3.5.egg/ginga/gw/PluginManager.py", line 49, in load_plugin
    obj = klass(self.fv)
```

This is perplexing because this same global plugin works in `stginga`. Also, I see that the custom plugins added this way are added to the *top* of the menu, not bottom, which is inconsistent with how other custom plugins are listed.

pllim:
Oh, my bad. Mosaic is actually a local plugin that I mistook for global. You can ignore my error reported above. I actually don't have any global custom plugin from `stginga` that I can use. Still, my comment about listing order stays.

pllim:
Update: I copied over a global plugin from your template to my proof-of-concept repo. It worked!

So, I think this is good to merge **after** we resolve the list ordering issue.

ejeschke:
> So, I think this is good to merge after we resolve the list ordering issue.

@pllim, do you feel that plugins registered this way should show up at the bottom of the list instead?

I am thinking that (as per #310), the menu structure needs to be addressed anyway, but in a separate PR.

ejeschke:
Also, I'll add some documentation before merging.

ejeschke:
@pllim, I changed the order in which the separately distributable plugins are added--they now get added *after* the others.  Is this preferable?

pllim:
> Is this preferable?

To me, yes. That is the order I am used to anyway. If you wish to reverse the order, it needs to be done consistently; That is, a custom plugin added in another way (e.g., via `stginga`) would also be ordered the same.

ejeschke:
@pllim, so are you ok to merge this?  I'm going to add a little documentation in the manual under the section on writing plugins before merging.

pllim:
@ejeschke , yes, please. I also think we can close #207, as it has been superseded. Thank you!

pllim:
Also close #230.

ejeschke:
I think there is one item to be addressed, which is the recommended way to specify modifications to the workspace hierarchy; e.g. if a plugin wants to add a new workspace.  But I think we'll address that in a separate issue.

PR #433, opened 2017-02-02
Resized screenshots and corrected typos in doc

pllim:
Resized screenshots and corrected typos in newly added doc.

Out of scope of this PR, but I think the screenshots should also include the image display area. For instance, it is more useful to see what the crosshair looks like in addition to the plugin UI. If you decide to do that, remember to increase the `:width:` value in the Sphinx image directive (800px should be sufficient).

p.s. Setting the width ensures that the screenshot does not appear too large. However, taking a smaller screenshot can also decrease the overall repo size.


ejeschke:
Thanks for this, @pllim!

Issue #438, opened 2017-03-09
Feature Request: Add option to pop out Help text into a separate window


pllim:
This is how Help currently displays for me, which requires me to either scroll back-and-forth non-stop or adjust the right panel (making image display area smaller), none of which is ideal for me.

![screenshot](https://cloud.githubusercontent.com/assets/2090236/23760801/c695bcea-04be-11e7-94cd-162c18b5df23.png)

While I understand that you prefer not to have too many windows open, I also would still like an option for separate window for Help text. However, as a compromise, we can have this feature purely optional via user clicking a button that clearly states the Help text would create a new window. What do you think?

ejeschke:
Hi @pllim,
I don't have any real problem with a pop out.  Ideally that could be a preference that you set in `general.cfg`.

I think the real issue though is that the text does not wrap nicely.  I played around a little bit with the docutils functions for generating HTML.  I'm wondering whether could have a solution that checked whether the webkit widget was installed (used by `WBrowser` plugin, currently) and also `docutils`, and if they are present it would render the class docstring as a web page in the plugin.  This would make the text flow smoothly and you'd only have to scroll vertically.  What do you think?

pllim:
@ejeschke , as long as I don't have to scroll left-right continuously, I am open to the ideal of HTML rendering.

ejeschke:
@pllim, I've updated the `TVMark`, `TVMask`, `ChangeHistory` and `SaveImage` to use the new help mechanism.  I've pretty much just inserted the ReST markup from the manual pages, and tweaked it a bit so that it renders without references to the rest of the manual (which give a nasty error message, since they are not available--things like cross-references, etc.).

ejeschke:
We might want to add `docutils` as a recommended package to install (or maybe even required?).

Also think about making it so that the images can be rendered into the help.  As we discussed offline, this might entail moving the documentation so that it installs with the package.  But I'm not sure that is the best way to proceed.

pllim:
Re: docutils

Technically, Ginga can still function without docutils. So, you can certainly list it as optional dependency. I see that you already have `have_docutils` in `Control.py`, so all you need to do is add a logic to handle `have_docutils = False` and display some generic error message when `Help` is pressed.

Re: package data

When the full rendered doc is somehow available as package data, can `Help` then just re-direct to `WBrowser`?

ejeschke:
> so all you need to do is add a logic to handle have_docutils = False and display some generic error message when Help is pressed.

[...]

PR #440, opened 2017-03-21
Improve plugin help by providing on-the-fly HTML conversion

ejeschke:
This PR adds support for converting docstrings to HTML on the fly so that the WBrowser plugin can display a more nicely formatted help text.  In this initial PR, only "Cuts" plugin is configured to show the help text in this way.
 
To use the plugin, the "docutils" package must be installed and the WBrowser plugin must be loaded.

### DIFF
--- a/ginga/rv/Control.py
+++ b/ginga/rv/Control.py
@@ -23,7 +23,7 @@ from ginga import cmap, imap
 from ginga import AstroImage, RGBImage, BaseImage
 from ginga.table import AstroTable
 from ginga.misc import Bunch, Timer, Future
-from ginga.util import catalog, iohelper, io_fits
+from ginga.util import catalog, iohelper, io_fits, toolbox
 from ginga.canvas.CanvasObject import drawCatalog
 from ginga.canvas.types.layer import DrawingCanvas
 from ginga.util.six.moves import map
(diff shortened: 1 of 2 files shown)


ejeschke:
@pllim, can you give this a look when you have a chance?  This is to address #438.

Right now only the "Cuts" plugin will show the text using the new feature, but if this looks like a reasonable approach we can convert the other plugins to use it.

pllim:
Looks good to me! In a way, I can get a separate window out via MDI (although mousing over the image window would automatically push the help window into the back but it is not big deal for me).

screenshot

I wonder how this would behave if I have a help window open for "Cuts", and then proceed to click Help on "Histogram" (if they both use `docutils`). I currently cannot test that because only "Cuts" have this new feature.

Thank you!

ejeschke:
> I wonder how this would behave if I have a help window open for "Cuts", and then proceed to click Help on "Histogram" (if they both use docutils).

With the current implementation it would replace the documentation in the WBrowser window with that from "Histogram" (this has nothing to do with docutils, but rather with the fact that WBrowser plugin is presently a global plugin and therefore a singleton).  It seems like it might be nice to have the option of opening another tab/subwindow.  What do you think?  Is it too many windows?

pllim:
> it would replace the documentation in the WBrowser window

For the same channel, that is fine. But if I have "Cuts" open in one channel but "Histogram" open in another, I am not sure, because the plugins will auto-focus themselves when you switch channels, but the Help page won't be. Something to think about...

ejeschke:
I think it comes down to how many help pages we want to support being open at the same time.  There is some simplicity in having just one.  Maybe we can leave that for a future enhancement if someone really wants it...?

pllim:
> Maybe we can leave that for a future enhancement

Sure! :smile:

ejeschke:
Merged.

PR #448, opened 2017-04-18
Fix section definition of Pick documentation

pllim:
Currently, the subsections of `Pick` are also listed under "local plugins" listing. This PR fixes the subsections definitions.


ejeschke:
Thanks, @pllim!

PR #449, opened 2017-04-18
Categories for local plugins

### LINKED ISSUE #310: Operations menu to handle a lot of local plugins

pllim:
Fix #310 

@ejeschke , I added the categories with names as you suggested. What do you think? I think I like this approach, as this allows `stginga` to insert its own plugins under a special category, if so desired.

untitled


ejeschke:
I agree, because I think we would do something similar (maybe by instrument name) here at Subaru. 

I'd suggest that you/we go ahead and categorize the global plugins that don't have `hidden=True` as well. This includes (at present) `WCSMatch`, `IRAF`, `SAMP`,  `RC`, `WBrowser`, `ColorMapPicker`, `SaveImage`, `ChangeHistory`, `Log`, `Command` as well as the global entries for the chameleon plugins `FBrowser` and `Blink`.

pllim:
> categorize the global plugins

@ejeschke , I thought about that but decided against it. There are currently not *too* many of those, so adding extra layer to get to them might hinder more than help. Maybe wait until there is more before we categorize them? What do you think?

ejeschke:
I think we may as well go ahead and categorize them. Otherwise they are all lumped under the somewhat unhelpful category of "Global".  As an end user thinking, e.g. "I want to link two images WCS together", that doesn't really help you find it `WCSMatch`. Whereas if it is under something like "WCS" I think you probably could.

pllim:
@ejeschke , I added the global categories as requested. Also, I notice that the category listing order is determined by how plugins are ordered in the lists, so I re-arranged those for a more "intuitive" category listing order. What do you think?

ejeschke:
I wonder if we need to separate all the global plugins under the first-level "Global" menu.  For example, it makes sense to me (thinking as an end-user) that "RC" and "SAMP" would be found under the top-level "Remote" menu along with "Catalogs".

What if the menu entry got an indicator if the plugin was a global; i.e. something like "SAMP (g)" ?

ejeschke:
In general I like the categories you've picked.  I think we have to choose carefully though, because once we've settled on a layout it will confuse users to move things around.

pllim:
> SAMP (g)

That is a good idea, but "(g)" needs to be documented somewhere, as a new user will not likely to figure it out what it means.

> once we've settled on a layout it will confuse users to move things around

Yes. It is quite impossible to pick something to make *everyone* happy but how do we decide that 90% Ginga users will be okay with it? Maybe consult your Subaru users? In addition, people can always customize their reference viewer, right?

pllim:
@ejeschke , now that I look at the code, I am not sure how to add a "(g)" next to the plugin name (not category) in the listing without breaking the code.

ejeschke:
> now that I look at the code, I am not sure how to add a "(g)" next to the plugin name (not category) in the listing without breaking the code.

Let's hold off on that for a bit then.

Do we need to distinguish global from local within the menus?

pllim:
> Do we need to distinguish global from local within the menus?

Since the behavior is different (i.e., affecting all channels or specific ones), I think it is better to be clear about it.

ejeschke:
I think what I'd like to do then is figure out a way to indicate which ones are global and show that in the menu item (color, icon, text, ?)  From a human interface standpoint I think it makes sense not to duplicate categories, but rather clump related plugins together whether they are local or global.

pllim:
@ejeschke , dumping local and global plugins together in the same sub-menu still does not quite work because now I see two "Blink" (and "FBrowser") being listed and there is no way to know which one is global or local; It also confuses the user in the sense, "Which one am I supposed to use?"

ejeschke:
> now I see two "Blink" (and "FBrowser") being listed and there is no way to know which one is global or local; It also confuses the user in the sense, "Which one am I supposed to use?"

Yes, that is a problem.  By having different text associated we might be able to make it clearer, such as "Blink (across channels)" and "Blink (within channel)".

pllim:
I don't think there is a way to name the menu item differently from actual plugin name, is there?

ejeschke:
> I don't think there is a way to name the menu item differently from actual plugin name, is there?

I think so, using the `tab=` part.  It specifies how the plugin should show up in the tab or a menu.  But even if I am forgetting some detail here, it could be added to achieve the goal.

pllim:
> I think so, using the `tab=` part. It specifies how the plugin should show up in the tab or a menu. 

`tab` only changes the text in the actual tab, not menu. This is already in `main.py`:
```python
Bunch(module='Blink', tab='Blink Channels', workspace='right', ...),
```
But the menu only displays "Blink", not "Blink Channels".

ejeschke:
@pllim, when you get a chance can you rebase this?

pllim:
Rebased.

ejeschke:
@pllim, in commit 04c2a5cf341dd7662c4a6c4b0fe1d3b52f535b25 I added support for menu names in the plugin spec (Bunches).  You can see the example for setting the menu name for the `Blink` plugin to be "Blink Channels" if invoked as a global plugin, and "Blink Images" if invoked as a local one.  Sorry, but you many need to rebase one more time to bring these changes into this PR.

But now I think we have enough capabilities to complete this PR.  I want to ask a couple of the users around here if the categories make sense to them.  I don't think we want to be reordering these very often, if at all.  But I think we have some flexibility as long as the changes are done in the current "dev" branch.

pllim:
>  I want to ask a couple of the users around here

I agree. This should not be something that changes once set. Please let me know once you have feedback.

pllim:
I rebased and removed conflict. I'll continue working on this once I hear back from you.

pllim:
I have been thinking: Is it possible to let user customize the categories organization within, say, `plugin_Operations.cfg`? That way, we don't have to worry about designing the categories in such a way that it will *never* change. What do you think, @ejeschke ?

ejeschke:
That would be nice, from a user's point of view, although I think most users won't bother.  The organization scheme will mostly impact new users, who are trying to find their way around the program.  I think the main issue is that we just don't want to be rearranging that very often, if at all.

But since it is easy enough for us to change it in our respective customized versions of the reference viewer, I am now thinking we don't have to be so overly cautious, at least with regard to our deployments. Maybe we should just plow ahead with this.  I'll check it out and review the categorization one more time.

ejeschke:
@pllim, did you bounce the idea of rearranging the menus with anyone at STScI?  Any opinions?

pllim:
> with anyone at STScI? 

Good idea. I'll ask.

ejeschke:
@pllim, check out [this branch](https://github.com/naojsoft/ginga/tree/alt-categories) for an alternative idea of categorization.  This incorporates the idea that the global plugins are not separated out with duplicated sub-categories, but instead integrated into a whole.  LMK what you think.

pllim:
> check out this branch for an alternative idea of categorization

I like this approach! The exact order of the plugins can still be improved (for example, maybe put "Blink Channels [G]" just above "Blink Images") but I think this is cleaner over all. :+1:

ejeschke:
> I like this approach!

Ok, I am going to submit that branch as an alternate PR and we can continue the discussion there.

PR #451, opened 2017-04-18
Auto install HTML doc from RTD to Ginga package data

pllim:
Address documentation discussion as per https://github.com/ejeschke/ginga/issues/438#issuecomment-293968767 .  Also some minor PEP8 clean-ups here and there.

On Linux (and probably Mac), aside from Japanese kanji not displaying on `WBrowser`, it seems to work as intended. However, I couldn't able to get it to work properly on Windows (e.g., the main page would show but links do not work).

**UPDATE:** As per discussions below, this PR is updated to download HTML files from RTD (if not previously downloaded already) and then display that instead. Or if the download failed, "Help" will attempt to display online content. I'll defer full Windows support as future work (if there is demand for it). As a result of this new behavior, the existing "help.html" stub is not displayed anymore, but it is still used to discover installation path for HTML content.

What do you think, @ejeschke ?


ejeschke:
Hmm, I think the HTML is just "derived" files and probably shouldn't be checked in to Git. 

How about if we have the `ginga/doc` directory (and all underneath) installed as package data, and just make it an optional step to build the doc tree and copy it into `ginga/doc` before the user does `python setup.py install`?  Maybe something like:
```bash
$ python setup.py build_doc
$ python setup.py install
```
If the user omits this step then they just get a single page in the `doc` directory that has links to the online documentation.

pllim:
That is more complicated...

* Adds extra dependency on Sphinx. If user has very outdated version of Sphinx and gets a bunch of doc build traceback, we have to debug the resultant issues. 
* Significantly increase installation time, as it now also includes time needed to build the entire doc.
* Might have a chicken-and-egg problem because traditionally, Sphinx grabs API doc from installed `*.py` files. So if you require doc build at install time, it won't find any installed `*.py` to build from. Maybe there is a way around that; I am not sure.

p.s. Doc build step should be skipped on Windows since I can't get `WBrowser` to work on Windows anyway.

pllim:
And what about installation via egg. Is install-time doc build even possible in that case?

pllim:
Maybe the compromise between installing them as static pages and building them on install is to attempt to download https://readthedocs.org/projects/ginga/downloads/htmlzip/latest/ when "Help" is called, if the documents are missing, save the ZIP somewhere, and unzip it into package data area. Will have to see if there is an OS-agnostic way to unzip stuff in Python...

ejeschke:
That's an interesting idea. I bet there is a python zip module.

pllim:
@ejeschke , I updated this PR -- No more static HTML, but rather a new module to auto-download from RTD. What do you think?

ejeschke:
@pllim, I prefer this approach.  I notice that it fetches the latest documentation, but maybe we should just download the zip file matching the `ginga.__version__` (or closest release)?

pllim:
@ejeschke , currently https://readthedocs.org/projects/ginga/downloads/ only has "latest" or "stable". To support version matching, we will need to retroactively build all the releases. We will also need to handle "dev" matching, because just having "dev" in version tag does not mean I installed the latest "dev" version.

ejeschke:
Since the code seems identical in the base class of each plugin, I'm wondering if we should now declare a base class for both local and global plugins and put the `help()` method there.  That might come in handy for future commonalities between the plugin types.

ejeschke:
> currently https://readthedocs.org/projects/ginga/downloads/ only has "latest" or "stable". To support version matching, we will need to retroactively build all the releases.

Hmm, I thought we had it set to automatically build the docs on `readthedocs.org` when a commit is made in github?  I remember seeing multiple versions available there.  Maybe for the downloads they don't provide that option?

pllim:
Maybe they only keep downloads for "active" versions? https://readthedocs.org/projects/ginga/versions/

As for your other comments, they are all good ideas but I probably won't have time till next week to address them.

ejeschke:
@pllim, this is showing a conflict for `Control.py`.  Can you rebase against master, and I'll test this PR.

pllim:
@ejeschke , yes, I did resolve the conflict locally on disk. I am applying the rest of your comments before pushing it out to the PR. Should have it done some time today. Thank you for your patience!

pllim:
@ejeschke , I rebased and also do version matching etc, as you requested. Please review.

ejeschke:
@pllim, I like this idea.  But I have couple of concerns/comments.

I think we will have to do something to show the progress of the download or similar.  When I clicked "Help", the whole app froze for about a minute and a half while the GUI thread downloaded the help zipfile.  If I wasn't keeping an eye on the terminal I wouldn't have any idea what was going on.  Maybe showing a progress bar in the `WBrowser` might work.  I don't know if the `astropy` download has that kind of callback (progress) function, but `ginga` has one--maybe we could use it in this case for this purpose.

It would be good to offload the download into another thread from the thread pool.  This is what I do for catalog fetches and loading URLs from `Control`--that way the GUI stays somewhat responsive.  (Offloading to another process would be even better--but perhaps a bit more complicated).

I'm not sure if this is an HTML compatibility thing, but when I look at the manual in the `WBrowser` window after clicking help in a plugin (say, `Pick`), the `readthedocs` navigation on the left (different Ginga topics) does not work as it does in a firefox or chrome browser; that is, it doesn't open into sections and subsections to allow you to navigate.

If you like I can take a look at the first couple of these concerns and post an update to your PR?

pllim:
> I'm not sure if this is an HTML compatibility thing, but...

No, it is how RTD distributes the HTML builds. It confused me too and I reported it at rtfd/readthedocs.org#2803.

As for the other concerns, I'll have a look.

pllim:
@ejeschke , if you would be so kind to point me to code snippets for Ginga's progress bar and thread pools, it would speed up my next commit. 😄

ejeschke:
>  if you would be so kind to point me to code snippets for Ginga's progress bar and thread pools, it would speed up my next commit

Sure. :+1: 

For the downloader with callback fn, take a look at [Control.py](https://github.com/ejeschke/ginga/blob/master/ginga/rv/Control.py) in the `get_fileinfo()` method.  It is designed to be run in a separate thread, which is why the callback uses `gui_do()` to update the GUI with the progress.

To run your downloader code in a separate thread, just use `nongui_do()`.

ejeschke:
*And* it's a very good idea to use `assert_gui_thread()` and `assert_nongui_thread()` at the top of functions that are invoked from the gui.  So in the downloader it would have `assert_nongui_thread()`, and in your update gui function it would have `assert_gui_thread()`.  These are all "fv" functions, BTW.

ginga/rv/plugins/WBrowser.py
```diff
@@ -0,0 +1,119 @@
...
+        self.show_help()
+        self.gui_up = True
+
+    def show_help(self, plugin=None, no_url_callback=None):
+        """See `~ginga.GingaPlugin` for usage of optional keywords."""
+        if not Widgets.has_webkit:
+            return
+
+        from ginga.doc.download_doc import get_doc
+        self.fv.assert_gui_thread()
```
pllim: Is this the right place to put this in?

ginga/rv/plugins/WBrowser.py
```diff
@@ -0,0 +1,119 @@
...
+
+        def _dl_indicator(count, blksize, totalsize):
+            pct = (count * blksize) / totalsize
+            msg = 'Downloading: {:.2%} complete'.format(pct)
+            self.fv.gui_do(self.entry.set_text, msg)
+
+        helpurl = get_doc(logger=self.logger, plugin=plugin,
```
pllim: I could not figure out how to put this in `nongui_do()`, because if I do that, it returns a `Task`, not the URL text I was expecting. Also, putting `assert_nongui_thread()` before this line raised exception (something about attempting to do non-GUI stuff in a GUI thread).

ginga/rv/plugins/WBrowser.py
ejeschke: No, see below

ginga/rv/plugins/WBrowser.py
ejeschke: You have to split the work units so that they are chained from thread to thread via functions and not inline anymore.  So this would become:
```python
    self.fv.nongui_do(self._download_doc, plugin, _dl_indicator, no_url_callback)

def _download_doc(self, plugin, fn_indicator, no_url_callback):
    self.fv.assert_nongui_thread()

    # this can block as long as it takes without blocking the UI
    try:
        helpurl = get_doc(logger=self.logger, plugin=plugin, reporthook=fn_indicator)

    except Exception as e:
        self.fv.gui_do(self.fv.show_error, "Error downloading documentation: %s" % (str(e)))

    self.fv.gui_do(self._load_doc, helpurl, no_url_callback)

def _load_doc(self, helpurl, no_url_callback):
    self.fv.assert_gui_thread()
    if helpurl is None:
￼        self.fv.entry.set_text('')
￼        if no_url_callback is None:
              self.fv.show_error("Couldn't load web page %s": (helpurl))
￼        else:
              no_url_callback()
￼    else:
￼         self.browse(helpurl)
```
Yes it breaks up the flow, but unfortunately there is no easy way to have a single code stream switch between threads.

ginga/rv/plugins/WBrowser.py
```diff
@@ -0,0 +1,119 @@
...
+        btns.add_widget(btn, stretch=0)
+        btns.add_widget(Widgets.Label(''), stretch=1)
+        container.add_widget(btns, stretch=0)
+
+        self.show_help()
+        self.gui_up = True
+
+    def show_help(self, plugin=None, no_url_callback=None):
+        """See `~ginga.GingaPlugin` for usage of optional keywords."""
+        if not Widgets.has_webkit:
```
ejeschke: I think we need to do something here other than a no-op?  Maybe show an error, at the least.

ginga/rv/plugins/WBrowser.py
ejeschke: @pllim, see my notes in the code.

That's a good touch to implement the agnostic browser widget.  I have to make a note to add it to the "pg" widget set which I hope will eventually support the full reference viewer.

ginga/doc/download_doc.py
```diff
@@ -0,0 +1,173 @@
...
+
+    url = ('https://readthedocs.org/projects/ginga/downloads/htmlzip/'
+           '{}/'.format(rtd_version))
+    local_path = urllib.request.urlretrieve(url, **kwargs)[0]
+
+    with zipfile.ZipFile(local_path, 'r') as zf:
+        zf.extractall(data_path)
+
+    # RTD makes an undesirable sub-directory, so move everything there
+    # up one level and delete it.
```
ejeschke: Ugh.

ginga/gtk3w/Widgets.py
```diff
@@ -196,6 +216,12 @@ def set_enabled(self, tf):
         super(TextEntrySet, self).set_enabled(tf)
         self.entry.set_sensitive(tf)
 
+
+class GrowingTextEdit(WidgetBase):
+    def __init__(self, *args, **kwargs):
+        raise NotImplementedError
+
```
ejeschke: I think `GrowingTextEdit` was a hack a I had to do for Qt and wasn't needed for Gtk because of the way the widget works.  I don't think it was meant to be instantiated directly by users.  But I should review that--it was far enough back that my memory is fuzzy on it,

ginga/gtk3w/Widgets.py
```diff
@@ -1466,6 +1517,11 @@ def add_widget(self, child, row, col, stretch=0):
         self.widget.show_all()
 
 
+class ToolbarAction(WidgetBase):
+    def __init__(self):
+        raise NotImplementedError
+
```
ejeschke: Ditto, `ToolbarAction`.

ginga/qtw/Widgets.py
```diff
@@ -1792,6 +1827,11 @@ def start_drag(self):
             self._drag.start(QtCore.Qt.MoveAction)
 
 
+class WidgetMoveEvent(object):
+    def __init__(self, src_widget, child):
+        raise NotImplementedError
+
```
ejeschke: Where is `WidgetMoveEvent` used?

ginga/rv/plugins/WBrowser.py
pllim: @ejeschke , the "missing webkit" message is already shown in `build_gui()`. This ensures that the message does not disappear.

ginga/gtk3w/Widgets.py
pllim: Okay, maybe the `NotImplementedError` additions are overkill. Should I just remove them?

ginga/qtw/Widgets.py
pllim: Under GTK and GTK3. I didn't realize you also have `pgw` widgets. Perhaps I should also add `__all__` in that. Since you do `import *` from them, I think `__all__` will help that import to be cleaner.

pllim:
@ejeschke , I think this is ready now. Thanks for the threading tips; they work great!

Just make sure Subaru staff will not abuse the `WBrowser` feature like this... 😂 
screenshot from 2017-04-27 13-28-35

pllim:
>  it doesn't open into sections and subsections to allow you to navigate

Ah, yes. I see that now. The menu icon (3 horizontal bars) on top left of `index.html` does not open a menu when I click on it (Qt5, Linux), but the menu works properly when I open the same page on Firefox. Pretty sure it used to work on `WBrowser`... 🤔 

Also, I think a "back" button would be nice. Since if you click on a link, `WBrowser` might take you to some external website, and then you cannot go back to the doc.

pllim:
I added back and forward buttons. They work using Qt5 on Linux. As I mentioned earlier, GTK version is untested.

As for the navigation, for whatever reason, `_static/js/theme.js` was supposed to show the side menu when three-horizontal-bar icon is clicked; it doesn't on `QWebView` *and* on my Firefox on the same machine, but works on my smart phone. To test that on your own machine, resize your browser to very narrow width and see for yourself. I think this is not an issue for Ginga but the machine's browser installation/setting.

The workaround is to resize Ginga and/or `WBrowser` such that the three-horizontal-bar icon disappears on its own to reveal the side menu. Does this make sense?

Do you think this is okay to merge, or require more work?

ejeschke:
@pllim, when I first run the Help from a plugin, it displays the "downloading" page and then this is quickly replaced by the top-level content from "https://ginga.readthedocs.io/en/latest"--the download seems to be failing (but no error message in the log) and it seems to be falling back to fetching the general Help link and not falling back to either a `docutils` translation of the plugin docstring or even a plain text display of the ReST text.

On the bright side, the content fetched from readthedocs in this way does seem to get the correct javascript or css and the collapsible menu options work correctly.  Digging further...

ejeschke:
Sorry, there is one error message in the log, repeated twice: "download_doc.py:162 (get_doc) | attempted to get a local data file outside of the ginga tree"

pllim:
@ejeschke I have not seen that error before but I know where it is coming from. Please do the following and report your full traceback:
```python
>>> from ginga.doc.download_doc import get_doc
>>> get_doc()
```
I got `'file:/.../lib/python3.5/site-packages/ginga-2.6.4.dev1440-py3.5.egg/ginga/doc/index.html'`

ejeschke:
> Please do the following and report your full traceback

```python
$ python3
Python 3.5.2 (default, Nov 17 2016, 17:05:23) 
[GCC 5.4.0 20160609] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> from ginga.doc.download_doc import get_doc
>>> get_doc()
'file:/home/eric/pub/arch/64/lib/python3.5/site-packages/ginga-2.6.4.dev1440-py3.5.egg/ginga/doc/index.html'
```

pllim:
That is weird. `get_doc()` returned a string, so that means the doc is downloaded. Which plugin did you start that caused the error?

ejeschke:
I think I may have had problems from previous compiled `WBrowser` files in my `build` directory.  Clearing all that and all pyc/pycache and doing a clean install, it now downloads and then displays the page--but then it seems to go ahead and *additionally* fetch the main page online, which replaces the page dialed in to the plugin.  Pressing `Back` takes me back to the plugin page.

Subsequent `Help` presses from different plugins work ok, as the document seems to be found in the right place right away.  Do you think this is a race condition of some sort?

ejeschke:
Maybe we should go ahead and add a "Reload" button?

ejeschke:
> Just make sure Subaru staff will not abuse the WBrowser feature like this...

Well, I think if there is bad weather astronomers and operators can get bored. :smile_cat:

pllim:
> and additionally fetch the main page online

Yes, I see that as well. I think it is a race condition. This is because when `WBrowser` is started from a plugin's help button, `build_gui()` is called to display `index.html` (because it has no knowledge that it was invoked from a plugin); And I think this happens right after(?) `obj.show_help(...)` from `GingaPlugin.py`. Subsequent help button presses do not have this problem because `WBrowser` has already started, so `build_gui()` is not called. Do you know a good way around this?

> add a "Reload" button?

Sure.

ejeschke:
I think we should not have a `show_help()` at the end of `build_gui()` because that starts one download going.  Just commenting that out you can see from the animated download progress update that there are not two competing downloads going, and now it takes me straight to the plugin documentation.

pllim:
>  we should not have a `show_help()` at the end of `build_gui()`

But then how is the page going to be displayed when user starts `WBrowser` from global plugin menu or select `Help -> Documentation`?

ejeschke:
The "Help"->"Documentation" menu works as expected, because it does an additional "browse" after popping up the plugin.  Try it!

ejeschke:
I think if the user stars `WBrowser` from the "Plugins" menu (or "Operations") it's ok for it just to display a blank page.  They are basically just popping up a web browser.

ejeschke:
We could maybe also add a "Home" button to take you directly to the top of the documentation.  It would be nice if this could fetch the local doc if it exists and if not go to the online version.

ejeschke:
Here's what I am seeing with gtk3/python3 on Linux:
```python
** Message: console message: file:///home/eric/pub/arch/64/lib/python3.5/site-packages/ginga-2.6.4.dev1440-py3.5.egg/ginga/doc/index.html#pick @60: ReferenceError: Can't find variable: READTHEDOCS_DATA

** Message: console message: file:///home/eric/pub/arch/64/lib/python3.5/site-packages/ginga-2.6.4.dev1440-py3.5.egg/ginga/doc/_static/jquery.js @2: jQuery.Deferred exception: Can't find variable: SphinxRtdTheme


(ginga:304): Gtk-WARNING **: drawing failure for widget 'WebKitWebView': invalid value (typically too big) for the size of the input (surface, pattern, etc.)

```

This may just indicate that WebKit is a little more fragile under Gtk, or perhaps needs some additional configuration.  But the download, messages, etc. all looks fine.

pllim:
@ejeschke , I think I implemented all your requests. Please grab the updated PR and have another look. I even threw in a bonus "stop" button.

ejeschke:
Under gtk2, I get the following error:
```python
2017-04-27 10:49:31,859 | E | download_doc.py:162 (get_doc) | addinfourl instance has no attribute '__exit__'
```

pllim:
That is actually a Python 2 incompatibility (http://stackoverflow.com/questions/30627937/traceback-attributeerroraddinfourl-instance-has-no-attribute-exit/30628128) in `_find_rtd_version()` I think. I'll see what I can do about that tomorrow...

ejeschke:
> Please grab the updated PR and have another look. I even threw in a bonus "stop" button.

:meat_on_bone: 

I think we will eventually want icons for these buttons.  But that can be a later enhancement.

ejeschke:
When I look at the `Pick` plugin doc through this help, some of the section headings show up as "nonprintable paragraph markers" as shown in the attached screenshot (see text near bottom).  Any idea how to improve that?

help_scrnshot

pllim:
Re: `Pick` section headers.

I am not sure if there is much we can do about that. What you are seeing is due to demotion of header level because RTD download mashed all the sub-pages into one page. So, when you view online, it is rendered as:
```html
<div class="section" id="usage">
<h2>Usage<a class="headerlink" href="#usage" title="Permalink to this headline">¶</a></h2>
<div class="section" id="defining-the-pick-area">
<h3>Defining the pick area<a class="headerlink" href="#defining-the-pick-area" title="Permalink to this headline">¶</a></h3>
<p>The default pick area is defined as a rectangle of approximately 30x30
```
But in the download doc is rendered as:
```html
<div class="section" id="usage">
<h7>Usage<a class="headerlink" href="#usage" title="Permalink to this headline">¶</a></h7>
<div class="section" id="defining-the-pick-area">
<h8>Defining the pick area<a class="headerlink" href="#defining-the-pick-area" title="Permalink to this headline">¶</a></h8>
<p>The default pick area is defined as a rectangle of approximately 30x30
```
Seven levels down!

pllim:
@ejeschke , it should now run on Python 2, please re-test with your GTK2 environment.

I also replaced the buttons with existing icons that I think are the closest match to what they are supposed to do. I am not icon artist, so if you want new flashy icons for this, someone else better do it. All needed to be done then is to add new icon images and replace the filenames.

I also added a "Help" button to the help plugin (very meta).

ejeschke:
When I test with python2/gtk2 now I get the following error:
```python
** Message: console message: file:///home/eric/pub/arch/64/lib/python2.7/site-packages/ginga-2.6.4.dev1443-py2.7.egg/ginga/doc/index.html#pick @60: ReferenceError: Can't find variable: READTHEDOCS_DATA


** (ginga:28155): WARNING **: The program 'ginga' received an X Window System error.
This probably reflects a bug in the Adobe Flash plugin.
The error was 'BadAlloc (insufficient resources for operation)'.
  (Details: serial 14875 error_code 11 request_code 53 minor_code 0)

```

I'm not sure what's going on, but I have a feeling that this is not a problem related to Ginga.

pllim:
Those remainder unresolved GTK errors are not familiar to me. I agree that they seem unrelated to Ginga itself, but rather to the toolkit.

pllim:
http://stackoverflow.com/questions/11528691/code-runs-in-shell-but-gives-cant-find-variable-error-when-run-from-a-file might give a hint to "can't find variable error"? But I am not sure where to apply this.

The "X Window System error" is more obscured. There are several bug reports out there about something similar but not exactly the same, and they are all marked as "fixed".

Do you have the latest webkit package installed for both GTK2 and GTK3?

ginga/doc/download_doc.py
```diff
@@ -0,0 +1,179 @@
...
+    else:
+        # plugin_page = None
+        plugin_name = None
+
+    try:
+        index_html = _download_rtd_zip(reporthook=reporthook)
+
+    # Download failed, use online resource
+    except Exception as e:
+        url = 'https://ginga.readthedocs.io/en/latest/'
```
ejeschke: Can this URL be set to go to the plugin section that the user is asking help for?

ejeschke:
> Do you have the latest webkit package installed for both GTK2 and GTK3?

I suspect not, but they are the stable versions for the OS (Ubuntu 16.04) that I am running.  What is interesting is that if I simply start `WBrowser` and browse to the online documentation it renders decently.  So there is something about the downloaded stack of HTML that it is not compatible with.

I'm wondering if (as a temporary workaround) we could put in a check for the toolkit in `get_doc()`, and just use a link to the online documentation if it is not qt (somewhat similar to what we are doing now if the download fails in the try/except).  I really dislike these forms of hacks, but sometimes they are necessary, and having the help just throw up a blank screen and an error message in the log does not sound good.  In the appropriate place (say inside the try/except), we could do something like:
```python
from ginga import toolkit
...
if not toolkit.family.startswith('qt'):
    raise ValueError("Downloaded documentation not compatible with this UI toolkit browser")
```
this will cause it to skip into the part were it just sets the URL to the online doc instead of the local doc.

ginga/doc/download_doc.py
ejeschke: Then this could be the fallback solution for Gtk.

pllim: Yes. It will be part of next commit.

pllim:
@ejeschke , please grab the latest commit and test on GTK again. Thanks!

ejeschke:
> please grab the latest commit and test on GTK again. Thanks!

I think the URL is incorrect, e.g.: https://ginga.readthedocs.io/en/latest/plugins_local/pick.html

It gives me a "SORRY, this page does not exist yet" error.

ejeschke:
You can simulate with qt by just inserting an error before or after the download to force the `except` clause.

pllim:
Yes, sorry. I should have tested it. There's a missing "manual" in the URL. It used to worked before I commented it out, so perhaps something went missing between the subsequent commits.

https://ginga.readthedocs.io/en/latest/manual/plugins_local/pick.html

pllim:
@ejeschke , the plugin URL should work now; I tested it. Please try again.

Are you sure you don't want to display plugin docstring for GTK and *always* use online doc?

ejeschke:
> the plugin URL should work now; I tested it. Please try again.

:+1: It works!

> Are you sure you don't want to display plugin docstring for GTK and always use online doc instead?

The best scenario would be to try to display online first (ideally for the matching version, but that might be asking too much) and if that fails, to fall back to the docstring (in some way, whether rendered by the browser or simple text widget).  I don't know how easy it is to detect a failure with the `WebKit` browser.

ginga/rv/plugins/WBrowser.py
```diff
@@ -0,0 +1,177 @@
...
+        try:
+            if url_is_content:  # This was load_html()
+                self.browser.load_html_string(url)
+                self.entry.set_text('')
+            else:
+                self.logger.debug("Browsing '{}'".format(url))
+                self.browser.load_url(url)
+                self.entry.set_text(url)
+        except Exception as e:
+            self.fv.show_error("Couldn't load web page: {}".format(str(e)))
```
pllim: > The best scenario would be to try to display online first (ideally for the matching version, but that might be asking too much) and if that fails, to fall back to the docstring (in some way, whether rendered by the browser or simple text widget). I don't know how easy it is to detect a failure with the WebKit browser.

Does a WebKit load failure go here? If so, it might be possible to replace `show_error` with some kind of docstring display. But if not, then there is no point modifying this.

ejeschke: That's the question.  It doesn't seem to me that the browser raises an exception if the page cannot be found, so I'm not sure how to implement this at the moment.

I think we can go ahead and merge.  I'm satisfied that gtk users should see a decent representation of the help page if they are online.

PR #455, opened 2017-05-01
[WIP] Improve LineProfile plugin

### LINKED ISSUE #427: LineProfile feedback
@ejeschke , I am using latest dev of Ginga including your changes in 00bcc853ed9159d17dde003e33be2a7ecd6b834c. I find that:

- [ ] NAXISn does not have a default value on start-up or image change or defocus, resulting in error messages. Maybe should default to NAXIS3?
- [ ] Y-axis label "flux" is not always accurate representation of the data. I don't have a good alternative label that is general enough though.
- [x] ~~The plugin does not work for other image cubes that I have. One is an IRIS image provided by SunPy when this plugin was developed some time ago. Another is a fake JWST image cube.~~ EDIT: It was a fluke. I restarted Ginga and it works now.

pllim:
Fix #427.

Also fix a bug where "inherit_primary_header" no longer works.

I will need these fixes for JWST NIRCam support.

### DIFF
--- a/ginga/rv/plugins/LineProfile.py
+++ b/ginga/rv/plugins/LineProfile.py
@@ -58,7 +58,7 @@ class LineProfile(GingaPlugin.LocalPlugin):
         self.tw = None
         self.mark_data_x = [None]
         self.mark_data_y = [None]
-        self.y_lbl = 'Flux'  # Can be changed in GUI
+        self.y_lbl = ''
         self.x_lbl = ''
 
         self.gui_up = False
@@ -75,25 +75,18 @@ class LineProfile(GingaPlugin.LocalPlugin):
                                width=400, height=300)
         ax = self.plot.add_axis()
         ax.grid(False)
+        self._ax2 = self.plot.ax.twiny()
 
         w = Plot.PlotWidget(self.plot)
         w.resize(400, 300)
         vbox.add_widget(w, stretch=0)
 
         fr = Widgets.Frame("Axes controls")
-        vbox3 = Widgets.VBox()
-        captions = (('Y Label:', 'llabel', 'ylabel', 'entryset'), )
-        w, b = Widgets.build_info(captions, orientation=orientation)
-        self.w.update(b)
-        b.ylabel.set_tooltip('Plot label for Y-axis')
-        b.ylabel.set_text(self.y_lbl)
-        b.ylabel.add_callback('activated', lambda w: self.set_ylabel_cb())
-        vbox3.add_widget(w)
         self.hbox_axes = Widgets.HBox()
         self.hbox_axes.set_border_width(4)
         self.hbox_axes.set_spacing(1)
-        vbox3.add_widget(self.hbox_axes)
-        fr.set_widget(vbox3)
+        fr.set_widget(self.hbox_axes)
+
         vbox.add_widget(fr, stretch=0)
 
         btns = Widgets.HBox()
@@ -216,17 +209,6 @@ class LineProfile(GingaPlugin.LocalPlugin):
             children[pos-1].set_state(tf)
             self.redraw_mark()
 
-    def set_ylabel_cb(self):
-        try:
-            val = self.w.ylabel.get_text()
-        except Exception as e:
-            errmsg = 'Error setting Y-label: {0}'.format(str(e))
-            self.fv.show_status(errmsg)
-            self.logger.error(errmsg)
-        else:
-            self.y_lbl = val
-            self.redraw_mark()
-
     def close(self):
         self.fv.stop_local_plugin(self.chname, str(self))
         self.gui_up = False
@@ -287,13 +269,32 @@ class LineProfile(GingaPlugin.LocalPlugin):
         slice_obj = self._slice(naxes, mk=mark)
         plot_y_axis_data = mddata[slice_obj]
 
+        # If few enough data points, add marker
+        if len(plot_y_axis_data) <= 10:
+            marker = 'o'
+        else:
+            marker = None
+
+        # https://github.com/matplotlib/matplotlib/issues/3633/
+        ax2 = self._ax2
+        ax2.patch.set_visible(False)
+
         self.clear_plot()
         self.plot.plot(plot_x_axis_data, plot_y_axis_data,
-                       xtitle=self.x_lbl, ytitle=self.y_lbl)
+                       xtitle=self.x_lbl, ytitle=self.y_lbl, marker=marker)
+
+        # Top axis to show pixel location across X
+        ax2.cla()
+        xx1, xx2 = self.plot.ax.get_xlim()
+        ax2.set_xlim((xx1 - self._crval) / self._cdelt,
+                     (xx2 - self._crval) / self._cdelt)
+        ax2.set_xlabel('Index')
+        self.plot.draw()
 
     def _slice(self, naxes, mk):
         # Build N-dim slice
         slice_obj = [0] * naxes
+        i_sel = self.selected_axis - 1
 
         # For axes 1 and 2
         if mk is not None:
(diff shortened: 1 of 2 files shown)


ginga/rv/plugins/LineProfile.py
```diff
@@ -248,25 +258,43 @@ def _plot(self, mark=None):
...
+        ax2.patch.set_visible(False)
+
+        self.clear_plot()
+        self.plot.plot(plot_x_axis_data, plot_y_axis_data,
+                       xtitle=self.x_lbl, ytitle=self.y_lbl, marker=marker)
+
+        # Top axis to show pixel location across X
+        ax2.cla()
+        xx1, xx2 = self.plot.ax.get_xlim()
+        ax2.set_xlim((xx1 - self._crval) / self._cdelt,
```
pllim: This is so that I know where exactly in the cube that a feature of interest appears. Just showing the velocity is not very useful to pinpoint location in cube at a glance without playing through `MultiDim`.

ginga/rv/plugins/LineProfile.py
```diff
@@ -275,45 +303,55 @@ def _slice(self, naxes, mk):
 
         # For axis > 3
         for i in range(2, naxes):
-            slice_obj[i] = int(round(self.image.revnaxis[i-2] + 1))
+            if i != i_sel:
+                slice_obj[i] = 0  # int(round(self.image.revnaxis[i-2] + 1))
```
pllim: I really don't know what `int(round(self.image.revnaxis[i-2] + 1))` is for but it breaks when the cube has more than 3 dimensions. Setting to 0 fixes it but does it change the plugin behavior?

ginga/rv/plugins/LineProfile.py
```diff
@@ -275,45 +303,55 @@ def _slice(self, naxes, mk):
...
-                units = self.image.get_keyword('CUNIT%d' % i, None)
-                if units is not None:
-                    self.x_lbl += (' (%s)' % str(units))
-            else:
-                self.x_lbl = ''
-            # Assume Y label should always be flux?
-            self.y_lbl = 'Flux'
-            return axis
+                self.logger.error("Missing FITS keyword: {}".format(str(e)))
+                crval_i = 0
```
pllim: This should be reasonable enough default if those keywords are not found.

ginga/rv/plugins/LineProfile.py
```diff
@@ -275,45 +303,55 @@ def _slice(self, naxes, mk):
...
+            self._crval = crval_i
+            self._cdelt = cdelt_i
+
+            units = self.image.get_keyword('CUNIT{}'.format(i), None)
+            if units is not None:
+                self.x_lbl += (' ({})'.format(units))
+
+            # Get pixel value info from header
+            self.y_lbl = self.image.get_keyword('BTYPE', 'Flux')
+            bunit = self.image.get_keyword('BUNIT', None)
```
pllim: I find that calibrated HST images and also the example image from SunPy has these keywords, so using these to set Y-axis label.

ginga/util/io_fits.py
```diff
@@ -157,19 +156,18 @@ def load_hdu(self, hdu, dstobj=None, **kwargs):
 
         return dstobj
 
-
     def load_file(self, filespec, numhdu=None, dstobj=None, memmap=None,
                   **kwargs):
-
+        inherit_primary_header = kwargs.pop('inherit_primary_header', False)
```
pllim: This is the fix for "inherit_primary_header" bug. It is unrelated to `LineProfile` but I throw it in here because I need this fix for the test image to work because it has messed up headers.

ginga/rv/plugins/LineProfile.py
ejeschke: I think this is so that the plugin updates correctly when then the slice is changed through `MultiDim` or other method that manipulates an `AstroImage`.

The `naxispath` is the path through the multidimensional data *after* the first two axes.  And `revnaxis` is that reversed.  The loop is iterating through the dimensions beyond 2.

pllim: I get `IndexError` for my test image. It has the following extensions but the cube of interest is in EXT 1.
```
Filename: nrca1_47Tuc_subpix_dither1_newpos_uncal.fits
No.    Name         Type      Cards   Dimensions   Format
  0  PRIMARY     PrimaryHDU     133   ()      
  1  SCI         ImageHDU        10   (2048, 2048, 10, 1)   float32   
  2  PIXELDQ     ImageHDU        10   (2048, 2048)   int32 (rescales to uint32)   
  3  GROUPDQ     ImageHDU        10   (2048, 2048, 10, 1)   uint8   
  4  ERR         ImageHDU        10   (2048, 2048, 10, 1)   float32   
  5  ZEROFRAME   ImageHDU         9   (2048, 2048, 1)   float32   
  6  ASDF        ImageHDU         7   (3495,)   uint8   
```

pllim: Also same error if the cube is `(2048, 2048, 10, 2)`. FYI.

pllim: > I think this is so that the plugin updates correctly when then the slice is changed through `MultiDim`...

I just tested this theory using an unmodified version of `LineProfile` and disproved it. When I play through the cube in `MultiDim`, the plot does not change. Also, for 3D cube, the value is always 1. So, I still don't understand what is the purpose of this complex calculation.

pllim:
Digging up some old correspondence, I found this:

> I am wondering whether the calculation for the axis itself is correct, even if the keywords ARE there.
> The  axis calculated is the X axis of the plot and the slice is the Y axis.  The X axis is calculated currently as 
>
> `axis = crval_i + np.arange(0, naxis_i, 1) * cdelt_i`
> 
> where "i" is the index of the axis being selected for the line profile.
>
> It seems that this assumes that:
>
> 1) CRVALi is the starting value of the series (i.e. CRPIXi is 0) when in fact usually CRPIXi conventionally defined as the midpoint of the NAXISi (although it needn't even be within the range of 0..NAXISi, to my understanding).
>
> 2) The values increase linearly along the axis according to the CDELTi (normally the pixel scale) and the number of pixels.
>
> I think neither 1 or 2 can be assumed.  Possibly for the case of selected axis = 2 ("z" dimension), this may be true, depending upon convention for this kind of cube.

@bhilbert4 , can you please comment on this? At least for NIRCam cube, how is CRVAL and CRDELT going to be used? And are they sufficient, or are there other additional/different keywords to be used?

ginga/rv/plugins/LineProfile.py
ejeschke: For a normal cube this would be expected, I think, since you are sending a "straight, perpendicular" ray through the cube to make the plot and so just changing the slice isn't going to change the pixels that the ray passes through.

The graph should update however if you change axes.

I need to look at the code more closely to reacquaint myself with what is going on there.  I didn't author that one and only sort of understand what it is supposed to be doing.  I now understand its usefulness, but given the wide variety of what the higher dimensions in these multidimensional objects *represents*, I think the hardest thing will be figuring out how to label the axes.  I suppose there should be some sort of quasi FITS "standard" that covers that...

bhilbert4:
For NIRCam apertures CRPIXi will indeed, in most cases, be in the center of the detector (e.g. pixel (1023.5, 1023.5) for full frame images). CDELTi is indeed the linear pixel scale for the detector. For NIRCam, this will be ~8e-6 degrees per pixel. These keywords do not completely describe the distortion of the NIRCam detectors. There are higher order terms. ds9 uses only the linear terms.

pllim:
So, the correct formula is actually this?
```python
n = np.arange(naxis_i) - (crpix_i - 1)
axis = crval_i + n * cdelt_i
```

pllim:
Note: I am just going to open a new PR due to conflicts, but leaving this open for now as a reminder.

PR #458, opened 2017-05-04
Fix inherit_primary_header setting being ignored

pllim:
This was buried in #455 but is actually a standalone fix, so I put it in a new PR.


ejeschke:
Looks OK to me.

PR #461, opened 2017-05-05
MultiDim improvement

### LINKED ISSUE #459: Cannot switch image via Thumbs when MultiDim is active
You can reproduce with the following steps:

1. Open an image with multiple extensions. In my case, image display chooses Extension 1 (as EXT 0 is just a header).
2. Open `MultiDim`. For whatever reason, it automatically jumps to EXT 0, which is the primary header.
3. Now I click on EXT1 thumbnail on `Thumbs` but it will not switch image display to EXT1. Instead the highlighting jumps back to EXT0 thumbnail.
4. Only when I change the extension via `MultiDim`, I can actually switch the displayed image.

pllim:
Fix #459 , follow up of a9943ebfb . @ejeschke , I think I fixed the behavior of switching active index when image is changed from another plugin (see https://github.com/ejeschke/ginga/issues/459#issuecomment-299299204 discussions). Also added `AstroTable` handling (there is a bit of lag there but we can figure out performance another time since it is not used by a lot of people) and minor PEP8 clean-ups.

The only thing I cannot figure out is how to move the slider to the actual slice being displayed when I switch away from a cube and switch back. Slider resets to 1 but the slice info (and slice image displayed) is the last active slice.

TODO:
- [x] Figure out fix for slider behavior.

### DIFF
--- a/ginga/rv/plugins/MultiDim.py
+++ b/ginga/rv/plugins/MultiDim.py
@@ -15,6 +15,7 @@ from ginga.misc import Future
 from ginga import GingaPlugin
 from ginga.util.iohelper import get_hdu_suffix
 from ginga.util.videosink import VideoSink
+from ginga.table.AstroTable import AstroTable
 
 import numpy as np
 import matplotlib.pyplot as plt
@@ -366,7 +367,10 @@ class MultiDim(GingaPlugin.LocalPlugin):
             self.fv.switch_name(chname, imname)
 
             # Still need to build datacube profile
-            mddata = self.image.get_mddata()
+            if isinstance(self.image, AstroTable):
+                mddata = None
+            else:
+                mddata = self.image.get_mddata()
             if mddata is not None:
                 dims = list(mddata.shape)
                 dims.reverse()
@@ -539,7 +543,7 @@ class MultiDim(GingaPlugin.LocalPlugin):
     def redo(self):
         """Called when an image is set in the channel."""
         image = self.channel.get_current_image()
-        if (image is None):
+        if image is None:
             return True
 
         path = image.get('path', None)
--- a/ginga/rv/plugins/MultiDim.py
+++ b/ginga/rv/plugins/MultiDim.py
@@ -62,7 +62,6 @@ class MultiDim(GingaPlugin.LocalPlugin):
         self.play_image = None
         self.play_axis = 2
         self.play_idx = 1
-        self.play_last = {}  # cache to store last active slice for each image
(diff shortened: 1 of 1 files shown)


ejeschke:
> The only thing I cannot figure out is how to move the slider to the actual slice being displayed when I switch away from a cube and switch back. Slider resets to 1 but the slice info (and slice image displayed) is the last active slice.

The active slice in an image can be figured out by looking at the `naxispath` attribute.

ejeschke:
This looks great, @pllim.  Should I merge or do you want to work on getting the slice control working?

pllim:
Please wait... :smile:

pllim:
OK it behaves for my test images now. Not sure if implementation is overly complicated or not but it works. Please review. Thanks!

ginga/rv/plugins/MultiDim.py
```diff
@@ -288,6 +304,7 @@ def play_axis_change():
                         self.play_max = dims[self.play_axis]
 
                     self.play_idx = self.play_indices[n - 2]
+                    self.play_last[imname] = self.play_idx
```
ejeschke: Is there anywhere that items get removed from `play_last`?

ejeschke: I think it should be possible to tell the slice from the `naxispath` attribute in the `AstroImage`.  I think this would be preferable to remembering which was the last item played through `MultiDim` (because there are other ways that the slice can be set externally to `MultiDim`).  This would also have the advantage that the information would persist even if `MultiDim` is closed and reopened.  And it would simplify the code here.

ejeschke: If you look at the `set_naxis()` method it is doing a
```python
image.set_naxispath(self.naxispath)
```
which means that the representation inside `MultiDim` of `naxispath` is compatible with that of the image.  So in `redo()` I think we can do something like
```python
naxispath = image.naxispath.copy()
```
and then set the controls from that.

pllim: Yes, I agree. I couldn't find any good API documentation for `naxispath` usage, so I used this workaround to at least get the behavior right first. I'll put it on my to-do list to incorporate your suggestions. Thank you.

pllim:
@ejeschke , I use `naxispath` as you suggested. I tested with several of my images of different dimensions (2-4D) and it works for me. Please review.

ejeschke:
Looks good to me, @pllim.  Thanks!

Merging.

PR #462, opened 2017-05-06
LineProfile now accepts region selection

### LINKED ISSUE #259: Extend LineProfile to accept aperture
This was requested in spacetelescope/stginga#40, but the parts where user can draw some kind of aperture and then use it for simple statistics up the ramp would be useful in general, no?

We should probably address #188 ~~and merge #187~~ first though.

### LINKED ISSUE #427: LineProfile feedback

pllim:
LineProfile GUI improvements and now accepts region selection.

Fix #427 
Fix #259 
Fix spacetelescope/stginga#40  c/c @bhilbert4
Supersede #455 

TODO:
- [x] Add doc with updated screenshot
- [x] Implement user settings to allow customization
- [x] ~~Figure out how to set default NAXIS without Ginga core dump~~ (I give up. This can be "future work".)


ginga/rv/plugins/LineProfile.py
```diff
@@ -155,29 +180,39 @@ def build_gui(self, container):
...
+                    self.selected_axis = None
                     chkbox.set_enabled(False)
-                else:
-                    # Add callback
-                    self.axes_callback_handler(chkbox, i)
+                    continue
+
+                # Add callback
+                self.axes_callback_handler(chkbox, i)
+
```
pllim: I attempted to set default NAXIS by doing `chkbox.set_state(True)` around here. But it results in Ginga crashing with core dump. Not sure where else I can set it.

ginga/rv/plugins/LineProfile.py
```diff
@@ -241,229 +245,373 @@ def redo(self):
...
-            # Assume Y label should always be flux?
-            self.y_lbl = 'Flux'
-            return axis
+                self.logger.error("Missing FITS keyword: {}".format(str(e)))
+                crval_i = 0
+                cdelt_i = 1
+                crpix_i = 1
+
+            n = np.arange(naxis_i) - (crpix_i - 1)
+            axis = crval_i + n * cdelt_i
```
pllim: I use a new formula that includes `CRPIXi` here.

ginga/rv/plugins/LineProfile.py
```diff
@@ -88,47 +92,68 @@ def build_gui(self, container):
...
+        b.delete_all.add_callback('activated', self.clear_all_cb)
+        b.delete_all.set_tooltip("Clear all marks")
 
         vbox2 = Widgets.VBox()
-        vbox2.add_widget(btns, stretch=0)
-        vbox2.add_widget(Widgets.Label(''), stretch=1)
+        vbox2.add_widget(w, stretch=0)
+
+        mode = self.canvas.get_draw_mode()
+        captions = (('Move', 'radiobutton', 'Draw', 'radiobutton',
```
pllim: Here, I replaced old GUI with new-style region drawing interface.

ginga/rv/plugins/LineProfile.py
```diff
@@ -241,229 +245,373 @@ def redo(self):
...
+            self.y_lbl = self.image.get_keyword('BTYPE', 'Flux')
+            bunit = self.image.get_keyword('BUNIT', None)
+            if bunit is not None:
+                self.y_lbl += (' ({})'.format(bunit))
 
         except Exception as e:
-            errmsg = "Error loading axis %d: %s" % (i, str(e))
+            errmsg = "Error loading axis {}: {}".format(i, str(e))
             self.logger.error(errmsg)
             self.fv.show_error(errmsg)
```
pllim: Hmm. I thought I removed this but maybe not. Didn't seem to bother me much now anyway.

ginga/rv/plugins/LineProfile.py
```diff
@@ -24,43 +24,46 @@ class LineProfile(GingaPlugin.LocalPlugin):
...
+        canvas.set_draw_mode('draw')
+        canvas.register_for_cursor_drawing(self.fitsimage)
         canvas.setSurface(self.fitsimage)
         self.canvas = canvas
 
-        # For "marks" feature
-        self.mark_radius = 10
-        self.mark_style = 'cross'
-        self.mark_color = 'purple'
-        self.select_color = 'cyan'
```
pllim: I find having multiple colors to be quite confusing, especially when there are multiple plugins active. So, I removed one of the colors. Planning to make this color user-configurable.

ejeschke:
@pllim, you might consider modifying the transition of the Move/Draw/Edit buttons slightly.  If you play with "Cuts" you can get a sense for how the flow works there.  It would be nice if it could be reasonably consistent between `LineProfile` and `Cuts` since there are similarities.

The way it works in "Cuts" is:
- when starting up for the first time, or when there is no cuts available to manipulate, only "Draw" is selected and others are disabled.
- as soon as a cut is drawn, the "Move" and "Edit" states become selectable *and* the state shits to "Move".  This assumes that the user does not want to draw a new cut immediately (to replace the current cut), but may want to reposition the cut.  At this point the user can reposition the cut simply by click and/or dragging, or select "Edit" to edit the cut, or "Draw" to *replace* the cut.
- At any time while a cut is selected, regardless if the state is in "Edit", "Move" or "Draw", if the user draws with the right mouse button it replaces the cut as though they had done a "Draw" (old-style interface is preserved)

If the user selects "New cut" from the drop down, then again "Move" and "Edit" become disabled and "Draw" is selected.

ejeschke:
@pllim, just so I understand correctly, with the new shapes, instead of just plotting the values along the path, it is plotting the mean of the region cutouts along the path?

ejeschke:
And if any of the `CRVAL*`, `CDELT*` and `CRPIX*` keywords are missing it just assumes a pixel scale along the axes?

pllim:
> it is plotting the mean of the region cutouts along the path?

Yes.

> it just assumes a pixel scale along the axes?

Yes.

ejeschke:
Would either ellipse or box types be useful?  Box is similar to a rectangle but is defined by a center point and radii for width and height. Boxes and ellipses can also have rotation, which rectangles don't have.

ejeschke:
> Would either ellipse or box types be useful? Box is similar to a rectangle but is defined by a center point and radii for width and height. Boxes and ellipses can also have rotation, which rectangles don't have.

I don't think that would add any complexity beyond adding them to the types list.

pllim:
Thanks for the feedback. I'll incorporate your suggestions for the radio buttons when I get the chance.

> I don't think that would add any complexity beyond adding them to the types list

Not so far. So yes, I can add them. However, if user wishes to have complete controls of the parameters (e.g., being able to type in exactly what radius etc), then the GUI gets complex very fast with each new shape. However, we can worry about that later.

nabobalis:
I'm not very good on code review but I've tested this branch on the data I have available to me and it works well. 

I think very long term, cut/slit/LP could be merged into one "meta" plugin which enables individual features of each via toggles.

pllim:
@ejeschke , regarding your comment on the Move/Draw/Edit behavior, I can fulfill the first two bullet points but not the last. See below:

>  if the user draws with the right mouse button

I find that this is no longer possible. When I click on the right mouse button, I go into "adjust image stretch" mode. Also, having too many ways to do the same thing can be confusing anyway, so I suggest we just stop supporting this. If user wants to replace a mark, user can delete the offending mark and draw a new one.

pllim:
@ejeschke , I think this is ready for merge if you are happy with it. I also fixed some Sphinx warnings in "quick ref".

ejeschke:
>  I suggest we just stop supporting this. If user wants to replace a mark, user can delete the offending mark and draw a new one.

@pllim, I think it is ok not to support the right-click draw, as many users are now on laptops using touchpads. You possibly are using a custom bindings file that binds right drag to contrast adjustment (a la ds9).

> I think this is ready for merge if you are happy with it. I also fixed some Sphinx warnings in "quick ref"

I'll take a look in the next day or two.

pllim:
> You possibly are using a custom bindings file

Yes, I am indeed!

ejeschke:
> Yes, I am indeed!

In that case the old-style drawing should continue to work for those using the standard binding.

pllim:
Sure, but I don't have motivation to add that support since most of my user-base are DS9 users and will certainly use DS9 bindings.

ejeschke:
> Sure, but I don't have motivation to add that support since most of my user-base are DS9 users and will certainly use DS9 bindings.

@pllim, I understand, but since Ginga supports the user to flexibly use their own bindings (as you did), I would like to insure that (as much as possible) users are able to use plugins with their selected choices, including (and maybe especially) with the "standard" bindings.

pllim:
So, is not supporting old style a deal breaker here?

ejeschke:
> So, is not supporting old style a deal breaker here?

@pllim, it is not an either/or/deal-breaker, simply a matter of testing, because the "old style" drawing is simply supporting a draw operation in the case where the user has either the "move" or "edit" checkboxes checked.  I will happily test it if you find it onerous  (BTW, you can achieve the same test even with using "ds9" bindings by putting the viewer into "draw" mode (e.g. by pressing space bar) and drawing a new shape with the touchpad/left mouse button while the plugin is in "move" or "edit" mode.)

If they draw a new shape it will invoke the draw callback.  That callback should check whether an existing mark is selected in the drop down or "New" is selected.  If "New", it creates a new object as expected.  If an existing mark is selected, it should replace the old mark/shape with the drawn one.  In either case, `redo()` is called afterward to update the plot.

pllim:
@ejeschke , okay, I added the "replace mark" functionality. Please review.

ejeschke:
> okay, I added the "replace mark" functionality. Please review.

Thank you, @pllim!  This now behaves similarly to the UI in `Cuts` and `PixTable`.

ejeschke:
This looks good to me.  My only remaining quibble is that the matplotlib plot seems potentially a little small, or at least without a way to resize it.  Maybe in the future we can enhance it with a splitter control or something (maybe similar to what is done in `Pick`).

Merging.

pllim:
I think the current size is still okay for quick-look. Thanks!

pllim:
Re: splitter control in `Pick` -- What exactly do you mean? I don't see any way to make the plot (e.g., FWHM) on `Pick` larger.

ejeschke:
If you look carefully just above the frame title for the lower "Pick" area there is a splitter control.  If you pull it down the plots are enlarged.

ejeschke:
I should say that `Cuts` and `Histogram` also lack such a control.  It would be nice to be able (in general) to make the plots larger vertically as it adds some readability for the user at times.

pllim:
> I should say that `Cuts` and `Histogram` also lack such a control.

Looks like you fix those in 8a8b16afa815fb2cb76 , correct?

ejeschke:
> Looks like you fixed those in 8a8b16a , correct?

Correct.

PR #469, opened 2017-05-14
Reformatting, clarification, add reference tag

supersandra:
* Added reference tag to the top
* Added a couple clarifying links (probably unnecessary)
* Fixed one link that I broke (whoops!)
* Restructured some lists to make the steps clearer


ejeschke:
Thank you!

PR #484, opened 2017-05-16
Pick save table auto-detect extension

### LINKED ISSUE #479: Feature Request: Allow Pick to save out as ASCII table
Someone asked me why `Pick` only allows saving out to FITS table, not ASCII. The use case is that user wants to use the output with `tweakreg` (part of [DrizzlePac](http://drizzlepac.stsci.edu/)), which only accepts ASCII table.

pllim:
Fix #479 

p.s. The "comment" is only saved out as FITS header keyword but not as a comment line in ASCII. See astropy/astropy#6079

### DIFF
--- a/doc/manual/plugins_local/pick.rst
+++ b/doc/manual/plugins_local/pick.rst
@@ -86,6 +86,7 @@ The blue lines show measurements in the X direction and the green lines
 show measurements in the Y direction.  The solid lines indicate actual
 pixel values and the dotted lines indicate the fitted 1D gaussians.
 The shaded green and blue regions indicate the FWHM measurements.
+(Moffat fitting is also available.)
 
 The "Radial" tab contains a radial profile plot:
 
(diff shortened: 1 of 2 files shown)


ejeschke:
@pllim, do you know what minimum astropy version supports the change in this PR?

pllim:
It seems like you already use "overwrite" keyword. I think that's 1.2 or 1.3 (hard to check on phone). So the changes here are compatible with Astropy that you already use.

pllim:
I notice that your Astropy requirement is 0.3. That is a very old version. Is it okay to bump the version up to 1.x?

ejeschke:
> It seems like you already use "overwrite" keyword. I think that's 1.2 or 1.3 (hard to check on phone). So the changes here are compatible with Astropy that you already use.

I just tested with python3 and astropy 1.1.1.  `IORegistryError` doesn't seem to be defined in that version at least, but the `overwrite` keyword is supported.

Could we just do something like:
```python
format = None
if filepath.lower().endswith('.txt'):
    format = 'ascii.commented_header'
tbl.write(filepath, overwrite=True, format=format)
```
?

pllim:
@ejeschke , okay, I implemented your suggestion. Also fixed the syntax as addressed by astropy/astropy#6097 and bumping Astropy requirement to v1.x.

setup.py
```diff
@@ -105,7 +105,7 @@
 # https://groups.google.com/forum/#!topic/astropy-dev/urYO8ckB2uM
 
 # pretty much needed
-install_requires = ['astropy>=0.3', 'numpy>=1.7', 'qtpy>=1.1',
```
ejeschke: @pllim, can we treat this change as a separate PR?  In general, I am not opposed, but there are some considerations which are unrelated to this PR and I'd rather treat it as a separate issue.

pllim: Okay. Removed.

pllim:
Removed Astropy bump. Please review.

ejeschke:
Looks good.

PR #485, opened 2017-05-17
Fix catalog GUI

pllim:
While testing spacetelescope/stginga#136 , I found that `Catalog` user interface is broken here and there, so I fixed it up. Also some PEP8 clean-ups.

p.s. I only tested this with a star catalog. I did not use the image server because I don't know what it is for.

### DIFF
--- a/ginga/rv/plugins/Catalogs.py
+++ b/ginga/rv/plugins/Catalogs.py
@@ -35,6 +35,14 @@ class Catalogs(GingaPlugin.LocalPlugin):
     def __init__(self, fv, fitsimage):
         super(Catalogs, self).__init__(fv, fitsimage)
 
+        prefs = self.fv.get_preferences()
+        self.settings = prefs.create_category('plugin_Catalogs')
+        self.settings.add_defaults(draw_type='circle',
+                                   select_color='skyblue',
+                                   color_outline='aquamarine',
+                                   click_radius=10)
+        self.settings.load(onError='silent')
+
         self.limit_stars_to_area = False
         self.pan_to_selected = False
         self.use_dss_channel = False
@@ -42,22 +50,16 @@ class Catalogs(GingaPlugin.LocalPlugin):
         self.plot_max = 500
         self.plot_limit = 100
         self.plot_start = 0
-        self.drawtype = 'circle'
+        self.drawtype = self.settings.get('draw_type', 'circle')
 
         # star list
         self.starlist = []
         # catalog listing
         self.table = None
 
-        self.color_outline = 'aquamarine'
+        self.color_outline = self.settings.get('color_outline', 'aquamarine')
         self.layertag = 'catalog-canvas'
         self.areatag = None
-
-        prefs = self.fv.get_preferences()
-        self.settings = prefs.create_category('plugin_Catalogs')
-        self.settings.add_defaults(select_color='skyblue',
-                                   click_radius=10)
-        self.settings.load(onError='silent')
 
         self.image_server_options = []
         self.image_server_params = None
@@ -69,9 +71,9 @@ class Catalogs(GingaPlugin.LocalPlugin):
         canvas = self.dc.DrawingCanvas()
         canvas.enable_draw(True)
         canvas.set_drawtype(self.drawtype, color='cyan', linestyle='dash')
-        canvas.set_callback('cursor-down', self.btndown)
-        canvas.set_callback('cursor-up', self.btnup)
         canvas.set_callback('draw-event', self.draw_cb)
+        canvas.add_draw_mode('select', down=self.btndown,
+                             move=self.btndown, up=self.btnup)
         canvas.register_for_cursor_drawing(self.fitsimage)
         canvas.set_surface(self.fitsimage)
         canvas.set_draw_mode('draw')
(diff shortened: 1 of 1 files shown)


ginga/rv/plugins/Catalogs.py
```diff
@@ -372,26 +401,38 @@ def redo(self):
...
         return True
 
-    def btndown(self, canvas, event, data_x, data_y):
+    def btndown(self, canvas, event, data_x, data_y, viewer):
         return True
 
-    def btnup(self, canvas, event, data_x, data_y):
-
-        objs = self.canvas.get_items_at(data_x, data_y)
+    def btnup(self, canvas, event, data_x, data_y, viewer):
```
ejeschke: This is rather a pronounced change which could possibly impact performance.  Before, the canvas is queried at the point of clicking to determine which objects are there.  The number of items returned is likely to be small.  But in the PR, the number of items returned may be large (e.g. in our case hundreds of stars circled) and then each one is tested for inclusion of the point of clicking.

Any particular reason for this change?

ginga/rv/plugins/Catalogs.py
```diff
@@ -1196,11 +1236,15 @@ def _unselect_tv(self, star, fromtable=False):
     def select_star_cb(self, widget, res_dict):
         """This method is called when the user selects a star from the table.
         """
-        key = list(res_dict.keys())[0]
-        idx = int(key)
-        star = self.starlist[idx]
-        if not self._select_flag:
-            self.mark_selection(star, fromtable=True)
```
ejeschke: I would suggest `keys` instead of `key`, since we are making a list here and not just trying to select the first one.

ginga/rv/plugins/Catalogs.py
pllim: Does this work for you? Because `get_items_at` always throws the following error at me (Qt5 on RHEL7):
```
reduce() of empty sequence with no initial value
```

ejeschke: Hmm, still looks like getting all objects with tag="star" and then iterating through them to see whether they contain the point.  

I'll check on the error you are getting for `get_items_at`--- it sounds like it should be easily fixed.

pllim: Okay, if you can fix that, then I can revert this change. That would be more elegant indeed. Thanks!

ejeschke: > reduce() of empty sequence with no initial value

@pllim, can you specify exactly what you are doing to get this error?

pllim: 1. Open `s1_mips_v0.30_sci.fits` (I put it on Box for you).
2. Open `Catalog` with [this ginga_config.py](https://github.com/spacetelescope/stginga/blob/master/stginga/examples/configs/ginga_config.py).
3. Select `GSC_1.2` and search.
4. Left click on any of the circled stars on the image in an attempt to highlight the corresponding table entry.

I see this (without this PR):
```
2017-05-19 16:57:14,716 | E | Callback.py:139 (make_callback) | Error making callback 'cursor-up': reduce() of empty sequence with no initial value
2017-05-19 16:57:14,717 | E | Callback.py:140 (make_callback) | Traceback:

  File ".../ginga/rv/plugins/Catalogs.py", line 387, in btnup
    objs = self.canvas.get_items_at(data_x, data_y)

  File ".../ginga/canvas/CompoundMixin.py", line 67, in get_items_at
    if obj.contains(x, y):

  File ".../ginga/canvas/CompoundMixin.py", line 61, in contains
    res = self.contains_arr(x_arr, y_arr)

  File ".../ginga/canvas/CompoundMixin.py", line 57, in contains_arr
    self.objects))
```

ejeschke: > can you specify exactly what you are doing to get this error?

Never mind, I was able to reproduce it under python 3.

ejeschke:
> I did not use the image server because I don't know what it is for.

An image server serves up images (e.g. DSS or SDSS) instead of tables of data.

pllim:
> An image server serves up images

Sure, but what does Ginga do with them? Do you simply overlay the image footprints, or download and display them? What are the use cases for this feature?

pllim:
Closing this one due to conflicts.

ejeschke:
> Sure, but what does Ginga do with them? Do you simply overlay the image footprints, or download and display them? What are the use cases for this feature?

We use it for guide star selection during operation.  When acquiring a field with an instrument that uses guiding, the coordinates of the target are fed into the plugin, along with a radius defined by the FOV.  We then fetch a DSS image (from one of several sources, depending on the instrument) and display it in a "DSS" channel.  We then use the catalog server feature to circle all the guide stars and allow the user to pick one (in manual mode) or pick the top ranked one (in automatic mode).  The guide stars are prioritized according to magnitude, position in the FOV and various catalog attributes.  The detector footprint and the telescope vignetting are also shown by ginga graphics on the DSS image.

pllim:
Ah, okay. So you actually download the image from image catalog. I can't think of any use case for that right now at STScI, as observatory control software is not something that I do. Thanks for the clarification!

PR #489, opened 2017-05-19
Handle cancellation of file save

### LINKED ISSUE #476: SaveDialog Qt5 connect warning and error on cancel
Qt5 on RHEL7.

1. Open an image.
2. Open Cuts plugin.
3. Draw a cut to enable Save button.
4. Click on Save button.

When button is clicked, I see this on the terminal:
```
QObject::connect: invalid null parameter
```

Then I click Cancel without saving the file and I see this:
```python
Error making callback 'activated': can only concatenate tuple (not "str") to tuple
Traceback:

  File ".../ginga/rv/plugins/Cuts.py", line 210, in <lambda>
    lambda w: self.save_cb(mode='cuts'))

  File ".../ginga/rv/plugins/Cuts.py", line 1015, in save_cb
    self.fv.error_wrap(image.save_as_file, target + '.fits')
```

The problem with the second traceback is that this is no longer a valid check:
```python
        # Save cancelled
        if not target:
            return
```

pllim:
Fix #476

### DIFF
--- a/ginga/qtw/Widgets.py
+++ b/ginga/qtw/Widgets.py
@@ -1815,9 +1815,22 @@ class SaveDialog(QtGui.QFileDialog):
         self.widget = self.getSaveFileName(self, title, '', selectedfilter)
 
     def get_path(self):
-        if (self.widget and self.selectedfilter is not None and
-                not self.widget.endswith(self.selectedfilter[1:])):
-            self.widget += self.selectedfilter[1:]
+        if isinstance(self.widget, tuple):
+            w = self.widget[0]
+        else:
+            w = self.widget
+
+        if (w and self.selectedfilter is not None and
+                not w.endswith(self.selectedfilter[1:])):
+            w += self.selectedfilter[1:]
+
+        if isinstance(self.widget, tuple):
+            wlist = [x for x in self.widget]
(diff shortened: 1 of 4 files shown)


pllim:
@ejeschke , in addition to your fix, there is an additional piece here to fix error that only occur if filter is provided. Please review.

ginga/qtw/Widgets.py
```diff
@@ -1815,9 +1815,22 @@ def __init__(self, title=None, selectedfilter=None):
...
+            w = self.widget[0]
+        else:
+            w = self.widget
+
+        if (w and self.selectedfilter is not None and
+                not w.endswith(self.selectedfilter[1:])):
+            w += self.selectedfilter[1:]
+
+        if isinstance(self.widget, tuple):
+            wlist = [x for x in self.widget]
```
pllim: Haha... yes, I could... I'll test your changes next week. Thank you and have a great weekend!

pllim:
> can you test latest master and see if this PR is still necessary? 

I still see traceback if I hit Cancel when a filter is given but I'll submit another PR due to conflicts.

pllim:
See #490

PR #490, opened 2017-05-20
Fix error in Savedialog with filter, AstroTable naxispath

### LINKED ISSUE #476: SaveDialog Qt5 connect warning and error on cancel

pllim:
Follow-up of 797e508ac . Supersede #489. Fix #476.

Also follow-up of #461.

### DIFF
--- a/ginga/qtw/Widgets.py
+++ b/ginga/qtw/Widgets.py
@@ -1818,16 +1818,17 @@ class SaveDialog(QtGui.QFileDialog):
 
     def get_path(self):
         res = self.getSaveFileName(self, self.title, '', self.selectedfilter)
-        if (res and self.selectedfilter is not None and
-                not res.endswith(self.selectedfilter[1:])):
-            res += self.selectedfilter[1:]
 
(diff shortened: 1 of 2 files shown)


ejeschke:
This looks good to me.

Issue #483, opened 2017-05-15
Feature Request: Use Astroquery for catalog queries


pllim:
Both Ginga and Astroquery are affilliated packages, so this makes sense.

Available services: http://astroquery.readthedocs.io/en/latest/#available-services

c/c @keflavich @bsipocz

ejeschke:
I think Ginga can/does use `astropy.vo.client.conesearch` for catalog searches.

pllim:
@ejeschke , I moved Cone Search to `astroquery` recently. It will be deprecated in Astropy 2.0 and removed in Astropy 3.0. FYI.

pllim:
We have example using `astroquery` now in

https://github.com/ejeschke/ginga/blob/8e51fd51ef2bece03f9e79b1191b19008c9cd3ad/ginga/rv/plugins/Catalogs.py#L18

and implementations in 

https://github.com/ejeschke/ginga/blob/8e51fd51ef2bece03f9e79b1191b19008c9cd3ad/ginga/util/catalog.py#L203

and 

https://github.com/ejeschke/ginga/blob/8e51fd51ef2bece03f9e79b1191b19008c9cd3ad/ginga/util/catalog.py#L67-L69

PR #491, opened 2017-05-20
Fix Catalog plugin GUI

pllim:
Fix up `Catalog` behavior. Add user preference. PEP8 clean-ups.

Supersede #485.

### DIFF
--- a/ginga/rv/plugins/Catalogs.py
+++ b/ginga/rv/plugins/Catalogs.py
@@ -72,8 +72,7 @@ class Catalogs(GingaPlugin.LocalPlugin):
         canvas.enable_draw(True)
         canvas.set_drawtype(self.drawtype, color='cyan', linestyle='dash')
         canvas.set_callback('draw-event', self.draw_cb)
-        canvas.add_draw_mode('select', down=self.btndown,
-                             move=self.btndown, up=self.btnup)
+        canvas.add_draw_mode('select', down=None, move=None, up=self.btnup)
         canvas.register_for_cursor_drawing(self.fitsimage)
         canvas.set_surface(self.fitsimage)
         canvas.set_draw_mode('draw')
@@ -407,13 +406,11 @@ class Catalogs(GingaPlugin.LocalPlugin):
         self._update_widgets(d)
         return True
 
-    def btndown(self, canvas, event, data_x, data_y, viewer):
-        return True
-
     def btnup(self, canvas, event, data_x, data_y, viewer):
         try:
             objs = self.canvas.get_items_at(data_x, data_y)
-        except Exception:
+        except Exception as e:
+            self.logger.warning('select failed: {}'.format(str(e)))
             return True
         for obj in objs:
             if (obj.tag is not None) and obj.tag.startswith('star'):
--- a/ginga/rv/plugins/Catalogs.py
+++ b/ginga/rv/plugins/Catalogs.py
@@ -13,7 +13,7 @@ from ginga.misc import Bunch
 from ginga import GingaPlugin
 from ginga import cmap, imap
 from ginga.util import wcs
-from ginga.util.six.moves import map, zip
+from ginga.util.six.moves import map
 from ginga.gw import ColorBar, Widgets
 
 
@@ -35,6 +35,14 @@ class Catalogs(GingaPlugin.LocalPlugin):
     def __init__(self, fv, fitsimage):
         super(Catalogs, self).__init__(fv, fitsimage)
 
+        prefs = self.fv.get_preferences()
+        self.settings = prefs.create_category('plugin_Catalogs')
+        self.settings.add_defaults(draw_type='circle',
+                                   select_color='skyblue',
+                                   color_outline='aquamarine',
+                                   click_radius=10)
+        self.settings.load(onError='silent')
+
         self.limit_stars_to_area = False
         self.pan_to_selected = False
         self.use_dss_channel = False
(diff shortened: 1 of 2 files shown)


pllim:
@ejeschke , please review.

ginga/rv/plugins/Catalogs.py
```diff
@@ -372,26 +401,38 @@ def redo(self):
...
+    def btndown(self, canvas, event, data_x, data_y, viewer):
         return True
 
-    def btnup(self, canvas, event, data_x, data_y):
-
-        objs = self.canvas.get_items_at(data_x, data_y)
+    def btnup(self, canvas, event, data_x, data_y, viewer):
+        try:
+            objs = self.canvas.get_items_at(data_x, data_y)
+        except Exception:
```
ejeschke: Should we log an error here?

ginga/rv/plugins/Catalogs.py
```diff
@@ -67,13 +71,15 @@ def __init__(self, fv, fitsimage):
         canvas = self.dc.DrawingCanvas()
         canvas.enable_draw(True)
         canvas.set_drawtype(self.drawtype, color='cyan', linestyle='dash')
-        canvas.set_callback('cursor-down', self.btndown)
-        canvas.set_callback('cursor-up', self.btnup)
         canvas.set_callback('draw-event', self.draw_cb)
+        canvas.add_draw_mode('select', down=self.btndown,
+                             move=self.btndown, up=self.btnup)
```
ejeschke: I wonder if the `move` element should be `None` (or omitted).  I don't think it makes sense to have a binding for that in the case of selection.

ejeschke: For that matter, it doesn't look like `btndown` is doing anything either, so maybe `down` can also be set to `None` or omitted.

ejeschke: @pllim, we might be able to take advantage of a feature that was added not too long ago.  Canvases have a "pick" mode, in addition to "draw" and "edit".  You might be able to use that one instead of defining a new "select" mode.  Just provide checkboxes for "draw" and "pick" (or you can call it "select", but make it put the canvas in "pick" mode).

There are `pick` callbacks which are defined for any canvas object.  In the place where we are adding the object with the "star"-prefixed tag, we can simply register for the "pick-up" callback on the object.  The callback function function signature would look like `(canvas, event, pt)`, where `pt` is a tuple of the point.  In registering for the callback, we can add the object itself (and it's star info) so that the callback looks like: `(canvas, event, pt, obj, starinfo)`.  Then there is no need to check the tag type or extract the info.

ejeschke: We could also do that as a separate enhancement.

pllim: I did see the "pick" callback in your code but I wasn't sure what it is for, so thanks for the clarifications. I'll work on this a bit more.

pllim: Actually, I spoke too soon. I cannot find any actual implementation in other `rv` plugins that use this "pick" callback. Maybe it is best I leave that to you *after* this PR is merged. I'll remove `btndown` as you suggested.

ginga/rv/plugins/Catalogs.py
pllim: I don't think so? If it fails for any reason, the table won't be highlighted, but that is about it. Nothing critical.

ginga/rv/plugins/Catalogs.py
ejeschke: > Maybe it is best I leave that to you after this PR is merged

That sounds good, as it probably has not been well tested.  But basically it generalizes what has been done in the `btnup` callback.  There are some other interesting callbacks, like 'pick-hover', so you could do things like pop up a text box with the star name and information as you move the cursor over an object, etc.

ginga/rv/plugins/Catalogs.py
ejeschke: Seems like we would want to know about any error here though.  As far as I know there shouldn't be any, so if there is an exception it would be good to know what it is.  Maybe a warning, at least?

pllim: Okay, I'll add a warning.

pllim:
I cleaned up `util/catalog.py` and switched the `conesearch` import. Adding more Astroquery sources can be part of #483 (future work).

ginga/util/catalog.py
```diff
@@ -4,35 +4,34 @@
...
-have_astropy = False
+# Do we have astroquery >=0.3.5 installed?
+have_astroquery = False
 try:
-    from astropy.vo.client import conesearch
     from astropy import coordinates, units
-    have_astropy = True
-
+    from astroquery.vo_conesearch import conesearch
+    have_astroquery = True
```
pllim: Not sure what this is for. It is not used anywhere in the module.

pllim:
I think I addressed all your comments. Please review again. Thanks!

ginga/util/catalog.py
```diff
@@ -4,35 +4,34 @@
...
 except ImportError:
```
ejeschke: By adding here:
```python
    from astropy.vo.client import conesearch
```
we can support older versions of `astropy`.  I think it is too early to require v2.0.

pllim:
> some odd messages

The warnings are because the VO table downloaded violated certain standards. If you don't care about them, I can add in a context manager to suppress all warnings from Cone Search.

As for the timer message, it is because I switched from manual timing to a built-in Cone Search function that returns both the run time and results. It is just info message, so should not do any harm.

pllim:
@ejeschke , I think I have addressed all your comments. Please review (again).

ejeschke:
Looks and works good.  Merging.

ejeschke:
@pllim, I made a few minor updates to `Catalogs` after merging.  First, I added an "edit" mode, where the user can move and resize the search area object.  For repeated searches, some people may prefer this to constantly redrawing the area.  Second, I changed the plugin to use the "pick" mode of canvases to implement the "select".

pllim:
@ejeschke , works great. Thank you! 👍

Issue #494, opened 2017-05-24
WCS conversion does not support multiple pixels at once (except for AstropyWCS)


pllim:
In `wcsmod.py`, I found out that `pixtoradec()` method (at least for `astropy.wcs`) only works when single pixel `[X, Y]` coordinate is given. When I give `[[X1, Y1], [X2, Y2], ...]`, it breaks even though `astropy.wcs.WCS.all_pix2world()` can calculate for multiple pixels at the same time.

Is this a feature or a bug?

In an attempt to address #391 , the ability to calculate multiple locations at once is very desirable. To work around this, I need to directly access `all_pix2world()`, which limits the code to `astropy.wcs`, not any WCS package that user chooses to use.

ejeschke:
> Is this a feature or a bug?

Neither, but just an omission in the initial design.  It would not be hard to support for all WCSes, since we could just do a cython iteration or some such for wcs packages that don't support array operations.  But I think that `kapteyn` does also support array operations, and maybe also `starlink-ast`.

It's been on my TODO list for a while, because it would also be very nice for the canvas graphics that use WCS coordinates to plot--because it could translate all the points at once.

If we do a PR like this we should also change the API so that the coordinates are passed in such that multi-dimensional points are supported; i.e. it is not `(x, y)`, but `(points)`.

ejeschke:
I would suggest that if we undertake this PR, that we introduce a new call in the API for the `pixtoradec` replacement (also because you are not always converting to RA/DEC!)--this will minimize breakage of many things that rely on the current API.

pllim:
The best I can do is implement that API for AstropyWCS and make the rest raise NotImplementedError, as I have never used those.

ejeschke:
I'll help with the other ones.

ejeschke:
I think we can close this, right @pllim?  API has been implemented for all wrappers, and 3 of them are truly vectorized (in the `numpy` sense).

pllim:
Yes, thanks!

PR #495, opened 2017-05-25
Add new WCSAxes local plugin

### LINKED ISSUE #391: Display RA and DEC lines on display
As discussed during JWST Data Analysis Workshop II on Nov 9, 2016. We can have a local (or global?) plugin that can toggle lines connecting RA and DEC tickmarks across the image, accounting for any WCS distortion. Maybe also allow user control; e.g., how finely to space those lines, line style, line color, or line thickness.

pllim:
Fix #391

TODO:
- [x] Support data cube
- [x] Address #494 (only implemented for `AstropyWCS` for now)
- [x] Add RST doc with screenshot.


ejeschke:
@pllim, is this to draw axes over the image?  I strongly urge you to consider adding it as a new canvas object in `ginga.canvas.types.astro`.  That will be more generally useful for any application of the widget.

Of course, we can always have a plugin to configure the parameters of the axes.

pllim:
Yes, it draw axes grid in WCS space. As-is it can already draw on 2D image with WCS in header, you can see for yourself if you want. I'll check out the canvas tomorrow.

ejeschke:
I realize it's WIP, but just mentioning...  In testing, flipping either X or Y axis was ok, but swapping axes produced a situation in which the angle of the text is orthogonal to the axes.  Also, rotating the image results in the text not aligning correctly.

ejeschke:
> I'll check out the canvas tomorrow

Based on your current implementation (using compound object) something like `Annulus` might be a good example approach, especially if there is no need to draw in canvas coordinates.  Since this kind of object is not really user-drawable, there is no need to implement the `idraw()` method, nor any editing features.  You probably need to override the `draw()` method to check for changes in certain state (e.g. rotation, transform or change of image) and update the local objects accordingly IFF they have changed since the last draw--this will allow your axes to follow swaps, rotations and change of image.

ejeschke:
Ok, so I was eager to see how this could work. :smile_cat:  Here is a working implementation derived from your code:
```python
from ginga.util.wcs import raDegToString, decDegToString
from ginga.util.wcsmod import AstropyWCS

class WCSAxes(CompoundObject):
    """Special compound object to draw WCS axes.

    """
    @classmethod
    def get_params_metadata(cls):
        return [
            Param(name='linewidth', type=int, default=1,
                  min=1, max=20, widget='spinbutton', incr=1,
                  description="Width of outline"),
            Param(name='linestyle', type=str, default='dash',
                  valid=['solid', 'dash'],
                  description="Style of outline (default dash)"),
            Param(name='color',
                  valid=colors_plus_none, type=_color, default='yellow',
                  description="Color of grid and text"),
            Param(name='alpha', type=float, default=1.0,
                  min=0.0, max=1.0, widget='spinfloat', incr=0.05,
                  description="Opacity of grid and text"),
            Param(name='font', type=str, default='Sans Serif',
                  description="Font family for text"),
            Param(name='fontsize', type=int, default=8,
                  min=8, max=72,
                  description="Font size of text (default: 8)"),
            ]

    def __init__(self, color='cyan',
                 linewidth=1, linestyle='dash', alpha=1.0,
                 font='Sans Serif', fontsize=8,
                 **kwdargs):

        # these could become supplied optional parameters, if desired
        self.show_label = True
        self.num_ra = 10
        self.num_dec = 10
        self._pix_res = 10
        self.txt_off = 4
        # for keeping track of changes to image and orientation
        self._cur_rot = None
        self._cur_swap = None
        self._cur_image = None

        CompoundObject.__init__(self,
                                color=color, alpha=alpha,
                                linewidth=linewidth, linestyle=linestyle,
                                font=font, fontsize=fontsize, **kwdargs)

        self.editable = False
        self.pickable = False
        self.opaque = True
        self.kind = 'wcsaxes'

    def _calc_axes(self, viewer, image, rot_deg, swapxy):
        self._cur_image = image
        self._cur_rot = rot_deg
        self._cur_swap = swapxy

        if not isinstance(image, AstroImage) or not image.has_valid_wcs():
            return []

        # TODO: Support all WCS packages
        if not isinstance(image.wcs, AstropyWCS):
            return []

        # TODO: Support data cube
        # Approximate bounding box in RA/DEC space
        xmax = image.width - 1
        ymax = image.height - 1
        radec = image.wcs.wcs.all_pix2world(
            [[0, 0], [0, ymax], [xmax, 0], [xmax, ymax]], 0)
        ra_min, dec_min = radec.min(axis=0)
        ra_max, dec_max = radec.max(axis=0)
        ra_size = ra_max - ra_min
        dec_size = dec_max - dec_min

        # Calculate positions of RA/DEC lines
        d_ra = ra_size / (self.num_ra + 1)
        d_dec = dec_size / (self.num_dec + 1)
        ra_arr = numpy.arange(ra_min + d_ra, ra_max - d_ra * 0.5, d_ra)
        dec_arr = numpy.arange(dec_min + d_dec, dec_max - d_ra * 0.5, d_dec)

        # RA/DEC step size for each vector
        min_imsize = min(image.width, image.height)
        d_ra_step = ra_size * self._pix_res / min_imsize
        d_dec_step = dec_size * self._pix_res / min_imsize

        # Create Path objects
        objs = []

        for cur_ra in ra_arr:
            crds = [[cur_ra, cur_dec] for cur_dec in
                    numpy.arange(dec_min, dec_max + d_dec_step, d_dec_step)]
            lbl = raDegToString(cur_ra)
            objs += self._get_path(viewer, image, crds, lbl, 1)
        for cur_dec in dec_arr:
            crds = [[cur_ra, cur_dec] for cur_ra in
                    numpy.arange(ra_min, ra_max + d_ra_step, d_ra_step)]
            lbl = decDegToString(cur_dec)
            objs += self._get_path(viewer, image, crds, lbl, 0)

        return objs

    def _get_path(self, viewer, image, crds, lbl, axis):
        pts = image.wcs.wcs.all_world2pix(crds, 0)

        from ginga.canvas.types.basic import Path, Text

        # Don't draw outside image area
        mask = ((pts[:, 0] >= 0) & (pts[:, 0] < image.width) &
                (pts[:, 1] >= 0) & (pts[:, 1] < image.height))
        pts = pts[mask]

        path_obj = Path(
            points=pts, coords='data', linewidth=self.linewidth,
            linestyle=self.linestyle, color=self.color,
            alpha=self.alpha)
        # this is necessary because we are not actually adding to a canvas
        path_obj.crdmap = viewer.get_coordmap('data')

        if self.show_label:
            # Calculate label orientation
            x1, y1 = pts[0]
            x2, y2 = pts[-1]

            if axis == 0:
                # x axis varying
                rot = math.asin((y2 - y1) / (x2 - x1)) * 180 / math.pi
                rot = self._cur_rot + rot

                x, y = abs(x2 - x1) / 2 + self.txt_off, y2
            else:
                # y axis varying
                rot = self._cur_rot + 90

                x, y = x2, abs(y2 - y1) / 2 + self.txt_off

            if self._cur_swap:
                # axes are swapped
                rot -= 90

            text_obj = Text(x, y, text=lbl,
                            fontsize=self.fontsize, color=self.color,
                            alpha=self.alpha, rot_deg=rot, coord='data')
            # this is necessary because we are not actually adding to a canvas
            text_obj.crdmap = viewer.get_coordmap('data')

            return [path_obj, text_obj]

        else:
            return [path_obj]

    def draw(self, viewer):
        # see if we need to recalculate our grid
        image = viewer.get_image()
        update = False
        if self._cur_image != image:
            # new image loaded
            update = True

        cur_swap = viewer.get_transforms()[2]
        if cur_swap != self._cur_swap:
            # axes have been swapped
            update = True

        cur_rot = viewer.get_rotation()
        if cur_rot != self._cur_rot:
            # rotation has changed
            # TODO: for a rotation or swap axes change, it would be
            # sufficient to simply calculate the new rotation angles
            # and update all the text objects in self.objects
            update = True

        if len(self.objects) == 0:
            # initial time
            update = True

        if update:
            # only expensive recalculation of grid if needed
            self.objects = self._calc_axes(viewer, image, cur_rot, cur_swap)

        super(WCSAxes, self).draw(viewer)


register_canvas_types(dict(ruler=Ruler, compass=Compass,
                           crosshair=Crosshair, annulus=Annulus,
                           wcsaxes=WCSAxes))
```

If you add this to the bottom of `ginga.canvas.types.astro` then you can lookup and add an object of type "wcsaxes" to any ginga canvas.  It also correctly handles rotation and a transform/swapped axes.

ejeschke:
The `get_params_metadata` class method is probably not necessary, but might be useful for constructing a GUI to alter the axes parameters.

pllim:
Wow, thanks! I'll update the PR accordingly. :smile:

pllim:
@ejeschke , I am trying to incorporate your new canvas class into the plugin. How do I pass `image` into `self.dc.WCSAxes`?

ejeschke:
> How do I pass image into self.dc.WCSAxes

You don't.  Just simply:
```python
self.axes = self.dc.WCSAxes()
self.canvas.add(self.axes)
```
The object will figure out whatever image is behind the viewer and draw accordingly.

ejeschke:
> The object will figure out whatever image is behind the viewer and draw accordingly.

If you want the axes to be visible without the plugin being active, we can simply add a control for it in the `Toolbar` plugin or even a keystroke binding.  There are already a few predecessor examples like the onscreen color bar, mode indicator, pan mark.  Probably `Crosshair` ought to become one of those as well.

pllim:
Thanks!

> axes to be visible without the plugin

Maybe in the future if there is demand, but not right now. Often times, people want the overlay for publication, so there is a desire to fine tune the display parameters (color, font size, etc) until it looks nice. In that case, a plugin with fine controls is better suited.

pllim:
The label positions and angles are weird for my image `jb5g05ubq_flt.fits`, which I added to Box for you to download if you want to test with it:
untitled

pllim:
I just pushed another WIP commit, you can test your canvas code with the plugin. FYI.

pllim:
p.p.s. If you want to change some code here, you can either open a PR on my fork against my PR branch, or just push directly to my PR branch as a maintainer.

ginga/rv/plugins/WCSAxes.py
```diff
@@ -0,0 +1,305 @@
...
+        else:
+            if obj.kind != 'wcsaxes':
+                return True
+        try:
+            self.canvas.delete_object_by_tag(self.overlaytag)
+        except Exception:
+            pass
+
+        # dc.WCSAxes will pick up image under the hood.
+        # UNTIL HERE - what about other params to adjust?
```
ejeschke: @pllim, I would add the `WCSAxes` object right after you create the new canvas in the constructor.  No need to create a new one for every `redo()`.

ginga/rv/plugins/WCSAxes.py
```diff
@@ -0,0 +1,305 @@
...
+        obj = self.dc.WCSAxes(
+            linewidth=self.linewidth, linestyle=self.linestyle,
+            color=self.linecolor, alpha=self.alpha, fontsize=self.fontsize)
+        self.overlaytag = self.canvas.add(obj)
+        self.canvas.redraw(whence=3)
+
+    def set_linecolor_cb(self, w, index):
```
ejeschke:
```python
self.axes.color = self.colornames[index]
self.canvas.update_canvas()
```

ginga/rv/plugins/WCSAxes.py
```diff
@@ -0,0 +1,305 @@
...
+        self.redo()
+        return True
+
+    def set_alpha(self):
+        try:
+            a = float(self.w.alpha.get_text())
+        except ValueError:
+            self.w.alpha.set_text(str(self.alpha))
+        else:
+            self.alpha = a
```
ejeschke: Similarly,
```python
self.axes.alpha = alpha
self.canvas.update_canvas()
```

ginga/rv/plugins/WCSAxes.py
```diff
@@ -0,0 +1,305 @@
...
+        try:
+            p_canvas.get_object_by_tag(self.layertag)
+        except KeyError:
+            # Add ruler layer
+            p_canvas.add(self.canvas, tag=self.layertag)
+
+        self.resume()
+
+    def pause(self):
+        self.canvas.ui_set_active(False)
```
ejeschke: I don't think there is any drawing or UI on the canvas, so probably no need for `pause()` and `resume()` methods.

ginga/rv/plugins/WCSAxes.py
```diff
@@ -0,0 +1,305 @@
...
+
+    def resume(self):
+        # turn off any mode user may be in
+        self.modes_off()
+
+        self.canvas.ui_set_active(True)
+        self.fv.show_status("Overlaying WCS axes if available")
+
+    def stop(self):
```
ejeschke:
```python
# so we don't hang on to a large image
self.axes._cur_image = False
```

ginga/rv/plugins/WCSAxes.py
```diff
@@ -0,0 +1,305 @@
...
+        self.ra_angle = self.settings.get('ra_angle', None)
+        self.dec_angle = self.settings.get('dec_angle', None)
+        self._pix_res = 10
+
+        self.dc = fv.get_draw_classes()
+
+        canvas = self.dc.DrawingCanvas()
+        canvas.enable_draw(False)
+        canvas.set_surface(self.fitsimage)
+        self.canvas = canvas
```
ejeschke:
```python
self.axes = self.dc.WCSAxes(color=self.linecolor, alpha=self.alpha, linestyle=self.linestyle, linewidth=self.linewidth,
                                                fontsize=self.fontsize)
# should these be added as constructor parameters of a WCSAxes object?
self.axes.num_ra = self.num_ra
self.axes.num_dec = self.num_dec
self.axes.show_label = self.show_label
self.axes 
self.canvas.add(self.axes)
```
etc.

ginga/rv/plugins/WCSAxes.py
ejeschke: In fact, `redo()` should not really be necessary at all, if the `WCSAxes` canvas object is working correctly.

ginga/rv/plugins/WCSAxes.py
pllim: This does not seem to change the axes displayed.

ejeschke: > This does not seem to change the axes displayed.

It doesn't change the color if you pick a new one?  Try an 
```python
self.fitsimage.redraw(whence=3)
```
instead of the `update_canvas()` call.  (Although the latter ought to work, and needs to be investigated if it doesn't).

pllim: That does not work either but when I implement a `WCSAxes.sync_state()` (see below) and followed by `self.canvas.update_canvas()`, then it works. However, that means for certain parameter change (like, `num_ra`), I need to find a way to force complete recomputation.
```python
    def sync_state(self):
        for obj in self.objects:
            obj.color = self.color

            if obj.kind == 'text':
                obj.fontsize = self.fontsize
                if self.show_label:
                    obj.alpha = self.alpha
                else:
                    obj.alpha = 0  # hide

            else:  # path
                obj.alpha = self.alpha
                obj.linewidth = self.linewidth
                obj.linestyle = self.linestyle
```

pllim: This seems to work to force redraw:
```python
            self.axes.num_ra = self.num_ra
            self.axes._cur_image = None  # Force redraw
            self.axes.draw(self.channel.viewer)
            self.canvas.update_canvas()
```

ejeschke: > but when I implement a WCSAxes.sync_state() (see below) and followed by self.canvas.update_canvas(), then it works

Ah, right. I forgot that this is should be similar to the `Annulus` in that regard.  In that case, what do you think of making the `num_ra`, etc properties instead of regular instance variables, in which case you can just call `_calc_axes()` "under the hood" when one of the important properties changes?

ejeschke: > This seems to work to force redraw

year, setting the image to None might be the easiest way to force the recalculation.  But I don't think you should need the call to `self.axes.draw(self.channel.viewer)`.  That should automatically happen with you call `self.canvas.update_canvas()`.  Can you confirm?

pllim: I just pushed a commit with your feedback but I can't test it till tomorrow. Thanks!

ginga/rv/plugins/WCSAxes.py
```diff
@@ -0,0 +1,323 @@
...
+            p_canvas.get_object_by_tag(self.layertag)
+        except KeyError:
+            # Add ruler layer
+            p_canvas.add(self.canvas, tag=self.layertag)
+
+        self.resume()
+
+    def stop(self):
+        # so we don't hang on to a large image
+        self.axes._cur_image = False
```
pllim: Note to self: Need to change this to None

pllim:
@ejeschke , I think the GUI itself is working pretty well but I can't figure out how to fix https://github.com/ejeschke/ginga/pull/495#issuecomment-304362236

ginga/canvas/types/astro.py
```diff
@@ -666,7 +679,238 @@ def move_to(self, xdst, ydst):
...
+        if not isinstance(image, AstroImage) or not image.has_valid_wcs():
+            return []
+
+        # Approximate bounding box in RA/DEC space
+        xmax = image.width - 1
+        ymax = image.height - 1
+        try:
+            radec = image.wcs.pointstocoords(
+                [[0, 0], [0, ymax], [xmax, 0], [xmax, ymax]],
+                naxispath=image.naxispath)
```
pllim: Note: `naxispath` is necessary for image cube support.

ginga/canvas/types/astro.py
```diff
@@ -666,7 +679,238 @@ def move_to(self, xdst, ydst):
...
+        except Exception:
+            return []
+        ra_min, dec_min = radec.ra.min().deg, radec.dec.min().deg
+        ra_max, dec_max = radec.ra.max().deg, radec.dec.max().deg
```
pllim: Note: Such syntax is very specific to Astropy but was necessary to support image in uncommon WCS system like NCP; Otherwise, Ginga complains about negative RA values and the displayed axes don't match the info bar.

ginga/canvas/types/astro.py
```diff
@@ -666,7 +679,238 @@ def move_to(self, xdst, ydst):
...
+        if self.show_label:
+            # Calculate label orientation
+            x1, y1 = pts[0]
+            x2, y2 = pts[-1]
+            dx = x2 - x1
+            dy = y2 - y1
+            m = dy / dx
+            c = y1 - m * x1
+
+            if abs(m) < 1:  # x axis varying
```
pllim: Note: RA and DEC don't always correspond with X/Y. Sometimes the image sky orientation is at an angle w.r.t. cartesian coordinates.

ginga/rv/plugins/WCSAxes.py
pllim: Done.

pllim:
@ejeschke , I tested this on several images and it seems to work for me. As for label positioning, it will never be pretty enough for a perfectionist but it conveys the info decently enough. If someone really wants to position the labels perfectly (e.g., for publication), I would recommend hiding the labels in this plugin and manually insert each label as "text" in the `Drawing` plugin. Please review.

ejeschke:
This looks good.  I don't think it needs much more.  Couple questions/points though:
1) how about your image where the labels were all at crazy angles.  Did this get fixed with your latest changes?
2) I'm wondering whether "WCSAxes" is a reasonable name for this plugin, since there is another astropy package by that name (which is not used here).  And people may expect the same functionality as that one.  What about something like "CoordAxes"?  And maybe we would need to change both the canvas type name and the plugin name.
3) If there is a bad WCS in the image it seems to cause "bad juju" :smile_cat: in the drawing code (QPainter errors, etc).  Maybe we should trap for WCS errors at the appropriate point in the canvas type and simply substitute a single text object (centered in the middle of the image) that says "BAD WCS" or some such.
4) I'm wondering about the names of "pointstosky" and "skytopoints".  There are all kinds of weird WCSes.  How about "points_to_coord" and "coord_to_points"?  "sky" may be a misnomer for future use of this capability.

ejeschke:
For the future enhancement we might consider to allow them to specify the coordinate space (e.g. galactic, ecliptic, etc.)

pllim:
Answers:

1. Yes. Turns out to be caused by my rusty trigonometry.
2. It is basically an incarnation of `wcsaxes`, only this one uses Ginga paths, and the other one uses `matplotlib`. Of course, the implementation is different. If you really want the name changes, let me know. Otherwise, I don't think it is necessary. But let me ask @astrofrog here in case he has an opinion.
3. I already fixed the "bad juju" in my second last commit (it was second last at the time this was written). Basically, nothing is drawn if there is no valid WCS. There is no need to print any "bad WCS" as the info bar already does that.
4. "coord" is very ambiguous to me (pix coord? sky coord?). "radec" might be better if you don't want "sky". What do you think?

pllim:
> we might consider to allow them to specify the coordinate space

But that would cause info bar (where it reports RA/DEC of mouse position) to be inconsistent with the plugin drawing. I think such feature should be global to Ginga, not just the plugin. Is there a way to do that already? I am only aware of the config to choose WCS *package* but not *system*.

ejeschke:
> Basically, nothing is drawn if there is no valid WCS. There is no need to print any "bad WCS" as the info bar already does that.

Well I am thinking of uses outside of the reference viewer, where there is/or may be, no info bar. But I'm ok with drawing nothing at the moment. That could maybe be a configurable enhancement to the canvas object later.

But I'm seeing some QPainter errors with the axes, particularly when an image with bad WCS is dragged into a channel that previously had an image with good WCS.  It might be specific to the images I was testing with; I'll try to reproduce it today.

> "radec" might be better if you don't want "sky". What do you think?

"coords" might sound ambiguous, but it covers the gamut of possibilities that data points can be mapped into via a world coordinate system. I think the coordinate system might not be in RA/DEC so that's not a good choice.  I'll try to brainstorm a couple of other possibilities.

ejeschke:
> But that would cause info bar (where it reports RA/DEC of mouse position) to be inconsistent with the plugin drawing. I think such feature should be global to Ginga, not just the plugin. Is there a way to do that already? I am only aware of the config to choose WCS package but not system.

There are two keys in each channel's settings: "wcs_coords" and "wcs_display".  These have values like e.g. 'ircs' and 'sexagesimal'.  You can play with the settings in the `Preferences` plugin to see what they do. You could have this plugin query the setting and set the appropriate variables in the axes object.  

Those plugins are receiving a "field-info" callback when the cursor passes over a data point that is different from what it was previously. In that call, the following is happening:
```python
settings = viewer.get_settings()
info = image.info_xy(data_x, data_y, settings)
```
where `viewer` is the channel viewer and `image` is the image.  Then the result `info` is a bunch with attributes about the coordinate at that pixel, and taking into account the settings of the viewer.  If the attribute `ra_txt` is present, then you can assume the following are defined: `info.ra_lbl`, `info.ra_txt`, `info.dec_lbl`, `info.dec_txt`.  If the canvas object used that to construct the labels then it would be consistent everywhere.

ejeschke:
> info = image.info_xy(data_x, data_y, settings)

If necessary, we can introduce a method called `info_points(points, settings)` which would cover the needs for both cubes and efficiency.

astrofrog:
Re: name, what about ``GingaWCSAxes`` to avoid confusion?

ejeschke:
> what about GingaWCSAxes to avoid confusion?

Well, when we add a Square or a Ruler to the canvas, we don't call it a GingaSquare or a GingaRuler.  WCSAxes would be generic enough if there wasn't already a package by that name.  I think we can just come up with some reasonable word with "Axes" at the end that should prevent any confusion.

pllim:
`WCSAxesOverlay`? `WCSAxesDisplay`? `WCSAxesDrawing`? `WCSAxesCanvasObject`?

ejeschke:
> WCSAxesOverlay? WCSAxesDisplay? WCSAxesDrawing? WCSAxesCanvasObject?

If it must start with "WCSAxes" then I'd prefer "WCSAxesOverlay" out of these choices.  But I think it is the "WCSAxes" that is the confusing part.  But, I'm not hung up on this.  @pllim, if you don't think it is a problem then you can leave it as "WCSAxes".

ejeschke:
> But I'm seeing some QPainter errors with the axes, particularly when an image with bad WCS is dragged into a channel that previously had an image with good WCS. It might be specific to the images I was testing with; I'll try to reproduce it today.

I am not able to reproduce this now. As you said, if there is a bad WCS it just does not draw the image. :tada:

ejeschke:
@pllim, I think I'm ok to merge with the exception of the "pointstosky" API name.  API changes are always fraught with problems so I would like to avoid later changes. The other things talked about we can deal with as separate enhancements, if desired later.  

Back to the API:
To me there are two main concerns: 1) is that the points are specifically points in data space (as referring to the data in the image array) and 2) the output coordinate is in whatever the WCS defines, and is therefore quite variable.  I think the name ought to be somewhat indicative of this.  Ginga canvas objects allow graphics to be plotted in several coordinate spaces, so a "point" might be in degrees, for example. Therefore I'm wary of just the term "point". How about something like `datapt_to_wcspt` (and `wcspt_to_datapt`) ?

pllim:
@ejeschke , sure, I'll make that change as requested when I get the chance. Thanks for the review!

ginga/util/wcsmod.py
```diff
@@ -229,9 +232,9 @@ def register_wcs(name, wrapper_class, coord_types):
         List of names of coordinate types supported by the WCS
     """
     global custom_wcs
-    custom_wcs[wcs_name] = Bunch.Bunch(name=name,
-                                       wrapper_class=wrapper_class,
-                                       coord_types=coord_types)
+    custom_wcs[name] = Bunch.Bunch(name=name,
```
pllim: Note: I think this is actually a bug caught by PEP8 check.

pllim:
I changed the API name. Please review. Thank you!

ejeschke:
Merged!

PR #499, opened 2017-06-02
Enhancement to Contents plugin to manage images in channels

### LINKED ISSUE #107: Context menu for Contents?
When user right-clicks on a thumbnail, a context menu to load or remove image pops up. A similar menu for `Contents` is desirable. Especially because mosaic is listed only under `Contents` and not in `Thumbs`, and closing unnecessary mosaics can free up memory (I hope?). This is not critical because mosaic can still be closed via `File -> Remove Image`, but it is convenient to have.

ejeschke:
This adds Display/Move/Copy/Remove buttons to the bottom of the Contents UI.  Behavior of the Contents treeview is different.  First of all, selecting a file does not display it.  Secondly, multiple files can be selected, and using the standard Shift or Ctrl click the selection can be modified.  An action is selected by clicking one of the buttons at the bottom of the UI.  Move/Copy/Remove are all backed by pop-up dialogs to confirm the operation.

EDIT: Fix #107

### DIFF
--- a/ginga/GingaPlugin.py
+++ b/ginga/GingaPlugin.py
@@ -71,6 +71,10 @@ class GlobalPlugin(BasePlugin):
         """This method is called when an image is set in a channel."""
         pass
 
+    def blank(self, channel):
+        """This method is called when a channel is no longer displaying any object."""
+        pass
+
 
 class LocalPlugin(BasePlugin):
     """Class to handle a local plugin."""
@@ -84,6 +88,8 @@ class LocalPlugin(BasePlugin):
             self.channel = self.fv.get_channel(self.chname)
             # TO BE DEPRECATED
             self.chinfo = self.channel
+
+            self.fv.add_callback('delete-channel', self._delete_channel_cb)
 
     def modes_off(self):
         """Turn off any mode user may be in."""
@@ -120,5 +126,18 @@ class LocalPlugin(BasePlugin):
         it is doing.
         """
         pass
+
+    def blank(self):
+        """
+        This method is called when no object is displayed in the channel
+        associated with the plugin.  It can optionally clear whatever operation
+        it is doing.
+        """
+        pass
+
+    def _delete_channel_cb(self, fv, channel):
+        # stop ourself if our channel is deleted
+        if channel is self.channel:
+            self.stop()
 
 # END
--- a/ginga/canvas/types/astro.py
+++ b/ginga/canvas/types/astro.py
@@ -161,6 +161,10 @@ class Ruler(TwoPointMixin, CanvasObjectBase):
         return (text_x, text_y, text_h)
 
     def draw(self, viewer):
+        image = viewer.get_image()
+        if image is None:
+            return
+
         points = self.get_points()
         x1, y1 = points[0]
         x2, y2 = points[1]
@@ -329,6 +333,10 @@ class Compass(OnePointOneRadiusMixin, CanvasObjectBase):
         return self.within_radius(viewer, pt, p0, self.cap_radius)
 
     def draw(self, viewer):
+        image = viewer.get_image()
+        if image is None:
+            return
+
         cr = viewer.renderer.setup_cr(self)
         cr.set_font_from_shape(self)
 
@@ -475,6 +483,8 @@ class Crosshair(OnePointMixin, CanvasObjectBase):
 
             else:
                 image = viewer.get_image()
+                if image is None:
+                    return
                 # NOTE: x, y are assumed to be in data coordinates
                 info = image.info_xy(self.x, self.y, viewer.get_settings())
                 if self.format == 'coords':
--- a/ginga/examples/configs/plugin_Thumbs.cfg
+++ b/ginga/examples/configs/plugin_Thumbs.cfg
@@ -28,6 +28,9 @@ rebuild_wait = 4.0
 # Max length of thumb on the long side
 thumb_length = 150
 
+# Separation between thumbs in pixels
+thumb_sep = 15
+
 # Sort the thumbs alphabetically
 #sort_order = 'alpha'
 sort_order = None
@@ -35,7 +38,7 @@ sort_order = None
 # Thumbnail label length in num of characters (None = no limit)
 label_length = 25
 
-# Cut off long label ('left' or 'right')
+# Cut off long label ('left', 'right' or None)
 label_cutoff = 'right'
 
 # Option to highlight images that are displayed in channels.
--- a/ginga/rv/Channel.py
+++ b/ginga/rv/Channel.py
@@ -82,25 +82,28 @@ class Channel(Callback.Callbacks):
         if self == channel:
             return
 
+        # transfer image info
+        info = self.image_index[imname]
+        channel._add_info(info)
+
         try:
             # copy image to other channel's datasrc if still
             # in memory
             image = self.datasrc[imname]
 
         except KeyError:
-            # transfer image info
-            info = self.image_index[imname]
-            channel._add_info(info)
             return
 
-        #channel.datasrc[imname] = image
         channel.add_image(image, silent=silent)
 
     def remove_image(self, imname):
+        info = self.remove_history(imname)
+
         if imname in self.datasrc:
             self.datasrc.remove(imname)
+            self.fv.make_async_gui_callback('remove-image', self.name,
+                                            info.name, info.path)
 
-        info = self.remove_history(imname)
         return info
 
     def get_image_names(self):
--- a/ginga/rv/Channel.py
+++ b/ginga/rv/Channel.py
@@ -100,7 +100,7 @@ class Channel(Callback.Callbacks):
 
     def remove_image(self, imname):
         info = self.image_index[imname]
-        self.remove_history(imname)
+        changed = self.remove_history(imname)
 
         if imname in self.datasrc:
             image = self.datasrc[imname]
@@ -113,6 +113,10 @@ class Channel(Callback.Callbacks):
 
         self.fv.make_async_gui_callback('remove-image', self.name,
                                         info.name, info.path)
+
+        if changed:
+            self.refresh_cursor_image()
+
         return info
 
     def get_image_names(self):
@@ -227,6 +231,10 @@ class Channel(Callback.Callbacks):
         self.fv.make_async_gui_callback('add-image-info', self, info)
 
     def refresh_cursor_image(self):
+        if self.cursor < 0:
+            self.viewer.clear()
+            return
+
         info = self.history[self.cursor]
         if self.datasrc.has_key(info.name):
             # image still in memory
@@ -285,8 +293,6 @@ class Channel(Callback.Callbacks):
         # image was newly added
         return True
 
-
-
     def add_history(self, imname, path, idx=None,
                     image_loader=None, image_future=None):
 
@@ -314,12 +320,24 @@ class Channel(Callback.Callbacks):
         return info
 
     def remove_history(self, imname):
+        changed = False
         if imname in self.image_index:
             info = self.image_index[imname]
             del self.image_index[imname]
+            i = self.history.index(info)
             self.history.remove(info)
 
+            # adjust cursor as necessary
+            if i < self.cursor:
+                self.cursor -= 1
+            elif i == self.cursor:
+                if i >= len(self.history):
+                    # loop
+                    self.cursor = min(0, len(self.history) - 1)
+                changed = True
+
             self.fv.make_async_gui_callback('remove-image-info', self, info)
+        return changed
 
     def get_current_image(self):
         return self.viewer.get_image()
--- a/ginga/rv/Channel.py
+++ b/ginga/rv/Channel.py
@@ -63,7 +63,7 @@ class Channel(Callback.Callbacks):
         self.extdata = Bunch.Bunch()
 
         self._configure_sort()
-        self.settings.getSetting('sort_order').add_callback(
+        self.settings.get_setting('sort_order').add_callback(
             'set', self._sort_changed_ext_cb)
 
     def connect_viewer(self, viewer):
@@ -80,6 +80,10 @@ class Channel(Callback.Callbacks):
 
     def copy_image_to(self, imname, channel, silent=False):
         if self == channel:
+            return
+
+        if imname in channel:
+            # image with that name is already there
             return
 
         # transfer image info
@@ -100,22 +104,19 @@ class Channel(Callback.Callbacks):
 
     def remove_image(self, imname):
         info = self.image_index[imname]
-        changed = self.remove_history(imname)
+        self.remove_history(imname)
 
         if imname in self.datasrc:
             image = self.datasrc[imname]
             self.datasrc.remove(imname)
 
-            # clear viewer if we are removing the currently displayed image
+            # update viewer if we are removing the currently displayed image
             cur_image = self.viewer.get_image()
             if cur_image == image:
-                self.viewer.clear()
+                self.refresh_cursor_image()
 
         self.fv.make_async_gui_callback('remove-image', self.name,
                                         info.name, info.path)
-
-        if changed:
-            self.refresh_cursor_image()
 
         return info
 
@@ -233,6 +234,7 @@ class Channel(Callback.Callbacks):
     def refresh_cursor_image(self):
         if self.cursor < 0:
             self.viewer.clear()
+            self.fv.channel_image_updated(self, None)
             return
 
         info = self.history[self.cursor]
@@ -320,7 +322,6 @@ class Channel(Callback.Callbacks):
         return info
 
     def remove_history(self, imname):
-        changed = False
         if imname in self.image_index:
             info = self.image_index[imname]
             del self.image_index[imname]
@@ -330,14 +331,11 @@ class Channel(Callback.Callbacks):
             # adjust cursor as necessary
             if i < self.cursor:
                 self.cursor -= 1
-            elif i == self.cursor:
-                if i >= len(self.history):
-                    # loop
-                    self.cursor = min(0, len(self.history) - 1)
-                changed = True
+            if self.cursor >= len(self.history):
+                # loop
+                self.cursor = min(0, len(self.history) - 1)
 
             self.fv.make_async_gui_callback('remove-image-info', self, info)
-        return changed
 
     def get_current_image(self):
         return self.viewer.get_image()
--- a/ginga/rv/Channel.py
+++ b/ginga/rv/Channel.py
@@ -38,7 +38,6 @@ class Channel(Callback.Callbacks):
         self.fv = fv
         self.settings = settings
         self.logger = fv.logger
-        self.lock = threading.RLock()
 
         # CHANNEL ATTRIBUTES
         self.name = name
@@ -98,6 +97,7 @@ class Channel(Callback.Callbacks):
 
         if was_not_there_already:
             channel.datasrc[imname] = image
+            image.add_callback('modified', channel._image_modified_cb)
 
             if not silent:
                 self.fv.gui_do(channel._add_image_update, image, info)
@@ -168,6 +168,7 @@ class Channel(Callback.Callbacks):
                                     image_loader=image_loader,
                                     image_future=image_future,
                                     idx=idx)
+            image.set(image_info=info)
 
         # we'll get notified if an image changes and can update
         # metadata and make a chained callback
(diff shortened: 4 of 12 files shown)


ejeschke:
ping @pllim.  Please review when you get a chance.

pllim:
What is the motivation behind this? This new behavior does not work well when buffer is modified by the user.

Scenario 1:
1. Open image in "Image" channel.
2. Use `BadPixCorr` in `stginga` to fix some bad pixels, which is also reflected in `ChangeHistory`.
3. Copy modified image to new "Image0" channel.
4. **Problem:** Copied image has no entry in `ChangeHistory`, thus treated as "not modified" by `SaveImage`.

Scenario 2:
1. Open image in "Image" channel.
2. Use `BadPixCorr` in `stginga` to fix some bad pixels, which is also reflected in `ChangeHistory`.
3. Move modified image to new "Image0" channel.
4. **Problem**: Get a "Error removing thumb for s1_mips_v0.30_sci[PRIMARY,1]" message on terminal. Its thumbnail is now missing from `Thumbs`. `ChangeHistory` is not updated, thus now showing an entry for non-existing buffer. `SaveImage` (global plugin version) at first shows the non-existent buffer as well, and then not showing the "moved" image at all upon GUI refresh (by switching channels in combo-box). Ginga core dumped on quit.

Even without buffer modification, the following does not work...

Scenario 3:
1. Open image in "Image" channel. This can be either single or multi-extension.
2. Move the image (displayed extension) to new "Image0" channel.
3. **Problem:** Get a "Error removing thumb for ua3u0101m_c0m[SCI,1]" message on terminal. `Thumbs` display is now empty even when `Contents` shows that it has been moved. (But copy seems to work fine.)

Scenario 4:
1. Open image in "Image" channel. 
2. Open `PixTable` and mark on a pixel.
3. Move the image to new "Image0" channel.
4. Click on "Dialogs", which will bring `PixTable` back to focus.
5. **Problem:** Save `Thumbs` error as above. Get a "Error changing highlight on treeview path" message as well.

ejeschke:
> What is the motivation behind this? 

To allow users to properly manage the images in channels.  Currently there is no way to move or copy an image from one channel to another, and only one (somewhat obscure) way to remove them.

ejeschke:
> Problem: Copied image has no entry in ChangeHistory, thus treated as "not modified" by SaveImage.

I'll have to look at `ChangeHistory`, but since it is a global plugin like `Contents`, it should be able to track the move/copy of the image by subscribing to the same callbacks.  If the history is being added to the "image info" that is being kept about the image in the channels it will follow the image as it is copied or moved from channel to channel.  I believe this is what is happening, if I remember the conversations around this topic at the time.

> Get a "Error removing thumb for s1_mips_v0.30_sci[PRIMARY,1]" message on terminal. 

> Even without buffer modification, the following does not work...

> Problem: Save Thumbs error as above. Get a "Error changing highlight on treeview path" message as well.

`Thumbs` (and maybe `Contents`) probably needs some additional work in this PR.  I'll try to reproduce your errors.

ejeschke:
@pllim, updated.  Can you test briefly when you get a chance? (no hurry)  This now incorporates the `Thumbs` plugin that I had put forward in an earlier PR.

pllim:
Sorry for the wait, I'll have a look today.

pllim:
`Thumbs` organizes the thumbnails weird now. The actual layout differ depending on how you resize it but still look weird nonetheless. Screenshot below:

untitled

pllim:
Follow up to https://github.com/ejeschke/ginga/pull/499#issuecomment-306252329 with your fixes here...

**Scenario 1:** Fixed, change history is transferred properly.

**Scenario 2:** `ChangeHistory` and `SaveImage` (global plugin) seem to work now. No core dump. But "Error removing thumb" still there. `Thumbs` shows *no* thumbnails. 

**Scenarios 3 and 4:** "Error removing thumb" still there. `Thumbs` shows *no* thumbnails.

ejeschke:
Progress! :smile_cat:

ejeschke:
@pllim, so aside from aesthetic issues, the main remaining issue seems to be the "removing thumb" error, which results in an empty Thumbs pane, or possibly missing some thumbs.  But the channel contents look correct from the `Contents` side when you do these operations?

pllim:
Yes, `Contents` seems fine. There are some "highlight" error messages but they don't prevent the plugin from functioning.

ejeschke:
Scenario 3 seems easiest to replicate the error, but so far I can't.  Can you clarify the sequence of events a little?  For example, I'm doing:
- open fresh ginga
- drag an image from nautilus to the "Image" pane
- click "+" in the workspace toolbar to create "Image0" channel
- go to `Contents` and select the one image 
- click "Move"
- in the pop up dialog, select channel "Image0" and then click OK

The image is moved, and no thumbnail error. The thumb is still visible in `Thumbs` pane.  I do get the "highlight error", but as you say, it is somewhat benign.

pllim:
That is the right sequence. I am using RHEL7, Qt5, Python 3.5. Also a dev version of Astropy but that should not matter...

pllim:
One small difference is I didn't drag from Nautilus, but open it the "old school" way using `File -> Load Image` menu.

pllim:
And also I used `Channel -> Add Channel` menu, although I don't see why that would make a difference.

ejeschke:
Aha, I was able to replicate this using "File"->"Load Image".  Getting somewhere...

ejeschke:
Rebased.

ejeschke:
> Aha, I was able to replicate this using "File"->"Load Image".

It seems to be consistently reproducible with "File"->"Load Image", but not if dragging the image from `FBrowser` or Nautilus.

ejeschke:
@pllim, please check it again when you have time.  I believe the issues with the missing thumbnails are fixed.  Although we still likely need some improvement to the visual of the thumbnails, I believe the layout is also working better.

ejeschke:
Added double-click to load the image in `Contents`.

ejeschke:
Rebased again.

pllim:
Works now. Thanks! :tada:

ejeschke:
- Rebased
- Fixed corruption of Thumbs pane on redraw with many files imported at once
- Fixed scrolling of Thumbs pane

@pllim could you take a look again when you have a chance?  I think this is approaching good to merge.

The biggest issue remaining is getting the text right on the thumbs.  It depends a lot on the fonts installed on the host.  Also, it is currently not consistent with what we had with the previous `Thumbs`.

pllim:
I would prefer to wait until #519 is merged and this rebased on top of that, then I can test everything together. Is that okay with you, @ejeschke ?

ejeschke:
Sure, sounds good.  I'm going to merge #519 shortly; then I'll rebase this and let you know. Tnx.

ejeschke:
Rebased against master, and ready for further testing/comment.

Some comments on the current (non-PR) `Thumbs` plugin vs. this one:
- this one has no right-click context menu. You can click to display, as before, but the display/remove functionality is now handled by `Contents`, which also allows you to copy and move images.  
- this one packs the thumbs tighter. The name is currently truncated after the cutoff point, and probably can be adjusted using `ginga.util.iohelper.shorten_name()` (as in the current version), but I find that `shorten_name()` as it currently exists chops out too much of the filename in favor of the suffix.
- The "tooltip" pop-ups in this version of `Thumbs` seem to be more responsive and reliably pop up and down as you move over the thumbs.  This helps with the point above because you can scan quickly over the thumbs to read the names.  One difference is that the popup does not rise above all widget elements, as in the current incarnation.

pllim:
Thanks, I'll have a look when I have a chance!

pllim:
With the rebased version of this PR:

* When I select nothing on `Contents` and click Move/Copy/Delete, the popup still appears but does nothing. And when I click Display, I get error message on `Errors`. I wonder if it is less confusing without the popup/error.
* When I move the last image from `Image` to `Image0`, the image still displays on `Image` but `Contents` and `Thumbs` indicate that the channel should be empty. I thought the previous behavior was a blank display? If I have a `PixTable` plugin active on `Image`, I can even still interact with the supposedly moved image on the supposedly empty channel!
* Maybe demote "Error changing highlight on ..." from error to warning? If it is benign, raising error on the log seems a bit of an overkill. Or perhaps even a debug level?

Also, I found a scenario that is buggy:

1. Open image in "Image" channel.
2. Use `BadPixCorr` in `stginga` to fix some bad pixels, which is also reflected in `ChangeHistory`.
3. Move modified image to new "Image0" channel.
4. Copy the modified image back from "Image0" to "Image".
5. **Problem:** The change history is missing from the copied image! `Thumbs` and `Contents` appear correct though.

pllim:
> The name is currently truncated after the cutoff point, and probably can be adjusted...

Yes, the label should be user-configured via [plugin_Thumbs.cfg](https://github.com/ejeschke/ginga/blob/master/ginga/examples/configs/plugin_Thumbs.cfg).

pllim:
Oh, wait... setting `label_cutoff` in `plugin_Thumbs.cfg` does not seem to have any effect anymore. Is that intentional or a bug?

ejeschke:
> When I select nothing on Contents and click Move/Copy/Delete, the popup still appears but does nothing.

Hmm, maybe go directly to a popup in `Errors`?

> When I move the last image from Image to Image0, the image still displays on Image but Contents and Thumbs indicate that the channel should be empty. I thought the previous behavior was a blank display? If I have a PixTable plugin active on Image, I can even still interact with the supposedly moved image on the supposedly empty channel!

Yes, it moves the image internally in the channels, but doesn't actually clear the viewer.  I suppose making the viewer go blank or display a different image in the channel would be best.

> Maybe demote "Error changing highlight on ..." from error to warning? If it is benign, raising error on the log seems a bit of an overkill. Or perhaps even a debug level?

That sounds reasonable to me.  "warning" or possibly "info".

> Problem: The change history is missing from the copied image! 

Hmm, I think on a copy we need to make sure all the bits that are added to the metadata get copied. 

> Oh, wait... setting label_cutoff in plugin_Thumbs.cfg does not seem to have any effect anymore. Is that intentional or a bug?

Intentional, but temporary.  I made a quick hack to display the names with the thumbs being closer together.

pllim:
> maybe go directly to a popup in `Errors`?

It would be more intuitive (for me, at least) to have the buttons actually disabled until something is selected. But if that is too complicated to implement, the I guess `Errors` is slightly less confusing than popup.

ejeschke:
@pllim, thanks for the excellent feedback!  I think I have addressed most of your concerns in the latest version of this PR, which has been rebased on some small fixes in master.  

At your convenience, could you recheck against your recent comments here and see if things are working as expected?  Thank you!

ejeschke:
I might still need to adjust the text/image separation a little.

pllim:
> I might still need to adjust

Is this ready for re-review now or you still need some time? :smile:

ejeschke:
It's ready! 😸

pllim:
When I have a table and an image open on `Image0`, and then move the table to `Image`, `Image0` does not automatically focus on the remaining image.

Unfortunately, more detailed feedback using `stginga` plugin is currently being held up by #522.

ejeschke:
> Unfortunately, more detailed feedback using stginga plugin is currently being held up by #522.

I rebased against master with the fix.

ejeschke:
> When I have a table and an image open on Image0, and then move the table to Image, Image0 does not automatically focus on the remaining image.

I pushed a change that will clear the viewer regardless of whether it is a table viewer or an image viewer, if you move the thing you are looking at.  Clicking a new image or table will then view it in that channel.

The behavior could be changed to show the previous or next object in the channel, if there is one.  Depends on what should have the least surprise for the user, I think.  Your thoughts?

ejeschke:
> The behavior could be changed to show the previous or next object in the channel

After thinking about this for a bit, it seems like showing something else in the channel is more useful than just blanking the viewer, and may be more in line with what the user might expect.  I've pushed a change that will load the image at the channel "cursor" if one is deleted. In practice, this means that the next image/table in the channel is displayed, unless you are on the last one, in which case it wraps around and shows the first one.  Let me know if this seems more intuitive.

pllim:
One more rebase on master and I can continue reviewing this... :smile:

ejeschke:
Rebased!

pllim:
> Let me know if this seems more intuitive.

Now that I move the table (or image), Ginga automatically focuses that table (or image) but in the channel it is moved to. Seems intuitive for me. :+1:

pllim:
>>Oh, wait... setting label_cutoff in plugin_Thumbs.cfg does not seem to have any effect anymore. Is that intentional or a bug?
> Intentional, but temporary. I made a quick hack to display the names with the thumbs being closer together.

`label_cutoff` set in `plugin_Thumbs.cfg` is still ignored. I would prefer it to work again before this gets merged, if possible. Otherwise, we should open a new issue for this.

pllim:
Here is another one.

1. Open an image in `Image` channel.
2. Open `Drawing` pluging.
3. Draw a shape and click `Create Mask` to create a new `drawing0` image.

Now, `Thumbs` look like this (alignment seems off?):
untitled

4. Select `drawing0` in `Contents` and click "Remove".

Now, I see this traceback on the terminal:
```
2017-07-07 16:47:33,409 | E | GwMain.py:74 (_execute_future) |
  gui event loop error: ('Image', 'drawing0', None)
2017-07-07 16:47:33,421 | E | GwMain.py:78 (_execute_future) | Traceback:
  File ".../ginga/gw/GwMain.py", line 71, in _execute_future
    res = future.thaw(suppress_exception=False)
  File ".../ginga/misc/Future.py", line 45, in thaw
  File ".../ginga/rv/plugins/Thumbs.py", line 876, in reorder_thumbs
    bnch = self.thumb_dict[thumbkey]
```

pllim:
Another scenario I found (almost by accident as I tried to be lazy and tried to re-test old scenarios above in the same Ginga session; so I don't know how likely people are going to encounter this in real-life):

1. Open image in `Image` channel.
2. Create a new `Image0` channel.
3. Copy image to `Image0` channel. Now, the image is in two different channels.
4. Move image from `Image` to `Image0`. Now, the image is only in the second channel.
5. Remove image in `Image0`. I see the traceback below.

```
2017-07-07 16:56:00,923 | E | Channel.py:418 (switch_name) |
  Can't switch to image 's12_ch1_v0.30_sci[PRIMARY,1]':
  No image by the name 's12_ch1_v0.30_sci[PRIMARY,1]' found
Error making callback 'activated': No image by the name 's12_ch1_v0.30_sci[PRIMARY,1]' found
Traceback:

  File ".../ginga/rv/plugins/Contents.py", line 496, in <lambda>
    action))

  File ".../ginga/rv/plugins/Contents.py", line 519, in action_images_cb
    src_channel.remove_image(info.imname)

  File ".../ginga/rv/Channel.py", line 118, in remove_image
    self.refresh_cursor_image()

  File ".../ginga/rv/Channel.py", line 245, in refresh_cursor_image
    self.switch_name(info.name)

  File ".../ginga/rv/Channel.py", line 419, in switch_name
    raise ChannelError(errmsg)
```

ejeschke:
> label_cutoff set in plugin_Thumbs.cfg is still ignored. I would prefer it to work again before this gets merged, if possible. Otherwise, we should open a new issue for this.

Hmm, I thought I had that back in pretty much as before.  Let me check the defaults.  I might have changed the cutoff default amount.  Don't worry, I'm not in a hurry to merge this until it is ready.   I'll take a look at the other issues you mentioned.

ejeschke:
> label_cutoff set in plugin_Thumbs.cfg is still ignored. I would prefer it to work again before this gets merged, if possible. Otherwise, we should open a new issue for this.

Could you confirm that you have `label_length` as non-None (should be a number like, say, 25) and `label_cutoff` as "right" in your `plugin_Thumbs.cfg`?  In the code it is calling `iohelper.shorten_name()` just as it was previously IFF both of these are set.

pllim:
@ejeschke , please ignore my `label_cutoff` "problem" report at https://github.com/ejeschke/ginga/pull/499#issuecomment-313785850 . I forgot that the algorithm does not chop of extension name (until I read the code again) and my name has a very long extension name ("[PRIMARY,1]"). Once I set the `label_length` long enough to include the extension name, but not long enough to include everything else, then the label behaves as expected. Sorry for the confusion! :sweat_smile: 

For completeness, this is the function docstring:
```
Shorten `name` if it is longer than `char_limit`.
If `side` == "right" then the right side of the name is shortened;
if "left" then the left side is shortened.
In either case, the suffix of the name is preserved.
```

ejeschke:
Ok, @pllim!  Hopefully I have squashed the last of these obvious :bug:s. Tested your scenarios pretty thoroughly.  Could you give it one more whirl, when you get the chance?

ejeschke:
It's rebased on master.

pllim:
I'm sorry, I found one more... 😅 

1. Open an image in `Image` channel.
2. Create new `Image0` channel.
3. Move image to `Image0` channel.
4. Modify the moved image in `Image0` channel (e.g., with `BadPixCorr` plugin from `stginga`).
5. I get the traceback below on the terminal and missing entry in `History`.

```
2017-07-10 12:31:01,349 | W | Thumbs.py:268 (_add_image) |
  no information in channel about image 's1_mips_v0.30_sci[PRIMARY,1]'
2017-07-10 12:31:01,351 | E | GwMain.py:74 (_execute_future) |
  gui event loop error: 's1_mips_v0.30_sci[PRIMARY,1]'
2017-07-10 12:31:01,356 | E | GwMain.py:78 (_execute_future) | Traceback:
  File ".../ginga/gw/GwMain.py", line 71, in _execute_future
    res = future.thaw(suppress_exception=False)
  File ".../ginga/misc/Future.py", line 45, in thaw
  File ".../ginga/rv/plugins/ChangeHistory.py", line 191, in redo
    iminfo = channel.get_image_info(imname)
  File ".../ginga/rv/Channel.py", line 202, in get_image_info
    return self.image_index[imname]
```

ejeschke:
Hmm, the error seems to be coming from `ChangeHistory`, which is not aware that the image moved.  I've pushed a couple small changes to `ChangeHistory`.  Would you mind trying this test again, @pllim?  I think you can just do a fetch/merge on the PR since I didn't rebase it.

pllim:
@ejeschke , unfortunately 621ca7f did not fix the problem. I still see the same error as https://github.com/ejeschke/ginga/pull/499#issuecomment-314161707 .

ejeschke:
@pllim, I have fixed the issue that causes the traceback with `BadPixCorr` (by actually testing with `stginga` and `BadPixCorr`).  Could you give it yet another whirl when you have a chance?

I've rebased against master you you'll need to apply the PR afresh.  Much appreciated!

pllim:
@ejeschke , thank you for your patience. Unfortunately, this is still a little problematic with the workflow in https://github.com/ejeschke/ginga/pull/499#issuecomment-314161707 -- There is no more traceback (step 5) but `History` is also *not* showing the changes made as it is supposed to.

p.s. Other workflows above seem to work fine now.

p.p.s. I also managed to produce an error with `Drawing` plugin after moving or copying an image, but I could not reproduce it. This is just a note to future self if I see it again.

ejeschke:
I've rebased against master, and force pushed up the PR.  I also updated my `stginga` from upstream and installed. Trying to reproduce the issue of `History` not showing the change.  I may need to understand your workflow in a little more detail.  Here is what I do:
1. open the image using File->Open (since I remembered you use this sometimes) into channel "Image".  I'm using a HST file I got from Warren way back when for testing, but it has the DQ flags HDU so that the `BadPixCorr` can work correctly.
2. Create a new channel "Image0" by pressing the '+' icon in the workspace toolbar
3. Go to `Contents`, select the image and move it to channel Image0
4. Start up `BadPixCorr` on channel Image0, draw a sufficiently large annulus on the image and click "Fix Bad Pixels".  The area is replaced and I get the message "BadPixCorr.py:864 (fix_bpix) | Bad pixel flag(s) replaced in ibm502hkq_flt[DQ,1]; dqflag=0, npix=192" in the log.
5. If I now look at the "History" tab and select the first terminal row I see the image shown below.

badpix

It looks Ok to me, including the channel name.  What am I doing differently from you?

ejeschke:
I think there may be some issues with certain plugins when you try to move an image they are working on actively to another channel.  I think these plugins probably need to be modified as needed on a case by case basis to handle having their working image "yanked out from under them".  I've added another method to the plugin API to take care of this case. When there is no image available to switch to in the channel, it calls `blank()` on the plugin (instead of `redo()`).  The plugin can then take such action as to correct its state to something sensible for no image loaded.

pllim:
@ejeschke , I used a single-extension FITS with no DQ, which you can download from https://stsci.box.com/s/zstyluc9wcujccquni6mpl7k6tpgjqmk . Not having DQ should not stop history from showing up, especially when you can clearly see the plugin made some changes to buffer. Also, the history shows up fine with this file I share here when I don't move it to another channel first.

Therefore, it is not clear to me what is wrong. The code on `stginga` that adds the change history is at about https://github.com/spacetelescope/stginga/blob/master/stginga/plugins/BadPixCorr.py#L838

ejeschke:
@pllim, what is the name of the filename in Box?  Clicking the link took me to the "Ginga-dev" folder.

ejeschke:
Rebased.

pllim:
Hmm, that's weird; It should have prompted you for downloading that file. Anyway, the filename is "s1_mips_v0.30_sci.fits".

ejeschke:
@pllim with this file I am able to reproduce the problem.

ejeschke:
Gosh, have we really been working on this since June 2?  :persevere: 

@pllim, I found the issue with this--I wasn't tracking the image `modified` callback when the image moved to a new channel.  I believe this is fixed now.  Additionally, I have rebased against master.  Could you take a quick look when you have time?  I believe we are spiraling toward a merge.

ejeschke:
Excellent! :+1:

ejeschke:
Rebased against master.

ejeschke:
Merging.

pllim:
That only took like 2.5 months and 66 comments. 😉

ejeschke:
Whew! :weary:

Issue #392, opened 2016-11-09
How does Ginga handle memmap-ed FITS handlers?


pllim:
As discussed during JWST Data Analysis Workshop II on Nov 9, 2016. Does Ginga close FITS handlers properly when `memmap=True` (default)? See http://docs.astropy.org/en/stable/io/fits/#working-with-large-files 

*When opening a file with memmap=True, because of how mmap works this means that when the HDU data is accessed (i.e. hdul[0].data) another handle to the FITS file is opened by mmap. This means that even after calling hdul.close() the mmap still holds an open handle to the data so that it can still be accessed by unwary programs that were built with the assumption that the .data attribute has all the data in-memory.*

*In order to force the mmap to close either wait for the containing HDUList object to go out of scope, or manually call del hdul[0].data (this works so long as there are no other references held to the data array).*

pllim:
Well, I think I answered my own question... See STScI-JWST/wss_tools#26

@ejeschke , how do I set `memmap=False` in Ginga reference viewer? I don't see any way to actually do that currently without changing `GingaShell.load_image()` code in `Control.py`.

ejeschke:
@pllim, in the `ginga.util.io_fits` module, the appropriate methods have a `memmap` kwarg that is passed to the opener.  Furthermore, that keyword can be passed as part of the kwargs in `Control`'s `load_image()` which is the basis for loading most files (as you note above).  

So there is a mechanism for passing it in, but it is currently not used.

The default is `None` in `io_fits`.  My understanding that unless `memmap=False` the default is `True` for astropy.  So I think it is being turned on by default.

Are you thinking of making a setting in `general.cfg` or something?

pllim:
Either "general.cfg" or "channel_Image.cfg" is fine by me. Another option is command line argument but I think config setting is more elegant.

pllim:
By the way, it does default to `memmap=True`: https://github.com/astropy/astropy/blob/master/astropy/io/fits/file.py#L86

```python
memmap = True if memmap is None else memmap
```

This line was added 3 years ago (according to "git blame"), so this is not something terribly new.

pllim:
Or... we can just force `memmap=False`. Does Ginga really need `memmap`?

ejeschke:
Well, in my own testing, `memmap=False` is actually faster for the case of mosaicing, or any typical use case in Ginga.  So maybe we can consider to set the default to `False`.  Can you see whether that solves your issue?

[...]

PR #501, opened 2017-06-06
Set default FITS memmap to False

pllim:
As discussed in https://github.com/ejeschke/ginga/issues/392#issuecomment-306285745 but does not really solve the "too many open files" issue.

### DIFF
--- a/ginga/util/io_fits.py
+++ b/ginga/util/io_fits.py
@@ -156,7 +156,7 @@ class PyFitsFileHandler(BaseFitsFileHandler):
 
         return dstobj
 
-    def load_file(self, filespec, numhdu=None, dstobj=None, memmap=None,
+    def load_file(self, filespec, numhdu=None, dstobj=None, memmap=False,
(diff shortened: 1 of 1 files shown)


pllim:
When I load a 6.7 GB data cube, `memmap=None` (`True`) loads it almost instantaneously in Ginga (because it only loads the slice being displayed) but took a lot longer when `memmap=False`. So, I think we should just close this without merging. What do you think, @ejeschke ?

ejeschke:
Good to hear of a counter case.  I think we can close the PR.  It would be nice to have a way (maybe a general or channel setting) to pass as the `memmap` kwarg when opening files.  The user could then have some control over that be setting the preference to `True`, `False` or `None`.  A future enhancement, maybe.

pllim:
I find that I can set `memmap=False` globally for Astropy in my `~/.astropy/config/astropy.cfg` anyway, so perhaps a new Ginga setting is not urgent.

PR #506, opened 2017-06-06
Speed up LineProfile calculation etc

### LINKED ISSUE #492: Feature Request: Overplotting multiple marks on LineProfile
User wishes to overplot multiple marks on `LineProfile`.

Update: Alicia would settle for a "all or selected" option.

c/c @bhilbert4 @aliciacanipe

### LINKED ISSUE #493: LineProfile mean calculation too slow for big cube
User reported performance issues ("a few seconds") when `LineProfile` operates on a cube of dimension `[190,2048,2048]`, with "circle" region and a radius of about 30 pixels.

Under the hood, I used Numpy's `mean(axis=...)` method for masked array. Not sure how to make that faster, so I am opening an issue for now as a reminder to investigate in the future.

c/c @bhilbert4 @aliciacanipe

Alicia's timing report:

> I did a quick check on a point shape versus a circle shape.
       
>        1st point after opening Ginga: ~5 sec to get the ramp
>        Clicking on another point after that: ~1-2 sec to get the new ramp

>        1st circle after opening Ginga (r = 5 pix): ~40 sec to get the mean ramp
>
> and I get the spinning beachball
>
>        Moving the circle to another position: ~8 sec to get the mean ramp

pllim:
Speed up `LineProfile` calculation by getting rid of Numpy masked array overhead. Fix #493.

I thought using masked array was a elegant solution but it is not feasible for large data cube. 😢 I probably over-designed it originally... Timing info available at https://gist.github.com/pllim/d0999e1c00d4ee45ccba2585c0b8146f .

Also fix #492.

c/c @bhilbert4 @aliciacanipe

### DIFF
--- a/ginga/rv/plugins/LineProfile.py
+++ b/ginga/rv/plugins/LineProfile.py
@@ -86,19 +86,21 @@ class LineProfile(GingaPlugin.LocalPlugin):
         top = Widgets.VBox()
         top.set_border_width(4)
 
-        vbox, sw, orientation = Widgets.get_oriented_box(container)
-        vbox.set_margins(4, 4, 4, 4)
-        vbox.set_spacing(2)
+        box, sw, orientation = Widgets.get_oriented_box(container)
+        box.set_border_width(4)
+        box.set_spacing(2)
+
+        paned = Widgets.Splitter(orientation=orientation)
 
         self.plot = plots.Plot(logger=self.logger,
-                               width=400, height=300)
+                               width=400, height=400)
         ax = self.plot.add_axis()
-        ax.grid(False)
+        ax.grid(True)
         self._ax2 = self.plot.ax.twiny()
 
         w = Plot.PlotWidget(self.plot)
-        w.resize(400, 300)
-        vbox.add_widget(w, stretch=0)
+        w.resize(400, 400)
+        paned.add_widget(Widgets.hadjust(w, orientation))
 
         fr = Widgets.Frame("Axes controls")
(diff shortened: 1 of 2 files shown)


pllim:
I think this is ready now. Please review.

pllim:
c/c @nabobalis

ejeschke:
This looks good to me, @pllim.  

I do think that the Move/Draw/Edit controls should be standardized (as much as possible) accounting for the different things that the plugins do.  And `LineProfile` is perfectly compatible in this regard, but the controls are constructed differently and show up more widely spaced than in other plugins (because they end up in a `GridBox` rather than in an `HBox`).  This is a minor point, but these types of minor inconsistencies can contribute to an overall sense of inconsistency for the application.  So I would suggest that in a future work we either make these look like the others, or make the others look like these.

pllim:
Sure. We can open a "standardize move/draw/edit" button issue.

ejeschke:
I'm going to go ahead and merge this then, and make a new release of ginga.  We can open up issues for fixes in the master branch for any that crop up.

nabobalis:
Sorry, I was a bit slow!

This looks very nice.

pllim:
Thanks!

ejeschke:
> Sorry, I was a bit slow!

Not at all, @nabobalis!  I am just moving quickly to get a release out, since a lot of features and bug fixes have piled up in the dev line.  Thanks for your feedback!

PR #508, opened 2017-06-07
[WIP] Stop override of plugin docstrings

### LINKED ISSUE #464: Is new Help compatible with __doc__ override?
For instance in `ChangeHistory.py`, we added this to auto-insert config file contents to online documentation:
```python
# Replace module docstring with config doc for auto insert by Sphinx.
# In the future, if we need the real docstring, we can append instead of
# overwrite.
from ginga.util.toolbox import generate_cfg_example  # noqa
__doc__ = generate_cfg_example('plugin_ChangeHistory', package='ginga')
```

However, `WBrowser` now reverts to displaying module docstring when RTD or online doc fails to display.

How should we consolidate these two? It seems like a chicken-and-egg thing.

pllim:
Stop override of plugin docstrings so `WBrowser` can display its native docstring when RTD download option failed. This should not change the visual of current RTD rendered doc.

Fix #464

TODO:
- [ ] Investigate the best way to insert config contents into RST.
- [ ] Implement this for all other affected plugins.
- [ ] Figure out a way to reduce redundancy between plugin RST and docstring management?

### DIFF
--- a/doc/conf.py
+++ b/doc/conf.py
@@ -29,6 +29,10 @@ import datetime
 import os
 import sys
 
+from astropy.extern import six
+from sphinx.util.compat import Directive
+from docutils import nodes, statemachine
+
 try:
     import astropy_helpers
 except ImportError:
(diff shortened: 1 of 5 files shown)


pllim:
Would such implementation work for you, @ejeschke ? Usually, I am wary of using `exec` but it seems necessary here.

doc/conf.py
```diff
@@ -164,3 +168,37 @@
 
     edit_on_github_source_root = ""
     edit_on_github_doc_root = "docs"
+
+# -- Insert generated string from code --------------------
+# https://stackoverflow.com/questions/7250659/python-code-to-generate-part-of-sphinx-documentation-is-it-possible
+
+
```
ejeschke: This looks like some serious "sphinx-fu" :smile_cat:

pllim:
@ejeschke , okay, I'll play around with this a bit more after Astropy v2 work, so it might not be for a few weeks.

pllim:
Superseded by #548.

PR #512, opened 2017-06-12
First step toward generalizing transforms to N dimensions

ejeschke:
This is a major refactoring of transforms and coordinate mapping for Ginga canvases.  The main goal of the refactoring is to allow numpy vectorization of coordinate transformations all the way up and down the chain for any graphics type.  This PR removes almost all uses of `map` and substitutes numpy arrays and vectorized transforms in many places.  The result is clearer and faster code.

The changes are to  (mostly) internal APIs, and there should be very little effect to users and developers. However, since this affects the rendering pipeline for all back ends, it should be tested quite thoroughly before merging.  One backend (matplotlib) is slightly broken by this PR and it needs a fix also before this can be merged.

- transform.py is vectorized, which changes the internal API
- files using transform or coordmap internal API changed accordingly

### DIFF
--- a/ginga/Bindings.py
+++ b/ginga/Bindings.py
@@ -541,14 +541,14 @@ class ImageViewBindings(object):
         arr = np.array([(mnwd, mnht), (mxwd, mnht),
                         (mxwd, mxht), (mnwd, mxht)],
                        dtype=np.float)
-        x, y = tr.to_(arr.T[0], arr.T[1])
+        x, y = tr.to_(arr).T
 
         rx1, rx2 = np.min(x), np.max(x)
         ry1, ry2 = np.min(y), np.max(y)
 
         rect = viewer.get_pan_rect()
         arr = np.array(rect, dtype=np.float)
-        x, y = tr.to_(arr.T[0], arr.T[1])
+        x, y = tr.to_(arr).T
 
         qx1, qx2 = np.min(x), np.max(x)
         qy1, qy2 = np.min(y), np.max(y)
@@ -594,7 +594,7 @@ class ImageViewBindings(object):
         arr = np.array([(mnwd, mnht), (mxwd, mnht),
                         (mxwd, mxht), (mnwd, mxht)],
                        dtype=np.float)
-        x, y = tr.to_(arr.T[0], arr.T[1])
+        x, y = tr.to_(arr).T
 
         rx1, rx2 = np.min(x), np.max(x)
         ry1, ry2 = np.min(y), np.max(y)
@@ -602,7 +602,7 @@ class ImageViewBindings(object):
         crd_x = rx1 + (pct_x * (rx2 - rx1))
         crd_y = ry1 + (pct_y * (ry2 - ry1))
 
-        pan_x, pan_y = tr.from_(crd_x, crd_y)
+        pan_x, pan_y = tr.from_((crd_x, crd_y))
         self.logger.debug("crd=%f,%f pan=%f,%f" % (
             crd_x, crd_y, pan_x, pan_y))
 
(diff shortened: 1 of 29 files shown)


ejeschke:
ping @pllim.  This is inspired by your PR that allowed vectorizing the first stage of the transformation (WCS). With these changes it is carried all the way down through the rendering pipeline.  Please test when you get a chance.  This would be a really nice "under the hood" update for Ginga.

ejeschke:
Updated.  Now most issues with matplotlib backend introduced by this PR are resolved.

ejeschke:
@pllim, I've added vectorized versions of `datapt_to_wcspt` and `wcs_to_datapt` for Kapteyn and Starlink WCS wrappers.  I've also renamed your `datapt_to_coords` as `datapt_to_system` (since there is a "coords" kwarg parameter that is not related (particularly) to the "coords" in the method name and also to make it compatible name-wise with the `pixtosystem` call).

pllim:
@ejeschke , what would be the best way to test this? For instance, does making a big mosaic take advantage of such vectorization?

ejeschke:
> what would be the best way to test this? For instance, does making a big mosaic take advantage of such vectorization?

I think the best test is just to test any plugins that you regularly use that do plotting on top of the image.  

I just pushed another commit on top of this that adds coordinate conversions--you can see this if you run the `Drawing` plugin and try to edit a drawn object--there is now a pulldown that let's you convert the object from one coordinate space to another. The conversions are somewhat approximate for now (points are not perfectly preserved for items with radii, for example), but it is close enough to be useful.

pllim:
I don't know whether (a) I just didn't pay attention before, (b) it is due to my remote SSH connection, (c) or this PR, but the drawing overlay lines seem "fuzzy". Do you see the same thing?

ejeschke:
> drawing overlay lines seem "fuzzy". Do you see the same thing?

This PR doesn't change the canvas renderers, so I don't think it could have anything to do with that.

On my local Linux workstation and my mac laptop with Qt5 back end everything looks pretty sharp to my eye.

pllim:
Okay, I'll double check when I'm physically logged into the machine again. I never had any performance problem with drawing, so hard to tell how much this improves stuff.

ejeschke:
I think for normal plotting of a few objects you will not notice a big difference.  But this should be a win for when an object with many points (large N polygon or some such) is plotted.  Perhaps more important, it cleans up the code quite a bit by removing many `map` calls.

pllim:
Playing around with this PR, I see the following when I have `MultiDim` active and open an image cube. Not sure if it is a bug in `master` or this PR and I don't have time to find out right now:

```
2017-06-22 16:43:58,952 | E | GwMain.py:74 (_execute_future) | gui event loop error: 'int' object has no attribute 'lower'
2017-06-22 16:43:58,956 | E | GwMain.py:78 (_execute_future) | Traceback:
  File ".../ginga-2.6.5.dev1559-py3.5.egg/ginga/gw/GwMain.py", line 71, in _execute_future
    res = future.thaw(suppress_exception=False)
  File ".../ginga-2.6.5.dev1559-py3.5.egg/ginga/misc/Future.py", line 45, in thaw
  File ".../ginga-2.6.5.dev1559-py3.5.egg/ginga/rv/plugins/MultiDim.py", line 601, in redo
    elif info.htype.lower() in ('bintablehdu', 'tablehdu',):
```

pllim:
Other than that, I don't have any objections to this PR. Unfortunately, I am not physically at my machine today and the other machine I use for another project is down, so it is quite impossible to have an apple-to-apple performance comparison. I glanced through the code and did not see anything shocking.

ejeschke:
I just tried a couple image cubes--no issue.  I think from the error message you posted here it is unrelated to the PR and has to do with HDU parsing/name stuff.

> Other than that, I don't have any objections to this PR.
In that case, I think I will merge it, and if any new issues crop up we can file them against master branch.

Issue #327, opened 2016-04-28
Add option to use gWCS


pllim:
@mcara suggested that Ginga should add the option to support [gWCS](http://gwcs.readthedocs.io/en/latest/) (https://github.com/spacetelescope/gwcs) for future JWST support. I am just opening an issue here so I don't forget that this request happened.

c/c @larrybradley, @nden, @stscieisenhamer

ejeschke:
I think that should be straightforward, given that we already have the pluggable WCS support.  I note from the link that `asdf` is required.  I think Mike D. did a quick hack to see if he could make a version of `MultiDim` that browsed the hierarchical structure of an asdf file and load the images.  This was at the PiA2015 session, I believe.  And I think he had something working within an hour or so.

pllim:
Thanks, that sounds promising! Unfortunately, Mike D. is no longer at STScI. Do you happen to have a copy of his hack?

pllim:
According to @nden , for uncalibrated JWST data, Astropy WCS will not be able to read the distortion information but GWCS can.

pllim:
Update: After some internal discussions, it looks like I might have to move this one to `stginga` instead. If that is indeed the case, I'll close the issue here. FYI.

ejeschke:
Given the community interest in gWCS, it seems like it would be useful to have that as a WCS option in Ginga.  But perhaps it can be added later copying the approach in `stginga`.

pllim:
My understanding so far is that it is only useful if you have data model schemas to go with it (e.g. schemas embedded in JWST pipeline). Discussion is still ongoing. Hopefully I'll know more next week. Thanks for your patience.

nden:
@pllim It's more like, if you want to display JWST raw data you will need the jwst package installed.
Asdf works with schemas in third party packages through extensions defined through entry points. The only hard dependence for ginga would be asdf and gwcs. Depending on where the data comes from and whether they are final products, users may need to have some other packages installed. I don't find this unreasonable especially if working with raw data.

pllim:
> Asdf works with schemas in third party packages through extensions defined through entry points

Exactly. It only works with schemas. And currently only JWST uses it (and perhaps SunPy but in a different way). I can't envision how to support such a thing in a generic way over here. It is not like FITS WCS where all you need is a header (for better or worse). Might be better to contain this in `stginga` for now until it can be more generic. That is my current thought but we can discuss next week.

pllim:
[...]

PR #516, opened 2017-06-23
Refactor wcsmod module into individual modules for each wrapper

ejeschke:
- puts each WCS wrapper in a separate module
- for easier administration and editing


pllim:
I think `__init__.py` needs some extra imports into the `ginga.util.wcsmod` namespace to make it backward-compatible.

ejeschke:
I was wondering about that.  But I didn't find any grepping through the Ginga source code or testing so far.  Are there some STScI plugins that are importing things that are not present in this version?

pllim:
> Are there some STScI plugins that are importing things that are not present in this version?

Not that I can `grep` but STScI is not the only user of Ginga. What about SunPy etc?

ejeschke:
@pllim, from what I can tell, the only "interesting" imports from `wcsmod` were "WCS", "display_types", "coord_types" and "use".  These are all preserved.  The only other thing of interest might be if someone wanted to import one of the WCS wrapper classes directly.  But there is no place in Ginga that does that, and we don't in any of our Subaru plugins.  Sounds like STScI doesn't either.  I don't think that wouldbe likely for any end users either--they would be using the native WCS, not the wrapper.  So I think we are ok here, actually, so long as these four are preserved.

pllim:
I briefly did import one of the wrapper classes directly for #495 but then undid it as part of your reviews. Therefore, I am satisfied by your explanation above.

ejeschke:
Ginga has unit tests for wcsmod, and these pass with the PR.

ejeschke:
One thing I am thinking about (and I guess I will leave it for a future PR) is to refactor `pixtoradec`, `radectopix` and `pixtosystem` to just be wrappers around the vectorized versions that you created.

ejeschke:
Another thing that this PR does is open the door for using different WCS packages with different images.  So for example, you could have a "gwcs" wrapper for some images and "astropy" for some others.  You can imagine other possibilities.  Pretty straightforward with the changes made here.

ejeschke:
Merging.

ejeschke:
Oh, you might need to remove `wcsmod.pyc` or other cached versions in order to avoid accidentally importing the old module.

pllim:
Yes, this would be useful for #327 and #434.

PR #518, opened 2017-06-23
Astropy v2.0 compatibility for MultiDim

### LINKED ISSUE #517: MultiDim breaks with Astropy v2
As discovered in https://github.com/ejeschke/ginga/pull/512#issuecomment-310497596, `MultiDim` breaks due to changes in astropy/astropy#6124. This is because https://github.com/ejeschke/ginga/blob/master/ginga/util/io_fits.py#L220 hardcodes the index locations:

```python
            if len(tup) > 5:
                d.setvals(htype=tup[2], dtype=tup[5])
```

The correct approach is to use the relevant FITS file handler attributes directly without using `.info()`.

pllim:
Fix #517


ejeschke:
Looks straightforward enough. Merging.

Thanks, @pllim.

PR #519, opened 2017-06-24
Refactor internal shapes code for point handling consistency

ejeschke:
Refactors shapes to have a more unified representation of points. 

This allows more efficient sharing of code that operates of points, removes "2D-isms" for a more unified approach to coordinate points in the API, and makes internal canvas code more consistent between shapes.
    
Objects with (x, y) and (x1, y1, x2, y2) attributes are refactored to store the points in an array of points notation similar to polygon-based shapes.  Properties are provided for backward-compatible access to these attributes.

### DIFF
--- a/ginga/ImageView.py
+++ b/ginga/ImageView.py
@@ -1414,9 +1414,9 @@ class ImageViewBase(Callback.Callbacks):
 
         Returns
         -------
-        rect : tuple
+        limits : tuple
             Bounding box in data coordinates in the form of
-            ``(x1, y1, x2, y2)``.
+               ``(ll_pt, ur_pt)``.
 
         """
         if self._limits is not None:
@@ -1448,7 +1448,7 @@ class ImageViewBase(Callback.Callbacks):
         ----------
         limits : tuple or None
             A tuple setting the extents of the viewer in the form of
-            ``((x1, y1), (x2, y2))``.
+            ``(ll_pt, ur_pt)``.
         """
         if limits is not None:
             assert len(limits) == 2, ValueError("limits takes a 2 tuple")
@@ -1547,8 +1547,8 @@ class ImageViewBase(Callback.Callbacks):
             #self.logger.warning("new scale would exceed max/min; scale unchanged")
             raise ImageViewError("new scale would exceed pixel max; scale unchanged")
 
-        # It is necessary to store these so that the get_data_xy()
-        # (below) calculations can proceed
+        # It is necessary to store these so that the get_pan_rect()
+        # (below) calculation can proceed
         self._org_x, self._org_y = pan_x - self.data_off, pan_y - self.data_off
         self._org_scale_x, self._org_scale_y = scale_x, scale_y
         self._org_scale_z = (scale_x + scale_y) / 2.0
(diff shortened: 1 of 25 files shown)


ejeschke:
@pllim, can you have a look/test when you get a chance?  Not urgent.  This is a follow on to #512.  As such, it shouldn't have any effect on users as it is mostly changes to "internal" APIs.

ejeschke:
Rebased.

ejeschke:
> can you have a look/test when you get a chance?

Please wait a little.  Pushing more commits.

ejeschke:
Ok @pllim, sorry for the hold up.  I think this is ready for review/test.  I might push one more commit, but it would just be to update one of the example programs.  There are no special things to watch for--the updates are all in the path of regular rendering.

ginga/ImageView.py
```diff
@@ -1545,25 +1547,21 @@ def _calc_bg_dimensions(self, scale_x, scale_y,
...
-        xlr, ylr = self.get_data_xy(win_wd, win_ht)
-        xll, yll = self.get_data_xy(0, win_ht)
+        pts_t = self.get_pan_rect().T
 
         # determine bounding box
-        a1 = min(xul, xur, xlr, xll)
-        b1 = min(yul, yur, ylr, yll)
-        a2 = max(xul, xur, xlr, xll)
-        b2 = max(yul, yur, ylr, yll)
+        a1, b1 = np.min(pts_t[0]), np.min(pts_t[1])
```
pllim: I think you can just call `np.min()` one time and use `axis=1`. For example:
```python
>>> a = np.arange(10).reshape((2,5))
>>> a
array([[0, 1, 2, 3, 4],
       [5, 6, 7, 8, 9]])
>>> np.min(a, axis=1)
array([0, 5])
```

ginga/ImageView.py
```diff
@@ -1545,25 +1547,21 @@ def _calc_bg_dimensions(self, scale_x, scale_y,
...
+        a2, b2 = np.max(pts_t[0]), np.max(pts_t[1])
```
pllim: Same comment as `np.min()` above.

ginga/rv/plugins/Catalogs.py
```diff
@@ -646,19 +646,19 @@ def filter_results(self, starlist, filter_obj):
         if filter_obj:
             num_cat = len(starlist)
             self.logger.debug("number of incoming stars=%d" % (num_cat))
-            # TODO: vectorize wcs lookup
-            coords = [image.radectopix(star['ra_deg'], star['dec_deg'])
-                      for star in starlist]
-            arr = numpy.array(coords)
-            self.logger.debug("arr.shape = %s" % str(arr.shape))
+            coords = numpy.asarray([(star['ra_deg'], star['dec_deg'])
```
pllim: I see that you changed `numpy` to `np` in other files. Maybe change it here too for consistency.

ejeschke: > are there any changes that are not backward compatible?

In terms of user-facing functions, only a few of the canvas API are affected, and these are preserved by compatibility functions at the bottom of `CanvasObject`.  If someone was hacking in their own transforms into the chain then yes, that would be affected.  But that is all fairly new code to ginga and still all fairly "internal".

ginga/ImageView.py
ejeschke: Good point. I'll refactor this a bit.

PR #520, opened 2017-06-30
Ensure SAMP and FITS compatibility with Astropy 2.0

pllim:
This is needed as a result of astropy/astropy#6213 and astropy/astropy#6203. Otherwise, you will get deprecation warnings when Astropy 2.0 is released.


ejeschke:
Thanks, @pllim!

PR #529, opened 2017-07-10
Allow text entry widgets to be resized

austinbeauch:
Was trying to resize some widgets and found the set_length method wasn't working. The size was being hard coded to 20 and self.length wasn't being used, perhaps this is what it's supposed to be? Did some manual tests and the resize works when used like

```
catalog = Widgets.TextEntrySet()
catalog.set_length(5)
```


ejeschke:
This seems reasonable.  Merging.

PR #534, opened 2017-07-23
Add license files to manifest

mwcraig:


ejeschke:
@mwcraig, I assume this is needed to get the license files included properly in the conda build?  (Extrapolating from the reference, since there was no description).

mwcraig:
Sorry about that...it is to get the license files into the sdist (which will in turn include it in the conda pkg). It used to be included by this line in the manifest:

```
recursive-include licenses *
```

but once the licenses were moved they were no longer included.

ejeschke:
Sounds good, merging.

PR #541, opened 2017-08-07
Clean-up setup script based on Astropy's

### LINKED ISSUE #540: Clean up setup.cfg
See https://github.com/ejeschke/ginga/issues/539#issuecomment-319491662

pllim:
Fix #540 by comparing how `setup(...)` is defined in Ginga and Astropy. I did not remove all metadata from `setup.cfg` but only changed what seem different, as I don't want to break things that already work. Wouldn't know if this fixes the PyPi metadata until next release.

xref #539


ejeschke:
I'd like to do another release toward mid/end of September, so we can find out then.

Issue #545, opened 2017-08-08
cursession undefined in ImageViewBokeh.py


pllim:
In the `ImageViewBokeh2.update_image()` method, there is `cursession().store_objects(d_src)`, where `cursession` is not defined anywhere. Is this Bokeh magic or some oversight?

ejeschke:
bokeh is not yet a fully supported back end, mostly because I haven't figured out a way yet to connect the event bindings.  `ImageViewBokeh2` class was one possible implementation of the `ImageViewBokeh` class, so it is in there for reference, but not used.  And probably the `cursession()` is from that.  That is some API call from the Bokeh library.

pllim:
> so it is in there for reference, but not used

Maybe it should be commented? Having a broken "live" code can cause some confusing traceback if someone tried to use this class. Just a thought.

Issue #544, opened 2017-08-08
unichr in IRAF.py not Python 3 compatible


pllim:
In `IRAF.py`, the `IRAF_AstroImage.info_xy()` method uses `unichr(...)` function that is not Python 3 compatible.

See https://stackoverflow.com/questions/2352018/cant-use-unichr-in-python-3-1

ejeschke:
I'm thinking about "decommissioning" the IRAF plugin by removing it from the default globals and maybe just dropping it into the "experimental" folder. When I wrote it I thought that maybe it might be useful for folks that are stuck on IRAF.  But my guess now is that most Ginga users do not want to use IRAF and, if they do, they are probably happy enough with the IRAF/ds9 combination.

That's not to say that we shouldn't fix this issue in the meantime.

pllim:
James Turner recently ported IRAF to Python 3, so for now, I agree that this plugin should at least also work on Python 3 until decommissioning. 😄

ejeschke:
Fixed in commit b642336bbebf475bb17d574979953262b3440ed4

pllim:
Thanks!

Issue #543, opened 2017-08-08
Undefined decoded_data, file, socketError in IIS_DataListener.py


pllim:
In the method below:
```python
    def decode_iis(self, data):
        f = file('/tmp/pippo', 'wb')  # <-- Not compatible with Python 3
        f.write(data)
        f.close()
        return (decoded_data)  # <-- Undefined!
```
and also
```python
    if not match:
        # Error
        raise socketError("I don't understand the format of addr IMTDEV: '%s'" % (imtdev))  # <-- Undefined exception
```

For `file()`, see https://stackoverflow.com/questions/112970/python-when-to-use-file-vs-open

ejeschke:
Grepping the code, it doesn't look like `decode_iis()` is used anywhere either in `IIS_DataListener` or `IRAF.py`.  It can probably be removed.

pllim:
What about `socketError`?

ejeschke:
I think it could be changed to `ValueError`.  It's not actually a socket error anyway.

Issue #542, opened 2017-08-08
Undefined variables in Task.py


pllim:
In `WorkerThread` class in `Task.py`:
```python
    def stop(self):
        # Put termination sentinal on queue
        self.queue.put((priority, task))  # <-- priority and task are undefined!
        self.ev_quit.set()
```

ejeschke:
Fixed in aef1e3817efaec00b52733b713cffc2862111a6e.

I think this error probably never showed up because the sentinal `ev_quit` is used to stop the thread pool and this is noticed after the queue wait timeout.

pllim:
Thanks for the quick fix!

PR #546, opened 2017-08-08
Remove bare excepts

### LINKED ISSUE #531: Replace bare except with "except Exception"
Saw some in `TVMask.py` (probably my bad when I wrote it). We should comb all the code for bare excepts and replace them with `except Exception`.

pllim:
Remove bare excepts (fix #531). Also some PEP8 clean-ups. In the process, found #542, #543 (fixed in a subsequent commit here), #544, and #545, which should be addressed separately (or here if you want to as subsequent commits but I need feedback on how to fix them).

I recommend we merge #499 first in case there are conflicts, as it might be easier to resolve them here.

### DIFF
--- a/ginga/rv/plugins/IIS_DataListener.py
+++ b/ginga/rv/plugins/IIS_DataListener.py
@@ -44,13 +44,13 @@ IMT_FBCONFIG      = 0o77
 XYMASK            = 0o77777
 
 MAX_FBCONFIG      = 128             # max possible frame buf sizes
-MAX_FRAMES        = 15              #  max number of frames (start from 0)
-MAX_CLIENTS       = 8               #  max display server clients
-DEF_NFRAMES       = 1               #  save memory; only one frame
-DEF_FRAME_WIDTH   = 512             #  512 square frame
-DEF_FRAME_HEIGHT  = 512             #  512 square frame
+MAX_FRAMES        = 15              # max number of frames (start from 0)
+MAX_CLIENTS       = 8               # max display server clients
+DEF_NFRAMES       = 1               # save memory; only one frame
+DEF_FRAME_WIDTH   = 512             # 512 square frame
+DEF_FRAME_HEIGHT  = 512             # 512 square frame
 
-SZ_LABEL          = 256             #  main frame label string
+SZ_LABEL          = 256             # main frame label string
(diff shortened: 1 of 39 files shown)


ginga/rv/plugins/IIS_DataListener.py
```diff
@@ -697,11 +692,10 @@ def handle_imcursor(self, pkt):
                     if abs(fb.ct.a) > 0.001:
                         sx = int((wx - fb.ct.tx) / fb.ct.a)
                     if abs(fb.ct.d) > 0.001:
-                        sy = int((wy - xt.ty) / fb.ct.d)
+                        sy = int((wy - fb.ct.ty) / fb.ct.d)
```
pllim: @ejeschke , `xt` is undefined, so I made this substitution. Is this correct?

setup.cfg
```diff
@@ -34,3 +34,10 @@ test_suite = ginga.tests.ginga_test_suite
 [entry_points]
 ginga = ginga.rv.main:_main
 ggrc = ginga.misc.grc:_main
+
+[flake8]
+# Ignoring these for now:
+# E265: block comment should start with '#'
+# E266: too many leading '#' for block comment
+# E501: line too long
+ignore = E265,E266,E501
```
pllim: Note: This is mostly for my own sanity. What this does it that if you use `flake8` as your PEP 8 checker, it will find this file and ignore the given error/warning codes here.

ginga/rv/plugins/IIS_DataListener.py
ejeschke: Probably. :smile:

ejeschke:
> I recommend we merge #499 first in case there are conflicts, as it might be easier to resolve them here.

Agreed.

pllim:
The only catch (pun) with removing bare excepts is that it is the wrong thing to do if you also want to catch system exit, keyboard interrupt, and the likes; but if not, then this should be okay.

ejeschke:
I'm strongly against silently ignoring errors except in what are "trivial" cases.  An example is where you are closing a plugin and you are destroying its widgets and there is some error from the canvas.  I think that tends to be the majority of the cases you converted here.

ejeschke:
@pllim, I will resolve this conflict, due to my updates to `Pick`.

PR #547, opened 2017-08-09
Add option to de-select a TreeView item

### LINKED ISSUE #532: Feature Request: Add "deselect_path" to TreeView widget
Currently, `TreeView` widget has a `select_path()` method but no public API to de-select, except to clear the whole tree. I propose a new method named `deselect_path()`. In Qt, it is simply the opposite of `select_path()`, as follows:

```python
class TreeView(...):

    def select_path(self, path):
        item = self._path_to_item(path)
        item.setSelected(True)

    def deselect_path(self, path):
        item = self._path_to_item(path)
        item.setSelected(False)
```

pllim:
Fix #532.

### DIFF
--- a/ginga/gtk3w/Widgets.py
+++ b/ginga/gtk3w/Widgets.py
@@ -936,10 +936,13 @@ class TreeView(WidgetBase):
         treeselection = self.tv.get_selection()
         treeselection.unselect_all()
 
-    def select_path(self, path):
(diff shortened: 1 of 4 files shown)


pllim:
Please merge when you think it can be done. Thanks!

ejeschke:
Ok, merging.

PR #550, opened 2017-08-30
Update font handling

ejeschke:
This updates font handling in Ginga to allow more consistent handling of fonts between widgets and ginga canvases, allows bundled fonts to be used in the default case (with Qt back end) and makes it possible
for the user to load custom fonts.

Some minor updates for matplotlib 2.0 and the "pg" web widgets are included.


ejeschke:
All changes are API backward compatible, so this should have minimal risk to merge.

Issue #396, opened 2016-11-09
Feature Request: Draw shape based on iso-flux contours


pllim:
As requested by Elena Sacchi in Ginga unconference session in JWST Data Analysis Workshop II, it would be nice to draw a shape based on some iso-flux contour.

pllim:
Note: This might overlap a bit with `imexam` functionality. Maybe we should wait to see how that `imexam` plugin works out first (c/c @sosey).

sosey:
I don't think this overlaps too much,  but would need to be implemented in both viewers (ginga and ds9). Are you talking about them hand-selecting anything or figuring out the contours based on some other user input and then just drawing?  And by shape do you mean an astropy/region?

pllim:
I didn't ask for a lot of details. It was just a single-line question asked during the workshop, which is what was recorded above. Naively, I would assume user wants to draw an irregular shape (in Ginga, it would use some kind of Path object, as Ginga has not yet adopted Astropy's region stuff) that follows the selected flux value (maybe +/- certain delta flux). Exactly how that flux is selected (by clicking or entering in a box), I am not sure. I am guessing user would expect something like http://matplotlib.org/examples/pylab_examples/contour_demo.html .

ejeschke:
This could also possibly be done with an overlay, similar to the `Overlays` plugin, but confined to some area around the drawing region.

ejeschke:
How does `matplotlib` represent contours internally? As polygons?

Wondering if you could represent a contour as a compound object made up of filled polygons.

pllim:
Maybe we can go the lazy route and have such a plugin just overlay `matplotlib` contours somehow?

PR #551, opened 2017-09-10
Update Pick plugin to have a more interactive contour plot

ejeschke:
This introduces the capability to conveniently add contour plots on top of ginga canvases.  It requires a recent version of the `scikit-image` package.  If that is not available Pick falls back to using a matplotlib plot.

### DIFF
--- a/ginga/AutoCuts.py
+++ b/ginga/AutoCuts.py
@@ -152,8 +152,8 @@ class Histogram(AutoCutsBase):
             if count < (self.crop_radius ** 2.0) * 0.50:
                 # if we have less than 50% finite pixels then fall back
                 # to using the whole array
-                self.logger.info("too many non-finite values in crop--"
-                                 "falling back to full image data")
+                self.logger.debug("too many non-finite values in crop--"
+                                  "falling back to full image data")
                 data = image.get_data()
         else:
             data = image.get_data()
--- a/ginga/rv/plugins/Pick.py
+++ b/ginga/rv/plugins/Pick.py
@@ -5,14 +5,14 @@
 # Please see the file LICENSE.txt for details.
 #
 import threading
-import numpy
+import numpy as np
 import time
 from collections import OrderedDict
 
 from ginga.gw import Widgets, Viewers
 from ginga.misc import Bunch
-from ginga.util import iqcalc, wcs
-from ginga import GingaPlugin, colors
+from ginga.util import iqcalc, wcs, contour
+from ginga import GingaPlugin, colors, cmap, trcalc
 from ginga.util.six.moves import map
 
 try:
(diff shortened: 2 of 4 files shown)


pllim:
I was excited at this feature, so I checkout this branch and installed it. However, I don't see any contour overlay. Instead I saw these warnings below. I have `skimage` 0.13.0, scipy 0.19.1, numpy 1.13.1,  astropy 3.0.dev, Python 3.5, Qt5, RHEL7 64-bit.

```
** (python:16504): WARNING **: invalid source position for vertical gradient

2017-09-11 15:12:12,375 | W | core.py:5777 (partition) | .../numpy/core/fromnumeric.py:639: UserWarning: Warning: 'partition' will ignore the 'mask' of the MaskedArray.
  a.partition(kth, axis=axis, kind=kind, order=order)

2017-09-11 15:12:12,621 | W | minpack.py:427 (leastsq) | .../scipy/optimize/minpack.py:427: RuntimeWarning: Number of calls to function has reached maxfev = 800.
  warnings.warn(errors[info][0], RuntimeWarning)
```

ejeschke:
The warning that sounds possibly related is the one:
```python
** (python:16504): WARNING **: invalid source position for vertical gradient
```
could you try with another type of image, maybe a more standard catalog image?  And could I get a copy (via Box maybe) of the image that doesn't work?

ejeschke:
I tried it with a few test HST images that I have and it seems ok.

ejeschke:
> However, I don't see any contour overlay.

Are you looking at the "Contour" tab?  With this particular PR it just replaces the plot in the "Contour" tab--it doesn't place an overlay on the main image (although that would be easy to do).

pllim:
@ejeschke , what about with `s1_mips_v0.30_sci.fits` (https://stsci.box.com/s/zstyluc9wcujccquni6mpl7k6tpgjqmk)?

pllim:
> Are you looking at the "Contour" tab?

Oh, I was expecting overlay on the main display! 😅 I did see something legit on the "Contour" tab though, so maybe these warnings did not stop it from actually working.

ejeschke:
> what about with s1_mips_v0.30_sci.fits

Works fine for me.  I just get the warning "UserWarning: Warning: 'partition' will ignore the 'mask' of the MaskedArray.  a.partition(kth, axis=axis, kind=kind, order=order)", but it is coming from deep in numpy and doesn't seem to affect the result.

ejeschke:
> I did see something legit on the "Contour" tab though, so maybe these warnings did not stop it from actually working.

You can zoom the contour like a regular ginga viewer, pan the image, shift contrast, etc.  It is a lot smoother than with the matplotlib widget.

ejeschke:
I'm going to go ahead and merge this.  @pllim, if you are interested in some kind of contour overlay for the channel viewer, let's make a new issue for it.

pllim:
I think there is already a similar issue in #396. 👍 to merging. I can always open follow-up issue if the warnings become too much of a bother.

Issue #563, opened 2017-10-11
Remove "allowed failure" status on PEP8 test on Travis


pllim:
Currently there are about 2700 warnings from `flake8` check on Travis CI. They are simple to fix, despite tedious. Once they are fixed, we can remove the "allowed failure" status of PEP8 check in `.travis.yml`. To make review less painful, this can even be broken into multiple small PRs, addressing one sub-section at a time.

To avoid conflict, this should wait until #548 is merged.

ejeschke:
How does `flake8` handle things that can't be converted, like camel-case APIs to Qt?

pllim:
In that case, a `# noqa` comment at the end of that line of code might work. Either that, or you disable that kind of check entirely by white-listing it inside `setup.cfg`.

ejeschke:
> a # noqa comment at the end of that line of code might work

I think that would be a tremendous amount of annotation, and seeing all the comments are probably more annoying than just seeing the camel-case.  Probably better to use a white listing, IMO.

pllim:
Either white-list the camel-case error/warning code or address #62 .

PR #565, opened 2017-10-13
PEP8 clean-ups of Ginga files not within a sub-directory

pllim:
Address #563 for files that are not inside any sub-directory under `ginga`. There are a few undefined variables that look like actual bugs but their corrections seemed straightforward.


pllim:
Will cherry pick into #585

PR #567, opened 2017-10-13
PEP8 fix for files under cairow sub-directory

pllim:
Address #563 with PEP8 fixes for files under `ginga/cairow` sub-directory.


pllim:
Will cherry pick into #585

PR #568, opened 2017-10-13
PEP8 fix for files under canvas sub-directory

pllim:
Address #563 for files under `ginga/canvas` sub-directory.

### DIFF
--- a/ginga/canvas/types/basic.py
+++ b/ginga/canvas/types/basic.py
@@ -8,16 +8,16 @@ import numpy as np
 
 from ginga.canvas.CanvasObject import (CanvasObjectBase, _bool, _color,
                                        Point as XPoint, MovePoint, ScalePoint,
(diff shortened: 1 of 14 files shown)


ginga/canvas/types/basic.py
```diff
@@ -89,8 +89,6 @@ def draw(self, viewer):
         cx, cy = viewer.get_canvas_xy(x, y)
         cr.draw_text(cx, cy, self.text, rot_deg=self.rot_deg)
 
-        if self.showcap:
-            self.draw_caps(cr, self.cap, cpoints)
```
pllim: Note: I can't figure out where `cpoints` is supposed to come from, so I just removed this line entirely, assuming that this is some copy-and-paste artifact. Is this the right approach?

pllim:
Will cherry-pick into #585.

PR #569, opened 2017-10-13
PEP8 fix for files under cvw sub-directory

pllim:
Address #563 for files under `ginga/cvw` sub-directory.


pllim:
Will cherry-pick into #585.

PR #576, opened 2017-10-14
Fix PEP8 for files under misc sub-directory

pllim:
Address #563 for files under `ginga/misc` sub-directory.

### DIFF
--- a/ginga/misc/Datasrc.py
+++ b/ginga/misc/Datasrc.py
@@ -65,8 +65,8 @@ class Datasrc(object):
         if len(args) == 0:
             return self.remove(self.history[0])
 
-        assert len(args) == 1, \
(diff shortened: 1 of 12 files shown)


ginga/misc/Datasrc.py
```diff
@@ -65,8 +65,8 @@ def pop(self, *args):
         if len(args) == 0:
             return self.remove(self.history[0])
 
-        assert len(args) == 1, \
-               ValueError("Too many parameters to pop()")
```
pllim: Note: I see several cases of doing `assert` only to raise a different exception. Perhaps it is cleaner to use `if` instead, as I demonstrated here. If you want to pursue this for all such cases in this package, it is a separate PR for another day.

pllim:
Will cherry-pick into #585.

PR #577, opened 2017-10-14
PEP8 fix for files under mockw and tkw sub-directories

pllim:
Address #563 for files under `mockw` and `tkw` sub-directories. (I attached `tkw` here because it uses `mockw`)

I must admit that I don't understand what the "mock" stuff is for, but I found a lot of undefined variables, so I tried to either raise an exception or define them. Please review this one carefully and advise what to do for those cases.


pllim:
Will cherry-pick into #585.

PR #579, opened 2017-10-14
PEP8 fix for files in opencl and opengl sub-directories

pllim:
Address #563 for files in `ginga/opencl` and `ginga/opengl` sub-directories.


pllim:
Will cherry-pick into #585.

PR #580, opened 2017-10-14
PEP8 fix for files in pilw sub-directory

pllim:
Address #563 for files in `ginga/pilw` sub-directory.

### DIFF
--- a/ginga/pilw/ImageViewPil.py
+++ b/ginga/pilw/ImageViewPil.py
@@ -6,11 +6,11 @@
 
 from io import BytesIO
 
-import numpy
+import numpy as np
+from PIL import Image
 
 from ginga import ImageView
 
-from PIL import Image
 from . import PilHelp  # noqa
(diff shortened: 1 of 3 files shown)


ginga/pilw/ImageViewPil.py
```diff
@@ -4,13 +4,13 @@
 # This is open-source software licensed under a BSD license.
 # Please see the file LICENSE.txt for details.
 
-from io import BytesIO
```
pllim: Note: This syntax is Python 3 only. Since Python 2 is still supported, I changed it to use `six` here and in some other relevant PRs as I see them. However, we should do a global sweep after these PEP8 PRs are merged to see if there are any remaining `from io import BytesIO`.

pllim: EDIT: Nevermind... I misread the `six` documentation. Will revert this. Sorry for the confusion!

pllim:
Will cherry-pick into #585.

PR #582, opened 2017-10-14
PEP8 fix for files under rv, table, and tests sub-directories

pllim:
Address #563 for files in `ginga/rv`, `ginga/table`, and `ginga/tests` sub-directories.

### DIFF
--- a/ginga/tests/__init__.py
+++ b/ginga/tests/__init__.py
@@ -1,7 +1,0 @@
-import unittest
-import ginga
-
-def ginga_test_suite():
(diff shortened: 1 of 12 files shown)


ginga/tests/__init__.py
```diff
@@ -1,7 +0,0 @@
-import unittest
```
pllim: Note: This is not needed anymore after #558.

pllim:
Will cherry-pick into #585.

PR #583, opened 2017-10-14
PEP8 fix for files in util sub-directory

pllim:
Address #563 for files in `ginga/util` sub-directory.


ginga/util/heaptimer.py
```diff
@@ -26,13 +26,12 @@
 __version__ = '1.0'
```
pllim: Note: Unrelated to this PR but I notice that this file has Apache License. Do we need to declare this in `ginga/licenses`?

pllim:
Will cherry-pick into #585.

PR #586, opened 2017-10-17
Make AstroImage respect set_ioClass and set_wcsClass methods.

mkelley:
The `__init__` method uses the class attributes `ioClass` and `wcsClass` in the method definition for keyword defaults.  The class methods `set_ioClass` and `set_wcsClass` change their resepective class attributes, but the `__init__` method function call is not updated to reflect the change.  Solution: set the `__init__` keywords `ioclass` and `wcsclass` to `None`, test these for `None` first, then test `self.ioClass` and `self.wcsClass` before falling back on the defaults.

### DIFF
--- a/ginga/AstroImage.py
+++ b/ginga/AstroImage.py
@@ -42,7 +42,7 @@ class AstroImage(BaseImage):
         cls.ioClass = klass
 
     def __init__(self, data_np=None, metadata=None, logger=None,
-                 name=None, wcsclass=wcsClass, ioclass=ioClass,
+                 name=None, wcsclass=None, ioclass=None,
                  inherit_primary_header=False):
 
         BaseImage.__init__(self, data_np=data_np, metadata=metadata,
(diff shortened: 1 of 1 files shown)


pllim:
Is this a fix for a particular use case?

ejeschke:
@mkelly, props for using the custom data format hooks!  I'd be interested to hear what kind of data format you are using.

As far as the PR, this looks straightforward enough and I think we can merge.  I've been meaning to revisit the way the custom wcs and loaders are supported, because I think the current method is a bit awkward and could be simpler, but I haven't gotten around to it.

mkelley:
The data I am working with is from the Planetary Data System, NASA's Solar System data archive.  Some of our data is in FITS format, but the rest are essentially simple binary arrays.  In order to make these data more accessible, I am writing a plugin for Ginga.  At a minimum, it would read the meta data (an XML file) and the array, then display it in Ginga with the correct orientation.  If there is enough interest from archive users, then I will expand on that.

Thanks!

ejeschke:
Nice.  Thanks for the info!

PR #609, opened 2017-10-31
Do not load deprecated Matplotlib colormaps

pllim:
Fix #553.

Re-enable auto-loading of `matplotlib` colormaps. However, do not load deprecated colormaps. This is because loading deprecated ones result in duplicate colormaps (you have the deprecated one *and* the new one that is replacing it). Instead, you will see these messages on the terminal when you start Ginga:

```Error adding colormap 'spectral': The spectral and spectral_r colormap was deprecated in version 2.0. Use nipy_spectral and nipy_spectral_r instead.
Error adding colormap 'Vega10': The Vega10 colormap was deprecated in version 2.0. Use tab10 instead.
Error adding colormap 'Vega20': The Vega20 colormap was deprecated in version 2.0. Use tab20 instead.
Error adding colormap 'Vega20b': The Vega20b colormap was deprecated in version 2.0. Use tab20b instead.
Error adding colormap 'Vega20c': The Vega20c colormap was deprecated in version 2.0. Use tab20c instead.
Error adding colormap 'spectral_r': The spectral and spectral_r colormap was deprecated in version 2.0. Use nipy_spectral and nipy_spectral_r instead.
Error adding colormap 'Vega10_r': The Vega10_r colormap was deprecated in version 2.0. Use tab10_r instead.
Error adding colormap 'Vega20_r': The Vega20_r colormap was deprecated in version 2.0. Use tab20_r instead.
Error adding colormap 'Vega20b_r': The Vega20b_r colormap was deprecated in version 2.0. Use tab20b_r instead.
Error adding colormap 'Vega20c_r': The Vega20c_r colormap was deprecated in version 2.0. Use tab20c_r instead.
```

PR #613, opened 2017-11-04
Do not print out error when matplotlib colormap import fails

pllim:
Follow up of #609. I find the printed messages too noisy and there is nothing we can do about them.


pllim:
Thanks!

PR #619, opened 2017-11-21
Changed categories of menus

### LINKED ISSUE #310: Operations menu to handle a lot of local plugins

ejeschke:
This is an alternate to #449.  It allows for setting categories for grouping plugins under the "Operations" menu.

EDIT: Close #449, fix #310


pllim:
Some feedback:

1. The order of the categories seem random to me. Is it better to order them alphabetically?
2. Similar idea as **1** but for plugins under each category.
3. `Help -> Help [G]` seems like an overkill. Perhaps `Help [G]` does not need its own category, but rather just be placed at the bottom of the first level of menus (in place of the "Help" category). This is because it is unlikely that we will have multiple things under "Help" category anyway, right?
4. Do you have plans to add more plugins under "Navigation"? If not, maybe fold `MultiDim` into "Utils"?

For the rest, it is best to seek feedback from other users. Thanks!

ejeschke:
Rebased.

ejeschke:
@pllim, re: 1 & 2, I agree, but wondering if alphabetically is best.  Maybe just put them in the order they are in the lists (or provide another ordering list as a preference?) and thus let there be programmatic control over it?

3: I'm not particular except that if we are going to standardize on having global plugins have an indicator of some sort then `Help` should have that as well, to minimize user surprise.

4: I'm thinking there may be more.  `MultiDim` is an important plugin, so thinking that it probably shouldn't be buried under `Utils`, since a lot of misc things might end up there.

pllim:
@ejeschke , sure, programmatic control is fine by me. No problem with your plan for Item 4 as well. As for Item 3, feedback from users would be helpful to determine the best course of action. Thanks!

ejeschke:
@pllim, have a look at the latest update to this PR.  This version unifies the global and local plugin lists, so that the overall list is ordered.  If you are manipulating the module lists directly from `stginga` you might need to update that accordingly, but if you are using the API (such as it was) it will continue to work as before.

The "Operations" menu is now ordered alphabetically.  I'm not entirely convinced this is the right way to go for the default, but it illustrates the programmatic ordering.  The menu bar "Plugins" menu (for starting global plugins) is not ordered correctly yet (because it doesn't use sub-menus). But perhaps the menu could be eliminated, made to mirror the `Operations` menu, or something else.

pllim:
@ejeschke , thanks, I'll have a look. In the meantime, you might want to rebase to resolve the merge conflict.

pllim:
@ejeschke , I like this "Operations" ordering! Though it is certainly a personal taste on my part, given that I already know what plugins are there, so it is easier for me to find them alphabetically. It is okay if it is not the default as long as I can tweak it so.

As for the "Plugins" menu, I have always find it useful to quickly find and launch a global plugin. However, I am open to alternatives as well.

Thank you!

ejeschke:
Rebased.

ejeschke:
@pllim, "Plugins" menu is now identical to that in `Operations`.  I suppose this may be the least confusing way to organize it--it starts both local and global plugins.

ejeschke:
Anything left to do on this?

pllim:
Please let me have one more look at the new Plugins menu first. Thanks for your patience!

pllim:
@ejeschke , now that I see the "Plugins" for myself, I wonder if it is needed at all. Is "Operations" alone enough? What is the benefit of having two exactly same menu under different places?

Also, when I provide an extra `--plugins=Imexam` argument, "Imexam" local plugin appears at the bottom under "Help". Is that desirable, or should an option be provided to put it in a category?

Meanwhile, when I provide `--modules=MyGlobalPlugin`, the custom global plugin also appear at the bottom like "Imexam" **and** it does not have a "[G]" to indicate that it is a  global plugin.

When I provide both those arguments, they appear in the order of "MyGlobalPlugin", followed by "Imexam", thus breaking the alphabetical scheme.

You mentioned earlier that whether the order is alphabetical or not is configurable, but I don't see a way to do that.

Thank you!

ejeschke:
I'll take a look at handling plugins imported via `--plugins` or `--modules`.  I suspect we can run a user-definable sort function on the plugins list, which might default to alphanumeric.

ejeschke:
@pllim, please try this again.  Now it should sort plugins in the menus no matter how they are added.  I also added a "[G]" suffix for plugins that are added via `--modules`.

ejeschke:
> I wonder if it is needed at all. Is "Operations" alone enough? What is the benefit of having two exactly same menu under different places?

That's a fair question.  I'd say that it's possible that someone might use a configuration that doesn't use the `Operations` plugin.  If so, then unless we have the menubar menu there would be no way to invoke a plugin unless they reimplement something like `Operations`. 

> Also, when I provide an extra --plugins=Imexam argument, "Imexam" local plugin appears at the bottom under "Help". Is that desirable, or should an option be provided to put it in a category?

It should now be sorted into order, although since there is no category information, it will show up at top level.  One way around this is to allow the plugin class to define a category, that can be overridden.  What do you think of that idea?

> Meanwhile, when I provide --modules=MyGlobalPlugin, the custom global plugin also appear at the bottom like "Imexam" and it does not have a "[G]" to indicate that it is a global plugin.

"[G]" suffix added.  category could be added as described above--in the class.

pllim:
> One way around this is to allow the plugin class to define a category, that can be overridden. What do you think of that idea?

I was thinking of something simpler... For example, if user provides something via `--modules` or `--plugins`, simply group them into "User" or "Custom" category (depending on where you want it to appear in alphabetical order).

pllim:
Also, with the current state of this PR, I see the following errors on the terminal:
```
2017-12-19 11:39:43,727 | E | Control.py:319 (add_local_plugin) | Unable to load local plugin 'TVMask': 'GingaShell' object has no attribute 'add_operation'
2017-12-19 11:39:43,731 | E | Control.py:345 (add_global_plugin) | Unable to load global plugin 'WCSMatch': 'GingaShell' object has no attribute 'add_operation'
```
There are a lot more but they are pretty much repetitive.

ejeschke:
> Also, with the current state of this PR, I see the following errors on the terminal

@pllim, is this trying with just ginga or with `stginga`?  If using former, could you confirm that if you do a `git -xdf clean` and reinstall that you don't get the errors?

ejeschke:
> simply group them into "User" or "Custom" category

That's a good idea.  At least it's probably sufficient at the present time and then maybe we can think about how to implement the category and other meta data for a future PR.

ejeschke:
Added code to assign plugins loaded from the command line to the "Custom" category.

ejeschke:
> Also, with the current state of this PR, I see the following errors on the terminal

@pllim, my bad!  I just realized that I wasn't checking my error log far enough back to see the errors.  Somehow the program still registered the plugins ok.  Pushing a fix.

ejeschke:
@pllim, looks like Travis is happy now.  Can you give this another whirl and let me know how it works out?

ejeschke:
Okey-doke. Merging!

PR #621, opened 2017-11-30
New change notification

### LINKED ISSUE #602: MultiDim unexpectedly inserts entries into ChangeHistory
Workflow to reproduce the problem:

1. Start `MultiDim` local plugin.
2. Start `ChangeHistory` global plugin.
3. Load a data cube.
4. With `ChangeHistory` ("History" tab) in view, click on "next slice" (double-arrow) button on `MultiDim` and watch a new entry magically appears on `ChangeHistory` for that image.
5. Click on "next slice" again or "prev slice" to see another such entry.

Example screenshot:
screenshot from 2017-10-24 14-22-46

I suspect this is caused by either `block_callback('modified')` or `unblock_callback('modified')` in `MultiDim.py`. I just never caught this before until now because I rarely start up `ChangeHistory` with data cubes, but I think this has been happening for a while.

This bug is harmless but confusing, in the sense that it does not actually change the input data cube, but merely reported that it has done so.

ejeschke:
This is a [fix for Ginga issue 602](https://github.com/ejeschke/ginga/issues/602).

It introduces a new API call for plugins to register change history messages when modifying images in place.

EDIT: Fix #602 , close #620

### DIFF
--- a/ginga/rv/Channel.py
+++ b/ginga/rv/Channel.py
@@ -5,7 +5,6 @@
 # Please see the file LICENSE.txt for details.
 #
 import time
-from datetime import datetime
 
 from ginga.misc import Bunch, Datasrc, Callback, Future
 
@@ -98,7 +97,6 @@ class Channel(Callback.Callbacks):
 
         if was_not_there_already:
             channel.datasrc[imname] = image
-            image.add_callback('modified', channel._image_modified_cb)
 
             if not silent:
                 self.fv.gui_do(channel._add_image_update, image, info)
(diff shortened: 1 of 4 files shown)


ejeschke:
@pllim, this should be identical to what you've already tested, but maybe check it briefly to make sure there are no surprises? Tnx.

pllim:
Using this PR, I get this error when I try to create a new mosaic by dragging and dropping an image from `FBrowser`:
```
Cannot open image: no value for metadata key 'path'
```
However, there is no associated traceback on terminal, just that message on `Errors`. Do you see the same thing on your side?

ejeschke:
> Do you see the same thing on your side?

No. It is working for me.  Is there a particular image set I can try?

ejeschke:
I tried a GMOS image, a MegaCam image and a Suprime-Cam image.  No errors.

ejeschke:
Ok, Tnx. Merging.

PR #624, opened 2017-12-05
Update mode change

ejeschke:
This adds a special "meta" mode, which is an intermediate mode used to change modes.  In addition, the standard modes are all changed to be started from meta mode.  Under the new scheme, you press Space to invoke meta mode, then press the mode key that then invokes the mode you want.

The primary benefit of this modification is that it frees many key bindings that can be used for other things, and allows more keys to be bound within a mode.

The mode indicator has been updated to show a different color for the meta mode and also to show an indicator for all mode types, not just "locked" (L) and "softlock" (S), but also "held" (H) and "oneshot" (O).

This change affects the standard key bindings--documentation on quickref and manual has been updated.  The old behavior can be retained by simply installing the appropriate bindings.cfg file to $HOME/.ginga.

### DIFF
--- a/doc/WhatsNew.rst
+++ b/doc/WhatsNew.rst
@@ -8,6 +8,11 @@ Ver 2.7.0.dev
 - Fix for broken polygon containment test
 - Addition of configurable zoom handlers for pan gestures
 - Fix for some broken tests under python 2.7
+- Update to mode handling via keyboard shortcuts
+  - addition of a new "meta" mode used primarily for mode switching
+  - most modes now initiated from meta mode, which frees up keys
+    for other uses
+  - see quick reference for details on how the new bindings work
 
 Ver 2.6.6 (2017-11-02)
 ======================
--- a/doc/manual/operation.rst
+++ b/doc/manual/operation.rst
@@ -90,16 +90,20 @@ For example:
 
 When zoomed in, panning is enabled. Panning takes two forms:
 
-1) *Proportional panning* or "drag panning" pans the image in direct proportion to the distance the mouse is moved. You can think of this as dragging the image canvas in the direction you want to move it under the window portal. To utilize a proportional pan, Ctrl-drag the canvas, or press
-"q" to go into pan mode, and then drag the canvas.
+1) *Proportional panning* or "drag panning" pans the image in direct
+   proportion to the distance the mouse is moved. You can think of this
+   as dragging the image canvas in the direction you want to move it
+   under the window portal. To utilize a proportional pan, Ctrl-drag the
+   canvas, or press Space followed by "q" to go into pan mode, and then
+   drag the canvas. 
 
-2) *Free panning* allows scrolling around the entire image by mapping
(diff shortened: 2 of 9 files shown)


ejeschke:
For possible inclusion in v2.7 or beyond

pllim:
So, to test this, I need to rename my current `~/.ginga/bindings.cfg`, is that correct?

ejeschke:
Yes, just move it out of the way so you pick up the defaults.  Or substitute one of the updated bindings files in the .../examples/bindings folder.

pllim:
@ejeschke , I am pretty unadventurous when it comes to key shortcuts and modes (I mainly stick to clicking the GUI buttons); This looks cool and useful but I am unable to provide a more concrete review otherwise.

My concern, given that my users with laptops would use shortcuts more than I, is backward compatibility. So, if they still have the old bindings file in their `~/.ginga` but using Ginga with these changes, what is going to happen? Will I get a lot of, "This used to work this way but now it does not work or it does something completely unexpected" questions? Or will Ginga be smart enough to issue a warning/error that says, "You are using old bindings that will no longer work as expected, please do such and such"? Or are my concerns unfounded? Please advise.

ejeschke:
Reference viewer case:
If the users have an old bindings file in their $HOME/.ginga, it should behave as before (did you try that?).  The question is whether to change the default for users that *don't* have a bindings file.  The PR as it stands does change the key bindings, but only for the mode switching keys.  I'm not sure how many users use those (I use them a lot, but I'm particular about being efficient with the keyboard shortcuts).  It doesn't affect things for the GUI buttons.

Bare widget case:
For users using the bare widget in their code, unless they are reading a bindings file and overriding the default they would get the new default mode key behavior.

I was assuming that we would announce that rather prominently in the release notes.

Another possibility for this PR is to add a commit that lets us merge this capability but keep the old bindings as default (those users that want to use the functionality right away can create their own bindings file), and then at some specified version in the future switch to the new default bindings.  At that time users will always be able to get the old bindings by copying the example bindings file from an older version.  

The problem is that the earlier design choice to map key strokes to modes means that a lot of keys are reserved/constrained in certain ways that the new scheme avoids through the intermediate mode. Sometimes to make progress you have to make a change that affects default backward compatibility. 
 At least in this case users can "opt out" of the new scheme in order to get the backward compatibility.

pllim:
Okay, that sounds sensible. If it is prominent in the release notes, then at least I can just point users to the notes. I agree that progress should not be impeded. Thank you for the clarifications!

> did you try that?

No. I don't use shortcuts much, so I wouldn't be able to tell the difference anyway. :sweat_smile:

ejeschke:
> If the users have an old bindings file in their $HOME/.ginga, it should behave as before (did you try that?). 

I tested it.  It works.

I think I will merge with the changes to the mode initiation keys then.  I will include a prominent note in the release notes for 2.7.

ejeschke:
Merging.

PR #625, opened 2017-12-05
Drawing to use new notification for ChangeHistory

pllim:
Follow up of #621.

### DIFF
--- a/ginga/rv/plugins/ChangeHistory.py
+++ b/ginga/rv/plugins/ChangeHistory.py
@@ -18,7 +18,7 @@ History should stay no matter what channel or image is active.
 New history can be added, but old history cannot be deleted,
 unless the image/channel itself is deleted.
 
-The ``redo()`` method picks up a ``'modified'`` event and displays
+The ``redo()`` method picks up an ``'add-image-info'`` event and displays
 related metadata here. The metadata is obtained as follows::
 
         channel = self.fv.get_channel_info(chname)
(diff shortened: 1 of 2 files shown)


ginga/rv/plugins/ChangeHistory.py
```diff
@@ -18,26 +18,24 @@
...
-        # This issues the 'modified' callback and sets the timestamp
+        # This changes the data buffer
         image.set_data(new_data, ...)
-        # Manually add the description
-        chname = self.fv.get_channel_name(self.fitsimage)
-        channel = self.fv.get_channel_info(chname)
-        iminfo = channel.get_image_info(image.get('name'))
-        iminfo.reason_modified = 'Something was done to this image buffer'
+        # Add description for ChangeHistory
+        info = dict(time_modified = datetime.utcnow(),
```
ejeschke: I think the spaces around the "=" will trigger the flake8 failure.

pllim: That is just the docstring, but yes, we should give people a good example. I'll fix it when I get the chance. Thanks!

ejeschke: No, in the dict(), right?

pllim: This is a code snippet *inside* a docstring, that is why `flake8` check is not catching it, but I agree it should be fixed.

ejeschke: Sorry, I guess I'm not looking at the full code, just the patch...

pllim:
@ejeschke , it is done. Thanks for the review!

ejeschke:
Ok, merging.

pllim:
Thanks!

PR #626, opened 2017-12-17
Efficiency improvement for Thumbs plugin

ejeschke:
This changes the canvas drawing logic for the Thumbs plugin so that it manages the thumbnails canvas much more efficiently, resulting in improved usability and responsiveness.  The main improvement is that it calculates which thumbnails should be visible based on the displayed area of the canvas and adds/removes thumbs as necessary to only show that set.

### DIFF
--- a/ginga/rv/plugins/Thumbs.py
+++ b/ginga/rv/plugins/Thumbs.py
@@ -404,9 +404,8 @@ class Thumbs(GingaPlugin.GlobalPlugin):
                 bnch = self.thumb_dict[thumbkey]
                 path = bnch.info.path
                 placeholder = bnch.image.get_image().get('placeholder', False)
-                self.logger.info("[%s] build missing %s" % (placeholder,
-                                                            str(thumbkey)))
-
+                if placeholder:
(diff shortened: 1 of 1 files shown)


ejeschke:
It would be nice to have this in v2.7 ...

Makes a difference when large numbers of thumbnails are present.

ejeschke:
@pllim, if you don't find any significant issue with this it probably would be good to merge earlier rather than later.  It's basically an isolated efficiency improvement for `Thumbs`.

ejeschke:
> One question: Does this take account of when user manually resize the Ginga window?

Yes.  The layout is recalculated (as before) when the user resizes the window.  But the mechanism of drawing the window is different, in this PR.  Previously, all the icons would be added to the canvas and the viewer would only bother to draw some of them because the others are not visible.  But there still is some overhead associated with adding all the icons, and then determining that they are not visible on redraws.  With this PR, because only the icons that are actually visible are maintained on the canvas, the redraws are snappier.  It's only really noticeable if you have a lot of thumbs.  But it makes for a better scrolling experience in that case.

ejeschke:
Merging.

Issue #482, opened 2017-05-15
Feature Request: Save current state of Ginga session


pllim:
Someone asked if it is possible to save the state of the current Ginga session, which includes but not limited to:

1. Ginga layout like number of channels, channel names, and arrangement.
2. Images opened, pointing, scaling, stretched, etc.
3. Plugins being used and their parameters.

The use case was to recreate some pain-stakingly hand-created chart with object of interest highlighted in 8 different observations using different telescopes.

I told the user that Ginga might not be the best solution for such a thing (i.e., maybe fully script it using `matplotlib`, `reproject`, etc) but I'll just open an issue anyway.

ejeschke:
I actually have this on my wish list as well.  However, I think the session parameters that are saved would be somewhat restricted to make it practical.  Maybe something like 1) only or 1) plus names and locations of images.  

I agree that probably something should be scripted to produce a publication plot.  They could certainly script a ginga widget if PNG output is acceptable.

pllim:
> They could certainly script a ginga widget if PNG output is acceptable.

In that case, an example for people to start with would be nice. The use case presented to me was the following:

1. Display four squares centered around object of interest. Each square (image) have different contrast/stretch to highlight the object optimally. Each image is from a different telescope/instrument, hence different pixel scale, background counts, pointing, etc; Using `WCSMatch` can align them but something else need to scale them to match each other.
2. Draw a circle around object of interest.
3. Save to file.

(Might have forgotten some small details...)

PR #639, opened 2018-03-13
Profile support

ejeschke:
This PR does two things: 1) refactors the way `RGBMap` objects manage their state, so that they can share the settings object of a ginga viewer widget; and 2) unifies the format of an "image profile" with a viewer `SettingGroup` object.  This simplifies the management of state between the viewer and the RGB mapper, but more importantly, opens up new possibilities for more advanced use cases for profiles.

Until now, profiles were only objects attached to the metadata of a `BaseImage`-subclassed object.  Using the Preferences (under the "Remember" category), certain state could be saved and restored to the viewer when moving between different images. Under this PR, basically almost all state from the viewer (meaning a channel image viewer in the context of a reference viewer) can be captured in a profile and saved and restored.  This opens up the interesting possibility of being able to manage profiles explicitly from the user's perspective (save from viewer, restore to viewer, attach to images).  Furthermore, using the sharing capabilities of the settings object, any parts of a profile could be shared with another profile, allowing fairly interesting setups in the reference viewer (think `WCSMatch`, but more powerful).  For example, using this it should be straightforward to implement the ds9 "lock" features, but with even more control by the user over what is "locked" (shared between channels).

### DIFF
--- a/ginga/rv/plugins/Preferences.py
+++ b/ginga/rv/plugins/Preferences.py
@@ -304,7 +304,7 @@ class Preferences(GingaPlugin.LocalPlugin):
         self.pancoord_options = ('data', 'wcs')
         self.sort_options = ('loadtime', 'alpha')
 
-        for key in ['color_map','intensity_map',
+        for key in ['color_map', 'intensity_map',
                     'color_algorithm', 'color_hashsize']:
             self.t_.get_setting(key).add_callback(
                 'set', self.rgbmap_changed_ext_cb)
@@ -1123,7 +1123,7 @@ class Preferences(GingaPlugin.LocalPlugin):
         params = self.t_['autocut_params']
         # NOTE: use gui_do?
         # TODO: set the params from this tuple
-        params_d = dict(params)
+        params_d = dict(params)   # noqa
         #self.ac_params.params.update(params_d)
         #self.ac_params.params_to_widgets()
 
--- a/ginga/rv/plugins/Preferences.py
+++ b/ginga/rv/plugins/Preferences.py
@@ -274,12 +274,11 @@ The "Num Images" setting specifies how many images can be retained in
 buffers in this channel before being ejected.  A value of zero (0) means
 unlimited--images will never be ejected.  If an image was loaded from
 some accessible storage and it is ejected, it will automatically be
-reloaded if the image is revisited by navigating the channel.  A setting
-of 1 to 4 is typical.
+reloaded if the image is revisited by navigating the channel.
 
 The "Sort Order" setting determines whether images are sorted in the
 channel alphabetically by name or by the time when they were loaded.
-This principally effects the order in which images are cycled when using
+This principally affects the order in which images are cycled when using
 the up/down "arrow" keys or buttons, and not necessarily how they are
 displayed in plugins like "Contents" or "Thumbs" (which generally have
 their own setting preference for ordering).
--- a/ginga/rv/plugins/Preferences.py
+++ b/ginga/rv/plugins/Preferences.py
@@ -268,6 +268,40 @@ thumbnail for images loaded into that channel.  In cases where many
 images are being loaded into a channel frequently (e.g., a low frequency
 video feed), it may be undesirable to create thumbnails for all of them.
 
+**General Preferences**
+
+The "Num Images" setting specifies how many images can be retained in
+buffers in this channel before being ejected.  A value of zero (0) means
+unlimited--images will never be ejected.  If an image was loaded from
+some accessible storage and it is ejected, it will automatically be
+reloaded if the image is revisited by navigating the channel.  A setting
+of 1 to 4 is typical.
+
+The "Sort Order" setting determines whether images are sorted in the
+channel alphabetically by name or by the time when they were loaded.
+This principally effects the order in which images are cycled when using
+the up/down "arrow" keys or buttons, and not necessarily how they are
+displayed in plugins like "Contents" or "Thumbs" (which generally have
+their own setting preference for ordering).
+
+The "Use scrollbars" check box controls whether the channel viewer will
+show scroll bars around the edge of the viewer frame.
+
+**Remember Preferences**
+
+When an image is loaded, a profile is created and attached to the image
+metadata in the channel.  These profiles are continuously updated with
+viewer state as the image is manipulated.  The "Remember" preferences
+control which parts of these profiles are restored to the viewer state
+when the image is navigated to in the channel:
+* "Restore Scale" will restore the zoom (scale) level
+* "Restore Pan" will restore the pan position
+* "Restore Transform" will restore any flip or swap axes transforms
+* "Restore Rotation" will restore any rotation of the image
+* "Restore Cuts" will restore any cut levels for the image
+* "Restore Scale" will restore any coloring adjustments made (including
+  color map, color distribution, contrast/stretch, etc.)
+
 """
 import math
 
@@ -786,7 +820,7 @@ class Preferences(GingaPlugin.LocalPlugin):
         b.sort_order.set_tooltip("Sort order for images in channel")
 
         scrollbars = self.t_.get('scrollbars', 'off')
-        self.w.use_scrollbars.set_state(scrollbars != 'off')
+        self.w.use_scrollbars.set_state(scrollbars in ['on', 'auto'])
         self.w.use_scrollbars.add_callback('activated', self.set_scrollbars_cb)
         b.use_scrollbars.set_tooltip("Use scrollbars around viewer")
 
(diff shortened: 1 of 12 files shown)


ejeschke:
Ping @pllim.  This is a WIP but should in it's current state be backward compatible with current master in terms of API and user capabilities.

I'd be interested in kicking around ideas for use-cases for profiles from the user's perspective.

ejeschke:
Currently not passing flake8 check, but that will be all be fixed up before any merge.

pllim:
There was a use case in #482

ejeschke:
Rebased.

ejeschke:
@pllim, I think this is ready for review. After some reflection, I think this might not address all the use cases I had in mind originally, but I think it is still a very worthwhile PR, as it cleans up some messy details by unifying the settings for the RGB mapper object and the viewer object.

No user-facing API changes. Tests passing, flake8 passing.

pllim:
As discussed during tag up, I would wait for a bit before reviewing this. Please ping me again when you need a review. Thanks!

ejeschke:
Actually, this is the one that is ready to go.  Can you give it a quick look over?  It's #649 we can wait on.

pllim:
Oh, ops! OK...

pllim:
Attached is the offending `channel_Image.cfg`. I had to made it `.txt` so GitHub would let me upload.

[channel_Image.cfg.txt](https://github.com/ejeschke/ginga/files/2279236/channel_Image.cfg.txt)

ejeschke:
Rebased.

ejeschke:
> 2018-08-10 15:27:15,552 | E | ModuleManager.py:89 (load_module) | Failed to load module 'Profiles': > No module named 'Profiles'
> 2018-08-10 15:27:15,552 | E | Control.py:328 (add_global_plugin) | Unable to load global plugin 'Profiles': No module named 'Profiles'

Fixed this. I was working on a `Profiles` plugin, which I may eventually get around to adding. It would allow you to manage different profiles, save them, load them, etc.  For now I've removed the reference to it.

> 2018-08-10 15:27:15,637 | W | Settings.py:245 (load) | Error loading '~/lim/.ginga/channel_Image.cfg' at line 49: unexpected EOF while parsing (<unknown>, line 1)

For this one, could you remove your `~/lim/.ginga/channel_Image.cfg` file and regenerate it by saving preferences for the `Image` channel?  This PR fixes an error with multi-line values when arrays are saved in settings files.

ejeschke:
> Also, I don't see any addition to the documentation to expose this new functionality to users.

Added some documentation to the `Preferences` plugin class docstring.  There really isn't any major user-facng change over the non-PR version except for the ability to now save all of the settings having to do with coloring an image in the profile and get restored correctly (so there is one additional checkbox under the "Remember" settings).

ginga/rv/plugins/Preferences.py
```diff
@@ -268,6 +268,40 @@
...
 video feed), it may be undesirable to create thumbnails for all of them.
 
+**General Preferences**
+
+The "Num Images" setting specifies how many images can be retained in
+buffers in this channel before being ejected.  A value of zero (0) means
+unlimited--images will never be ejected.  If an image was loaded from
+some accessible storage and it is ejected, it will automatically be
+reloaded if the image is revisited by navigating the channel.  A setting
+of 1 to 4 is typical.
```
pllim: Hmm. On the contrary, I recommend my users to set it to unlimited. This is because in their use case, they need to create mosaic or modify the buffer (e.g., background subtraction plugin). In those cases, they might lose the mosaic or modified images if it is limited. I recommend a deeper discussion about when to set buffer limit to what, if feasible.

ginga/rv/plugins/Preferences.py
```diff
@@ -268,6 +268,40 @@
...
+
+The "Sort Order" setting determines whether images are sorted in the
+channel alphabetically by name or by the time when they were loaded.
```
pllim: Perhaps also note that the order they are loaded is random due to multi-threading (it is, right?). Some users have asked about how things are loaded in Ginga.

ginga/rv/plugins/Preferences.py
```diff
@@ -268,6 +268,40 @@
...
+This principally effects the order in which images are cycled when using
```
pllim: "effects" -> "affects"

ginga/rv/plugins/Preferences.py
```diff
@@ -268,6 +268,40 @@
...
+
+The "Use scrollbars" check box controls whether the channel viewer will
+show scroll bars around the edge of the viewer frame.
+
+**Remember Preferences**
+
+When an image is loaded, a profile is created and attached to the image
+metadata in the channel.  These profiles are continuously updated with
+viewer state as the image is manipulated.  The "Remember" preferences
+control which parts of these profiles are restored to the viewer state
```
pllim: Perhaps also an explanation of what would happen if restoring a previous state becomes impossible. For example, pan is saved for Image A, but it is an invalid pan (is such a thing possible?) for Image B, which is loaded instead the next time.

pllim:
Unfortunately, I still see these errors:
```
2018-08-11 12:43:17,644 | W | Settings.py:245 (load) | Error loading
  '~/.ginga/channel_Image.cfg' at line 49:
  unexpected EOF while parsing (<unknown>, line 1)
2018-08-11 12:43:24,637 | W | Settings.py:245 (load) | Error loading
  '~/.ginga/channel_Image0.cfg' at line 48:
  unexpected EOF while parsing (<unknown>, line 1)
```
I had 2 images in `Image` channel and 1 in `Image0` channel. I chose to restore Pan for the former, and Scale and Cuts for the latter. The `cfg` files are attached.
[channel_Image.cfg.txt](https://github.com/ejeschke/ginga/files/2280394/channel_Image.cfg.txt)
[channel_Image0.cfg.txt](https://github.com/ejeschke/ginga/files/2280395/channel_Image0.cfg.txt)

ginga/rv/plugins/Preferences.py
ejeschke: Sounds like we should devote a section in the manual to this. I agree that it makes sense if you are working with modified images in memory, but for the user who is mostly looking through data they are loading from disk there is no downside and a very big upside to having a limited size--because after only a few large images the machine will become slow and grind to a halt as the memory fills up.

ginga/rv/plugins/Preferences.py
ejeschke: Yes, if you load several images simultaneously and the sort order is set to load time then the order can appear random due to the use of threading.

ginga/rv/plugins/Preferences.py
ejeschke: So for your example, if you set a pan position in image A (lets say its a large one, and you pan out to the edge of it) and then view a new small image B, the new image does not have an existing profile. If the autocenter feature is turned to "off" then the pan position is just left where it was when you were viewing the large image and you may see nothing, until you press 'c' or center button in the toolbar. This is normal behavior without having anything to do with profiles.

With "restore pan" checked and you visit image B and center it, then go back to image A, you will be returned to the edge of the image A where you last had panned in it.  When you go back to B now, it will pan you to the center of it.

The point here is that the profiles don't change anything fundamentally with respect to having invalid settings. If you have autocuts turned off and you visit an image for which the existing cuts are invalid the image will usually look wrong.  The profiles can help in that if you constantly find yourself redoing some setting when going back and forth between images in a channel you might be able to avoid that by using some of the Remember settings.

This PR only changes things in that it adds the ability to restore the color map, color distribution and contrast changes (basically everything devoted to coloring the image) from the profile as well. So if you change the color map, or contrast for image A and then change it to something else for image B, then return to image A (*and* you have the "Restore color map" checkbox turned on) then it will restore the color settings exactly as they were when you were viewing that image.

ejeschke:
I fixed the typo, removed the sentence about "typical" setting for `numImages` and fixed the parsing issue for reading in the settings.  Could you try the PR again when you get the chance? Tnx!

pllim:
Thanks for the quick fix! Errors are gone, but "restore color map" does not seem to have any effect. What did I do wrong? My workflow:

* Open image in default channel.
* Create a new channel (`Image0`) and move it there.
* Open `Preferences` for `Image0`.
* Change colormap from gray to some colorful one.
* Check "restore color map" option.
* Click "Save Settings" button.
* Quit Ginga.
* Start Ginga again and repeat creating channel and moving the same image to channel.

I briefly saw the colorful color map in the color bar between image display and cursor display. But then it quickly reverted back to gray. 🤔

ejeschke:
This experiment may also be helpful in understanding what is going on:

* Start Ginga and create an additional channel ("Image0")
* Open Preferences for channel Image0 and check "Restore color map".
* Make sure the color map for Image0 channel is "gray".  Close Preferences.
* Add add image ("A") to the channel "Image0"; it should appear gray.
* Open image ("B") in channel "Image".
* Open Preferences for Image.
* Change colormap from gray to some colorful one.
* Go to Contents and move Image B from channel Image to channel Image0
* Double click image B in Contents to display it in Image0
* In channel Image0 you should see the B image with its colorful color map

ejeschke:
> Perhaps this could be better clarified in the documentation.

Looks like we are going to need a good explanation.  By default, all "Restore" options are off.  So I envision we show how to use it with some use cases:
* A person is tired of always manually resetting cuts when they go back and forth between images. Solution: check "Restore cuts".
* A person is tired of resetting contrast when going back and forth between images. Solution: check "Restore color map"
* A person is tired of pressing the "align N-E" button when they go back to an image. Solution: check "Restore rotation" and "Restore transform".

Let's work on that in the documentation PR!

ejeschke:
Rebased.

Issue #644, opened 2018-03-15
GInga takes ages to load


olebole:
(Python 3) When I start ginga, it shows a blank screen for ~30 seconds, before it is actually ready to go.
```
2018-03-15 10:25:15,976 | I | main.py:364 (main) | Chosen toolkit (qt5) family is 'qt'
2018-03-15 10:25:16,949 | D | Task.py:1012 (startall) | startall called
2018-03-15 10:25:16,949 | D | Task.py:1037 (startall) | starting threads in thread pool
2018-03-15 10:25:16,950 | D | Task.py:926 (taskloop) | Starting worker thread loop.
[...]
2018-03-15 10:25:52,975 | D | ImageView.py:1522 (get_rgb_object) | times: total=0.0047
2018-03-15 10:25:52,975 | D | ImageViewQt.py:172 (render_image) | redraw pixmap=<PyQt5.QtGui.QPixmap object 
2018-03-15 10:25:52,975 | D | ImageViewQt.py:175 (render_image) | drawing to pixmap
2018-03-15 10:25:52,975 | D | ImageViewQt.py:149 (_render_offscreen) | data shape is 664x664x4
2018-03-15 10:25:52,986 | D | ImageViewQt.py:252 (update_image) | updating window from pixmap
2018-03-15 10:25:52,986 | D | ImageView.py:1265 (redraw_now) | widget 'colorbar' redraw (whence=0) delta=0.0
```
The full log is attached.
[ginga.log](https://github.com/ejeschke/ginga/files/1814732/ginga.log)

pllim:
How did you start it exactly? And which versions of Ginga and Astropy do you have? Are you doing something custom in your `ginga_config.py` on start-up? What data are you loading? I see some `get_rgb_object` calls and mention of a 664x664x4 data cube.

olebole:
I started it just with
```
$ ginga --log-level=9 --log=ginga.log
```
(also happens without any command line arguments)

Versions:
* GInga 2.7.0
* astropy 3.0
* Python 3.6.4

No customization; not loading anything. There was a `~/.ginga` when I recorded the log file, but the delay happens also when I delete it.

pllim:
Using the following versions, mine started up in 3 seconds. Log: 
[ginga.log](https://github.com/ejeschke/ginga/files/1815772/ginga.log)

* RHEL7 64-bit
* Python 3.6.3
* Numpy 1.14.1
* Scipy 1.0.0
[...]

PR #646, opened 2018-03-16
Add experimental to MANIFEST

pllim:
As part of #644 investigation, I found that "experimental" folder is not distributed via PyPI tarball, so I am adding it here. ~~I think this will be useful as people who do `pip install` will still be able to use experimental plugins if they so choose.~~ (see below)

### DIFF
--- a/MANIFEST.in
+++ b/MANIFEST.in
@@ -12,6 +12,7 @@ recursive-include doc *
 recursive-include licenses *
 recursive-include cextern *
 recursive-include scripts *
+recursive-include experimental *
 
 prune build
 prune doc/_build


pllim:
I am actually now sure how to include this in `setup.py`. Package data appears to be relative to `ginga` sub-dir but `examples` are not *inside* it. And it also appears that `conda` ignores MANIFEST (conda/conda-build#383).

ejeschke:
> I am actually now sure how to include this in setup.py. Package data appears to be relative to ginga sub-dir but experimental is not inside it. 

So where does it get installed then via `pip`, if not inside the package?

pllim:
@ejeschke , I don't think it gets installed at all. I just did a `pip install ginga` and I cannot find it under `site-packages` area where `ginga` was installed. So, I think it is bundled in tarball but not installed.

ejeschke:
I see.  Well, I don't know how many people will dig around in the tarball to find it vs. just visiting github, but no matter.  Merging.

pllim:
If you want it installed, we will have to move `experimental` inside `ginga` folder.

ejeschke:
For now I don't think that is necessary.

PR #649, opened 2018-05-02
Enhance loader functionality

ejeschke:
This adds a "first-level" loader module to ginga.  This module allows opening a data item without knowing what it is first.  It works based off of the mime type of the file and also the magic function, if available.

This loader works with both the reference viewer as well as when using just the ginga widget.  This summarizes the difference:

Before:
```python
  image = AstroImage(logger)
  image.load_file(some_path)
```
After:
```python
  from ginga.util.loader import load_data
  image = load_data(path, logger=logger)
```
Although it is shorter, the main reason it is better is that in the first instance we commit to creating a `AstroImage`, but if the file references an RGB file this is not necessarily the right type.  In the second case an `AstroImage`, a `RGBImage` or some other kind of image might be returned.

### DIFF
--- a/ginga/RGBImage.py
+++ b/ginga/RGBImage.py
@@ -9,7 +9,7 @@ import numpy as np
 from ginga import trcalc
 from ginga.util import io_rgb
 from ginga.misc import Bunch
-from ginga.BaseImage import BaseImage, Header, ImageError
+from ginga.BaseImage import BaseImage, ImageError
 
 
 class RGBImage(BaseImage):
--- a/ginga/RGBImage.py
+++ b/ginga/RGBImage.py
@@ -4,12 +4,12 @@
 # This is open-source software licensed under a BSD license.
 # Please see the file LICENSE.txt for details.
 #
+import numpy as np
+
 from ginga import trcalc
 from ginga.util import io_rgb
 from ginga.misc import Bunch
-from ginga.BaseImage import BaseImage, Header
-
-import numpy as np
+from ginga.BaseImage import BaseImage, Header, ImageError
 
 
 class RGBImage(BaseImage):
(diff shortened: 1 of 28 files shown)


ejeschke:
Rebased.

ejeschke:
> In addition, is it worth adding some tests to take advantage of the CI?

Completely agree.  Since this is the first loading test (from a file) to be added, I'm thinking we might as well add a few smallish FITS files to the package as test data.  Shall we add these as a subdirectory of `tests`, or in `examples/data` or somewhere else?  I'm thinking those probably should also get installed with the viewer.

ejeschke:
Rebased to remove conflicts.

pllim:
For small data files, I think it is okay to include with package install. We can then also use them for other tests or demos, as needed.

I don't understand the errors on CI. Looks like a problem with `conda` or Travis.

ejeschke:
> I don't understand the errors on CI. Looks like a problem with conda or Travis.

S'ok, I'm rethinking/reworking some of this and will be pushing more commits.

ejeschke:
Rebased.

pllim:
p.s. 2.7.2 is out, so I cleared the milestone -- FYI

ejeschke:
Rebased.

ejeschke:
@pllim, I think the ASDF load compatibility fits in with this PR, for the reference viewer anyway.  We would be looking to modify `ginga.util.loader` to identify and load ASDF files as necessary.

pllim:
@ejeschke , should we merge this and add `ginga.util.loader` as a follow up over at #719 ?

ejeschke:
Rebased.

ejeschke:
> should we merge this and add ginga.util.loader for ASDF as a follow up over at #719 ?

@pllim, after rebasing this still looks ok.  Shall we merge and then look at #719?

PR #652, opened 2018-05-18
DOC: How to autostart plugins

pllim:
A user reported confusing documentation at http://ginga.readthedocs.io/en/latest/manual/customizing.html#adding-or-removing-plugins . Turns out what was actually needed was not clearly documented, so this PR adds the missing info to documentation.

c/c @rjavila

### DIFF
--- a/doc/manual/customizing.rst
+++ b/doc/manual/customizing.rst
@@ -253,6 +253,12 @@ A plugin can be started automatically in ``post_gui_config()`` using the
         ginga_shell.add_channel('Image')
         ginga_shell.start_local_plugin('Image', 'Histogram', None)
 
+Alternately, you can also start plugins via the command line interface using
+``--plugins`` and ``-modules`` for local and global plugins, respectively.
+To load multiple plugins at once, use a comma-separated list. For example::
+
+    ginga --plugins=MyLocalPlugin,Imexam --modules=MyGlobalPlugin
+
 ==============
 Adding Plugins
 ==============
@@ -292,6 +298,20 @@ Other keywords that can be used in a spec:
 
 * To prevent a control icon from appearing in the Operations plugin
   manager tray specify `optray=False`.
+
+=================
+Disabling Plugins
+=================
+
+Both local and global plugins can be disabled (thus, not shown in the
+reference viewer) using the ``--disable-plugins`` option in the
+command line interface. To remove multiple plugins at once,
+use a comma-separated list. For example::
+
+    ginga --disable-plugins=Zoom,Compose
+
+Some essential plugins, like ``Info``, cannot be disabled (i.e., attempt to
+disable them would have no effect).
 
 ==============================
 Making a Custom Startup Script
--- a/doc/manual/customizing.rst
+++ b/doc/manual/customizing.rst
@@ -236,9 +236,26 @@ workspace for creating channels, and the "dialogs" workspace is where
 most local plugins are instantiated (when activated), by default.
 These two names should at least be defined somewhere in default_layout.
 
-==========================
-Adding or Removing Plugins
-==========================
+==================
+Auto-Start Plugins
+==================
+
+Not all plugins provided by Ginga are automatically started up by default.
+A plugin can be started automatically in ``post_gui_config()`` using the
+``start_global_plugin()`` or ``start_local_plugin()`` methods, as appropriate::
+
+    def post_gui_config(ginga_shell):
+        # Auto start global plugins
+        ginga_shell.start_global_plugin('Zoom')
+        ginga_shell.start_global_plugin('Header')
+
+        # Auto start local plugin
+        ginga_shell.add_channel('Image')
+        ginga_shell.start_local_plugin('Image', 'Histogram', None)
+
+==============
+Adding Plugins
+==============
 
 A plugin can be added to the reference viewer in ``pre_gui_config()``
 using the ``add_plugin()`` method with a specification ("spec") for
--- a/doc/manual/customizing.rst
+++ b/doc/manual/customizing.rst
@@ -315,8 +315,8 @@ file. For example::
 
     disable_plugins = "Zoom,Compose"
 
-Some essential plugins, like ``Info``, cannot be disabled (i.e., attempt to
-disable them would have no effect).
+Some plugins, like ``Operations``, when disabled, may result in
+inconvenient GUI experience.
 
 ==============================
 Making a Custom Startup Script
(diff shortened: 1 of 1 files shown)


doc/manual/customizing.rst
```diff
@@ -236,9 +236,26 @@ workspace for creating channels, and the "dialogs" workspace is where
 most local plugins are instantiated (when activated), by default.
 These two names should at least be defined somewhere in default_layout.
 
-==========================
-Adding or Removing Plugins
```
pllim: @ejeschke , I find this title misleading, as the section does not mention anything on how to remove a plugin (is removing a plugin even possible?), so I modified it.

rjavila:
Thanks @pllim , this is extremely helpful.

doc/manual/customizing.rst
ejeschke: @pllim, I think the only way to "remove" a plugin is to disable it with `--disable-plugins` command line option or setting of the `disable_plugins` setting in `general.cfg`.  Probably something which should be rectified. 

We could add a little text about these options.  I think global plugins will be auto started if they appear in the `--modules` command line option, but not local plugins.

ejeschke: Otherwise this looks like a helpful addition to the docs.

pllim: @ejeschke , setting `disable_plugins` in `general.cfg` does not work but using `--disable-plugins` in command line does. FYI.

pllim:
@ejeschke , I added more doc, as requested. Please review.

ejeschke:
Hmm, `disable_plugins = "Thumbs"` in `general.cfg` doesn't work for you?  It does for me with current master.

pllim:
Ah, seems to work in `master` but not in 2.7.0 (latest release). Is it wise to document features not yet in a release? I guess `latest` RTD reflects what is in dev, however http://ginga.readthedocs.io/ (default link) also points to `latest`, which some could mistaken for "stable". What do you think, @ejeschke ?

ejeschke:
That's a good question.  I've always assumed that the latest online doc matches the latest online (dev) version, since the doc build is triggered from the github commits.  I think release versions of the documentation can be called up on `readthedocs` using the appropriate link, right?

pllim:
@ejeschke , another way to ask this question is, should http://ginga.readthedocs.io/ route to "latest" or "stable" by default? Perhaps this can be another issue. Should I open a new issue?

doc/manual/customizing.rst
```diff
@@ -276,6 +299,20 @@ Other keywords that can be used in a spec:
...
+
+Both local and global plugins can be disabled (thus, not shown in the
+reference viewer) using the ``--disable-plugins`` option in the
+command line interface. To remove multiple plugins at once,
+use a comma-separated list. For example::
+
+    ginga --disable-plugins=Zoom,Compose
+
+Some essential plugins, like ``Info``, cannot be disabled (i.e., attempt to
+disable them would have no effect).
```
ejeschke: Actually, there is no problem to disable `Info`, `Pan`, `Cursor`, `Colorbar`, `Toolbar` or basically any plugin.  You can even disable `Operations` and things still work (although maybe not as conveniently).

pllim: @ejeschke are you sure? I put `disable_plugins = "Info"` in my `general.cfg` but I still see "Info" loaded.

doc/manual/customizing.rst
```diff
@@ -276,6 +299,25 @@ Other keywords that can be used in a spec:
...
+
+    ginga --disable-plugins=Zoom,Compose
+
+Alternately, plugins can also be disabled via ``general.cfg`` configuration
+file. For example::
+
+    disable_plugins = "Zoom,Compose"
+
+Some essential plugins, like ``Info``, cannot be disabled (i.e., attempt to
+disable them would have no effect).
```
ejeschke: Actually, you can disable `Pan`, `Info`, `Colorbar`, `Toolbar`, `Cursor` and basically any others and Ginga will start up fine (there should be no error messages relating to the missing plugins).  Even `Operations` can be disabled and you can still use Ginga (although some things may not be as convenient).

pllim: @ejeschke , doesn't work like that for me even using dev. Please see https://github.com/ejeschke/ginga/pull/652#discussion_r194886211

ejeschke:
> @ejeschke , another way to ask this question is, should http://ginga.readthedocs.io/ route to "latest" or "stable" by default? Perhaps this can be another issue. Should I open a new issue?

You could if you think the current scheme should be changed.

doc/manual/customizing.rst
ejeschke: I just checked using a fresh instance in a docker container (so downloaded and installed in a clean environment).  I can disable plugins with no problem, and `disable_plugins` works in `general.cfg`.

Can you double check that you don't have an old installation somewhere in your path?  or maybe clear out all the compiled files and try again?

pllim: I'll follow up tomorrow. Thanks.

ejeschke: Sorry, I might be confused by different topics in this thread.  My remarks apply to the "dev" version only.  It's very possible that you can't disable any old plugin in the last "release" version, and the `disable_plugins` feature in `general.cfg` is only available in dev.  But all this will work in 2.7.1.

pllim: @ejeschke , I was using the latest dev of Ginga but I know why we are both confused now. This is because when "Info" is disabled, I still see the "Info" tab, but I did not realize that "Synopsis" sub-tab was indeed disabled. So, in a way, "Info" is still shown but not functional. :woman_shrugging:

pllim: (Anyway, I'll remove that blurb from the doc.)

pllim:
@ejeschke , I think I have addressed all your comments. Do you think this is okay to merge now?

doc/manual/customizing.rst
ejeschke: Whew!   Yes, that Info tab is part of the workspace, actually, so unless the workspace is altered it will still be created, even though there is no plugin present to populate it.

doc/manual/customizing.rst
ejeschke: > setting disable_plugins in general.cfg does not work but using --disable-plugins in command line does

@pllim, did you confirm that using `disable_plugins` in `general.cfg` is working in dev?

pllim: @ejeschke , it is confirmed working. Thanks!

ejeschke:
@pllim, feel free to merge anytime.

PR #660, opened 2018-06-26
Minor bug fixes so Info and Crosshair would work

pllim:
`Info` and `Crosshair` were crashing on me but they seem to work again with this patch.


ejeschke:
Now we are just getting the expected `numpy` error.

ejeschke:
Hold the merge, think I found another issue with `Crosshair`.

ejeschke:
> Hold the merge, think I found another issue with Crosshair.

No, after double check I see the PR is working fine.

PR #665, opened 2018-06-29
Fix writing out to PNG with PIL

pllim:
I got an error writing using `viewer.save_rgb_image_as_file("screenshot.png")` for `EnhancedCanvasView` under `jupyterw`. This patch fixed the error for me. I am not sure if this is specific to PIL or  not.

xref eteq/astrowidgets#1

### DIFF
--- a/ginga/ImageView.py
+++ b/ginga/ImageView.py
@@ -3407,7 +3407,7 @@ class ImageViewBase(Callback.Callbacks):
             See :meth:`get_rgb_image_as_buffer`.
 
         """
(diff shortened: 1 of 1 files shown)


pllim:
Test failure is unrelated. FYI.

pllim:
What about Python 2? Does it matter?

ejeschke:
> What about Python 2? Does it matter?

Yes, currently it matters, but as PNG is a binary format, this should be fine.

pllim:
Note to self: This was included in 2.7.1

Issue #668, opened 2018-07-14
DOC: Plugins doc should reflect new menu layout


pllim:
#619 changed "operations" menu layout but plugin instructions are not updated accordingly, as reported by @mkb0517 in https://github.com/ejeschke/ginga/issues/603#issuecomment-405052500 . A sweep through all the plugins to correct this is necessary.

pllim:
Following discussions from #671

@ejeschke -- "I wonder about the whole idea of 'hardcoding' the menu locations of plugins in the documentation, given that the user can actually create their own menu order quite easily! Maybe we need some language that allows for this flexibility."

@mkb0517 -- "Maybe if we add “by default” to it? I’d assume that by the time they were making they were making their own menus, they’d be familiar enough with the layout that they would disregard that part."

PR #671, opened 2018-07-15
Changed the Compose documentation to reflect the menu nesting. Clarif…

mkb0517:
…ied that the tab that appears under Dialogs is a result of selecting compose

### DIFF
--- a/ginga/rv/plugins/Compose.py
+++ b/ginga/rv/plugins/Compose.py
@@ -10,8 +10,9 @@ channel.  An instance can be opened for each channel.
 
 **Usage**
 
-Start the ``Compose`` plugin from the "Operations" menu -- the tab should
-show up under "Dialogs".
+Start the ``Compose`` plugin from the "Operation->RGB" (below) or
+"Plugins->RGB" (above) menu. The tab should show up under the
+"Dialogs" tab in the viewer to the right as "IMAGE:Compose".
 
 1. Select the kind of composition you want to make from the "Compose Type"
    drop down: "RGB" for composing three monochrome images into a color


ginga/rv/plugins/Compose.py
```diff
@@ -10,8 +10,9 @@
 
 **Usage**
 
-Start the ``Compose`` plugin from the "Operations" menu -- the tab should
-show up under "Dialogs".
+Start the ``Compose`` plugin from the "Operation->RGB" (below) or
+"Plugins->RGB" (above) menu. The tab should show up under the
+"Dialogs" tab in the viewer to the right as "IMAGE:Compose".
```
pllim: To clarify a little further, the "IMAGE" in "IMAGE:Compose" is actually the channel name that the "Compose" local plugin is tied to. For example, if you start "Compose" from "DETECTOR1" channel name instead (there is a way to create new channels where you can provide useful names), it would appear as "DETECTOR1:Compose". Ways to potentially clarify this include but not limited to:

1. `"<channel name>:Compose"`
2. `"IMAGE:Compose", where "IMAGE" is the associated channel name`

I'll leave it to your good judgement. Thanks again for helping!

ejeschke:
Merging, but I wonder about the whole idea of "hardcoding" the menu locations of plugins in the documentation, given that the user can actually create their own menu order quite easily!  Maybe we need some language that allows for this flexibility.

ejeschke:
Thanks for contributing, @mkb0517 !

mkb0517:
Hmm I didn’t know it could do that. Maybe if we add “by default” to it? I’d assume that by the time they were making they were making their own menus, they’d be familiar enough with the layout that they would disregard that part.

pllim:
Thanks. I think these comments are valid and should be addressed package-wide as part of #668.

PR #676, opened 2018-07-27
Pluggable renderers

ejeschke:
This is a large PR that separates the formerly tight binding of renders and back ends (aka "toolkits").  Basically, with these changes you can now to some degree mix and match renderers with back ends.  There are some caveats:
1. the performance of a back end will generally be better if matched with the renderer that comes with that back end, and that is the default configuration (in other words, the Qt renderer with the Qt backend will generally work best).  However, this may change for certain situations (e.g. some renderers may acquire vector drawing capability or graphics accelerated, etc.)
2. some back ends are specialized and can only be used with their matched renderer.  

The PR basically preserves backward compatibility as far as API (only one internally used method has changed) so users shouldn't notice any difference.


ejeschke:
> How do I best test this?

I pushed an additional commit. If you pull the changes you can test with the `examples/gw/example2.py` program.  You give it a toolkit using the `-t` option (e.g. `-t qt5`) then if you specify the `-r` option you can set a renderer other than the default Qt renderer.  With the Qt backend you can use `-r qt` (the default), `-r pil`, `-r agg`, `-r opencv`, or `-r cairo`.  You need to have the appropriate package installed, of course.

Probably you already have `pillow`, `agg`, `opencv` installed. If you want to try cairo, it's available in conda as `pycairo`.

pllim:
I downloaded the `gw/example2.py` from this PR and I get `qtpy.PythonQtError: No Qt bindings could be found` because I don't have PySide installed. But I have `pyqt` installed and Ginga was working fine without PySide. I am hesistant to `conda install PySide` because it then attempts to downgrade a bunch of unrelated packages.

Here is the error when I run `python example2.py -t qt -r pil`:

```
  File ".../qtpy/__init__.py", line 148, in <module>
    from PySide import __version__ as PYSIDE_VERSION  # analysis:ignore
ModuleNotFoundError: No module named 'PySide'

During handling of the above exception, another exception occurred:

  File "example2.py", line 362, in <module>
    main(options, args)
  File "example2.py", line 294, in main
    viewer = FitsViewer(logger)
  File "example2.py", line 24, in __init__
    self.dc = get_canvas_types()
  File ".../ginga/canvas/CanvasObject.py", line 582, in get_canvas_types
    import ginga.canvas.types.all  # noqa
  File ".../ginga/canvas/types/all.py", line 5, in <module>
    from .basic import *  # noqa
  File ".../ginga/canvas/types/basic.py", line 15, in <module>
    from ginga.misc.ParamSet import Param
  File ".../ginga/misc/ParamSet.py", line 8, in <module>
    from ginga.gw import Widgets
  File ".../ginga/gw/Widgets.py", line 8, in <module>
    from ginga.qtw.Widgets import *  # noqa
  File ".../ginga/qtw/Widgets.py", line 10, in <module>
    from ginga.qtw.QtHelp import (QtGui, QtCore, QTextCursor, QIcon, QPixmap,
  File ".../ginga/qtw/QtHelp.py", line 35, in <module>
    from qtpy import QtCore
  File "/redkeep/ironthrone/anaconda/envs/py36/lib/python3.6/site-packages/qtpy/__init__.py", line 154, in <module>
    raise PythonQtError('No Qt bindings could be found')
qtpy.PythonQtError: No Qt bindings could be found
```

My `conda list qt`:

```
# Name                    Version                   Build  Channel
pyqt                      5.6.0            py36h8210e8a_6    conda-forge
pytest-qt                 2.4.1                      py_0    conda-forge
qt                        5.6.2               h974d657_12  
qtconsole                 4.3.1            py36h8f73b5b_0  
qtpy                      1.3.1            py36h3691cc8_0  
```

pllim:
Ah, thanks! I get these two errors below but otherwise it works:

```
Error in scaling: window size undefined
Error in scaling: window size undefined
```

I tried both

`python example2.py -t qt5 -r pil`

and

`python example2.py -t qt5 -r opencv`

but can't tell any difference other than they both worked. I guess that is good news.

ejeschke:
> but can't tell any difference other than they all worked. I guess that is good news.

It's especially good news if you can't tell the difference, because for some of them at least there is a performance penalty. But I think if you simple scroll the image you may be able to tell that the text rendering is slightly different for some renderers.

ejeschke:
Rebased.

PR #680, opened 2018-08-06
Fix issue #677, for FutureWarning with numpy>=1.15.0

### LINKED ISSUE #677: FutureWarning: Using a non-tuple sequence for multidimensional indexing is deprecated
I get the following warning when using Ginga:
```
...\ginga\AstroImage.py:183: FutureWarning: Using a non-tuple sequence
for multidimensional indexing is deprecated; use `arr[tuple(seq)]` instead
of `arr[seq]`. In the future this will be interpreted as an array index,
`arr[np.array(seq)]`, which will result either in an error or a different result.
  data = self.get_mddata()[view]
```

I think it is from numpy/numpy#9686 ?

ejeschke:
This fixes an issue for the error message issued by numpy >= 1.15.0:

FutureWarning: Using a non-tuple sequence for multidimensional indexing is
deprecated; use `arr[tuple(seq)]` instead of `arr[seq]`. In the future this
will be interpreted as an array index, `arr[np.array(seq)]`, which will
result either in an error or a different result.

EDIT: Fix #677

### DIFF
--- a/ginga/AstroImage.py
+++ b/ginga/AstroImage.py
@@ -179,7 +179,7 @@ class AstroImage(BaseImage):
         revnaxis.reverse()
 
         # construct slice view and extract it
-        view = revnaxis + [slice(None), slice(None)]
+        view = tuple(revnaxis + [slice(None), slice(None)])
         data = self.get_mddata()[view]
 
         if len(data.shape) != 2:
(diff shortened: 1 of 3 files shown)


ejeschke:
Tnx, @pllim. Merging.

PR #682, opened 2018-08-10
Cursor info updates when image changes in viewer

pllim:
@dthatte reported that cursor info (coordinates and value) does not update when images are blinked in a channel (e.g., pressing up/down arrow keys to switch images in a channel). This PR is an attempt to fix that.


pllim:
OK thanks for the correct fix. I'll close this one.

Superseded by #684

PR #686, opened 2018-08-22
Enhancements for the PixTable plugin

ejeschke:
This adds the following enhancements:
- add RMS to the summary statistics
- add standard move/draw/edit buttons for editing marks
- add a caption box to leave comments with marks
- add keystroke shortcuts to add, move or delete a mark

### DIFF
--- a/ginga/examples/configs/plugin_PixTable.cfg
+++ b/ginga/examples/configs/plugin_PixTable.cfg
@@ -8,3 +8,18 @@ font = 'fixed'
 
 # Default font size
 fontsize = 12
+
+# default size for mark point radius
+mark_radius = 10
+
+# style of point to draw
+mark_style = 'cross'
+
+# color of non-selected marks
+mark_color = 'purple'
+
+# color of selected mark
+select_color = 'cyan'
+
+# whether to update the pixel table when moving a mark around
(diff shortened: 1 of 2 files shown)


ejeschke:
> Now, it starts at mark4 instead of back at mark1. Not sure if this is a bug or feature, but does not stop the plugin from working.

That's how it has always worked. I thought about "recovering" the deleted numbers (which I think is what "Cuts" does, actually), but I thought with marks it might be more confusing to have a newer mark have a lower number than an older one.  Something we could revise later if we get feedback that users would prefer a non-monotonically increasing scheme.

pllim:
> non-monotonically increasing scheme

I agree that would be confusing. I was thinking maybe the numbering can reset when and only when *all* markers are deleted. However, I agree that is out of scope of this PR.

ejeschke:
Rebased.

ejeschke:
@pllim, going to merge, esp as this PR fixes an issue that prevented marks from working with RGB images.  We can take up an extra improvements in a separate PR.

ejeschke:
BTW, I did make the numbering restart if the user does a "Delete All". I think that makes sense.

pllim:
Thanks!

PR #689, opened 2018-08-27
Fix for github issue #257

### LINKED ISSUE #257: Sorting by numbers with TreeView

ejeschke:
- possible fix for an issue sorting the TreeWidget in Qt backend

EDIT: Fix #257 

EDIT: Fix spacetelescope/stginga#83

### DIFF
--- a/ginga/qtw/Widgets.py
+++ b/ginga/qtw/Widgets.py
@@ -652,9 +652,11 @@ class TreeWidgetItem(QtGui.QTreeWidgetItem):
     def __lt__(self, otherItem):
         column = self.treeWidget().sortColumn()
         try:
-            return float( self.text(column) ) < float( otherItem.text(column) )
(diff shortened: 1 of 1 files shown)


ejeschke:
@pllim, had an urgent need to fix #257.  Can you check your plugins that use TreeView and see if it is doing the right things?  We can maybe find a less hacky fix later but this seems to do the job.

ejeschke:
Looks like gotta fix a couple pep8 issues...

pllim:
Before 

screenshot from 2018-08-27 12-22-53

After

screenshot from 2018-08-27 12-18-58

:tada:

ejeschke:
:+1:  Merging.

PR #694, opened 2018-08-28
Enable calling plugin methods from remote interface

ejeschke:
This adds two methods to the ginga shell that allow methods in a global or local plugin to be called.  This can be used e.g. from the RC interface, or internally.

### DIFF
--- a/ginga/rv/Control.py
+++ b/ginga/rv/Control.py
@@ -283,6 +283,28 @@ class GingaShell(GwMain.GwMain, Widgets.Application):
 
     def call_local_plugin_method(self, chname, plugin_name, method_name,
                                  args, kwargs):
+        """
+        Parameters
+        ----------
+        chname : str
+            The name of the channel containing the plugin.
+
+        plugin_name : str
+            The name of the local plugin containing the method to call.
+
+        method_name : str
+            The name of the method to call.
+
+        args : list or tuple
+            The positional arguments to the method
+
+        kwargs : dict
+            The keyword arguments to the method
+
+        Returns
+        -------
+        result : return value from calling the method
+        """
         channel = self.get_channel(chname)
         opmon = channel.opmon
         p_obj = opmon.get_plugin(plugin_name)
@@ -300,6 +322,25 @@ class GingaShell(GwMain.GwMain, Widgets.Application):
 
     def call_global_plugin_method(self, plugin_name, method_name,
                                   args, kwargs):
+        """
+        Parameters
+        ----------
+        plugin_name : str
+            The name of the global plugin containing the method to call.
+
+        method_name : str
+            The name of the method to call.
+
+        args : list or tuple
+            The positional arguments to the method
+
+        kwargs : dict
+            The keyword arguments to the method
+
+        Returns
+        -------
+        result : return value from calling the method
+        """
         p_obj = self.gpmon.get_plugin(plugin_name)
         method = getattr(p_obj, method_name)
         return self.gui_call(method, *args, **kwargs)
--- a/ginga/rv/Control.py
+++ b/ginga/rv/Control.py
@@ -281,6 +281,13 @@ class GingaShell(GwMain.GwMain, Widgets.Application):
         opmon = channel.opmon
         opmon.deactivate(opname)
 
+    def call_local_plugin_method(self, chname, plugin_name, args, kwargs):
+        channel = self.get_channel(chname)
+        opmon = channel.opmon
+        p_obj = opmon.get_plugin(plugin_name)
+        method = getattr(p_obj, plugin_name)
+        return self.gui_call(method, *args, **kwargs)
+
     def start_global_plugin(self, plugin_name, raise_tab=False):
         self.gpmon.start_plugin_future(None, plugin_name, None)
         if raise_tab:
@@ -289,6 +296,11 @@ class GingaShell(GwMain.GwMain, Widgets.Application):
 
     def stop_global_plugin(self, plugin_name):
         self.gpmon.deactivate(plugin_name)
+
+    def call_global_plugin_method(self, plugin_name, args, kwargs):
+        p_obj = self.gpmon.get_plugin(plugin_name)
+        method = getattr(p_obj, plugin_name)
+        return self.gui_call(method, *args, **kwargs)
 
     def start_plugin(self, plugin_name, spec):
         ptype = spec.get('ptype', 'local')
--- a/ginga/rv/Control.py
+++ b/ginga/rv/Control.py
@@ -281,11 +281,12 @@ class GingaShell(GwMain.GwMain, Widgets.Application):
         opmon = channel.opmon
         opmon.deactivate(opname)
 
-    def call_local_plugin_method(self, chname, plugin_name, args, kwargs):
+    def call_local_plugin_method(self, chname, plugin_name, method_name,
+                                 args, kwargs):
         channel = self.get_channel(chname)
         opmon = channel.opmon
         p_obj = opmon.get_plugin(plugin_name)
-        method = getattr(p_obj, plugin_name)
+        method = getattr(p_obj, method_name)
         return self.gui_call(method, *args, **kwargs)
 
     def start_global_plugin(self, plugin_name, raise_tab=False):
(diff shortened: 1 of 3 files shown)


ejeschke:
Ping @jhennawi, @pllim.

This small patch allows you to call a plugin method in either a global or local plugin from the `RC` interface.  To use it, use the "ginga" handle to call `call_local_plugin_method(chname, plugin_name, method_name, args, kwargs)` or `call_global_plugin_method(plugin_name, method_name, args, kwargs)`.  The plugin must already be loaded and started, so use existing machinery for that.

Caveat: *I have not tested this*!   It is a WIP, but it's such a simple change, and potentially useful for the `RC` interface it's worth trying.

ejeschke:
> But since this is not merged into the repository, I presume I have to fork in order to use it?

No, but if you want to test it, just clone that branch (or download the zip file) and install.

I think it is no problem to merge it as soon as we give it a quick spin, as it only adds new functionality and does not involve changing any existing calls. 

Regarding doing your own plugins, it is highly encouraged, as that is the way you can really leverage the power of the package.  @pllim and I are doing a poster on plugins for ADASS, and planning to update the documentation as a part of that.

It is not necessary to fork Ginga to write/use your own plugins.  There are several ways to load your plugins into a stock Ginga install when it starts up:
* command line option to load additional plugins
* configuration file option to load additional plugins
* use the [ginga-plugin-template](https://github.com/ejeschke/ginga-plugin-template) to write plugins that can be installed independently, but are loaded when ginga starts up

ejeschke:
> Suggestions on how to proceed?

[current documentation on writing a plugin](https://ginga.readthedocs.io/en/stable/manual/developers.html)

After perusing that, take a look at the [collection of distributed plugins](https://github.com/ejeschke/ginga/tree/master/ginga/rv/plugins), especially examining ones that do something similar to what you want to do.

We'll be happy to field any questions!

pllim:
Yes, you can distribute your own plugins separately. See https://github.com/spacetelescope/stginga for example.

ejeschke:
> any chance you could add docstrings to explain these new methods?

Sure, I'll update the PR.

ginga/rv/Control.py
```diff
@@ -281,6 +281,36 @@ def stop_local_plugin(self, chname, opname):
...
+
+        args : list or tuple
+            The positional arguments to the method
+
+        kwargs : dict
+            The keyword arguments to the method
+
+        Returns
+        -------
+        result : return value from calling the method
```
pllim: Is this a `bool`, `str`, `obj`, ...?

pllim:
Actually... I am not sure if these methods work properly or I am just using them wrong... I was investigating https://github.com/ejeschke/ginga/issues/691#issuecomment-417108222 and decided to use this branch, but I encounter errors.

Server side:
```

```

Client side:
```python
>>> import numpy as np
>>> from ginga.util import grc
>>> viewer = grc.RemoteClient('localhost', 9000)
>>> ch = viewer.channel('Image')
>>> img = np.random.rand(2040, 2040) * 10000.0
>>> ch.load_np('Image_Name', img, 'fits', {})
>>> sh = viewer.shell()
>>> sh.start_operation('TVMask')
```

And then when I call `sh.call_local_plugin_method('Image', 'TVMask', 'load_file', 'MOS_8458_G395H_testmask1.fits')`, I see this error:
```
127.0.0.1 - - [30/Aug/2018 12:26:25] "POST /RPC2 HTTP/1.1" 200 -
2018-08-30 12:27:17,342 | E | GwMain.py:76 (_execute_future) | gui event loop error: call_local_plugin_method() missing 1 required positional argument: 'kwargs'
2018-08-30 12:27:17,342 | E | GwMain.py:80 (_execute_future) | Traceback:
  File .../ginga/gw/GwMain.py", line 73, in _execute_future
    future.thaw(suppress_exception=False)
  File .../ginga/misc/Future.py", line 46, in thaw
```

That method only take `filename` as positional arg and accepts no keyword, but okay, I'll try `sh.call_local_plugin_method('Image', 'TVMask', 'load_file', 'MOS_8458_G395H_testmask1.fits', {})` instead but get a different error:
```
127.0.0.1 - - [30/Aug/2018 12:23:52] "POST /RPC2 HTTP/1.1" 200 -
2018-08-30 12:26:25,354 | E | GwMain.py:76 (_execute_future) | gui event loop error: load_file() takes 2 positional arguments but 30 were given
2018-08-30 12:26:25,354 | E | GwMain.py:80 (_execute_future) | Traceback:
  File ".../ginga/gw/GwMain.py", line 73, in _execute_future
    future.thaw(suppress_exception=False)
  File ".../ginga/misc/Future.py", line 46, in thaw
  File ".../ginga/rv/Control.py", line 312, in call_local_plugin_method
    return self.gui_call(method, *args, **kwargs)
  File ".../ginga/gw/GwMain.py", line 184, in gui_call
    return method(*args, **kwdargs)
```

https://github.com/ejeschke/ginga/blob/fa239ab8d43a9e5070663ed2b5f535e676f3c5c4/ginga/rv/plugins/TVMask.py#L259-L265

ginga/rv/Control.py
ejeschke: It depends what the method you are calling returns. What is the docstring convention for specifying any valid python type?

ejeschke:
> sh.call_local_plugin_method('Image', 'TVMask', 'load_file', 'MOS_8458_G395H_testmask1.fits')

@pllim, please call it like this:
```python
sh.call_local_plugin_method('Image', 'TVMask', 'load_file', ['MOS_8458_G395H_testmask1.fits'], {})
```
[`call_local_plugin_method()`](https://github.com/ejeschke/ginga/pull/694/files#diff-91b277d99dafe70464e0f396983eac0fR284) takes a `list` or `tuple` for the positional arguments and a `dict` for the keyword arguments.

ginga/rv/Control.py
pllim: I am not sure if there is one for *any*

pllim:
Ah, thanks, @ejeschke ! It works now.

```python
>>> sh.call_local_plugin_method(
...     'Image', 'TVMask', 'load_file', ('MOS_8458_G395H_testmask1.fits', ), {})
'#UNDEFINED'
```

@jhennawi , are you able to display the mask following this command?

ginga/rv/Control.py
ejeschke: Ok, so maybe the current docstring is ok? ... :grin:

jhennawi:
Hi @ejeschke and @pllim,

Thanks!  Here is what I executed: 

viewer = grc.RemoteClient(host, port)
shell = viewer.shell()
shell.start_operation('TVMask')
shell.call_local_plugin_method(channel_name, 'TVMask', 'load_file', [maskfile], {})

This worked but a few thoughts: 

1) My understanding is that TVMask is a local plugin, i.e. channel specific. As such it seems
like the call to launch a TVMask for a given channel should take the channel name or something, whereas start_operation seems ambiguous. I think it just fires up the TVMask plugin in whatever channel you are currently on. 

To me a more intuitive calling sequence would be something like: 

viewer = grc.RemoteClient(host, port)
shell.start_operation(channel_name, 'TVMask')
shell.call_local_plugin_method(channel_name, 'TVMask', 'load_file', [maskfile], {})

Or even more intuitive to me would be if the global plugins were launched form the shell, and the local plugins launched from the  ch = viewer.channel(channel_name) object. Not sure if that is possible though. 

2) Right now as it stands, I have to write out a file called mask file for every mask which is a bit tedious. It would be great if there was a way to pass in the mask an array, but I suspect that would require an all new plugin. 

3) My wish list would be something like @pllim  DQInspect plugin, which could be passed a bitmask image and a dictionary of bit codes/values (like in DQInspect) and overlay each bit with colors. And I'd prefer to be able to launch this from the RC interface. 

Joe

ejeschke:
> shell.start_operation(channel_name, 'TVMask')

You can use `shell.start_local_plugin(channel_name, 'TVMask', None)` using the current API.

> Or even more intuitive to me would be if the global plugins were launched form the shell, and the local plugins launched from the ch = viewer.channel(channel_name) object. Not sure if that is possible though.

Could be done with some code added to the `Channel` class.  Currently that class does not know much about plugins, and I'm not sure it should be complicated with API calls that deal with them.  But possibly it could be added to the `ginga.util.grc` module as a method in the [`_channel_proxy`](https://github.com/ejeschke/ginga/blob/master/ginga/util/grc.py#L54) class.

ejeschke:
> My wish list would be something like @pllim DQInspect plugin, which could be passed a bitmask image and a dictionary of bit codes/values (like in DQInspect) and overlay each bit with colors. And I'd prefer to be able to launch this from the RC interface.

Perhaps a PR that takes the code from `DQInspect` and modifies it to be a bit more generic?  You can design the API of the plugin to take in to account the possibility of using it from the `RC` interface.

But there is a certain amount of work to polish things up to be useful in the general way, so if you just need something to work for your particular pipeline, it might quicker and easier to just make a custom plugin.  What do you think, @pllim?

pllim:
Re: DQInspect -- I'll have to think about it but not against adding RC-only methods. Since that is unrelated to this PR, it should not hold up the merging of this one. Thanks!

Issue #673, opened 2018-07-24
MNT: Python 3.7 support


pllim:
1. Ensure Ginga works with Python 3.7
2. Add Python 3.7 to Travis CI matrix

ejeschke:
Just tested with a fresh `miniconda` install.  Tests all pass.  Reference viewer seems to run fine with the sampling of plugins I tried.

pllim:
Thanks for checking! When I get the chance, I'll try it more extensively and report back.

ejeschke:
I've been testing almost exclusively lately with 3.7.  I haven't found any issues.  I think this will be a straightforward PR.

pllim:
Fixed in #699

PR #699, opened 2018-09-08
Add python 3.7 to Travis tests, remove python 3.5

ejeschke:
Addresses #673.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,7 +1,7 @@
 language: python
 
 python:
-    - 3.6
+    - 3.7
 
 # Setting sudo to false opts in to Travis-CI container-based builds.
 sudo: false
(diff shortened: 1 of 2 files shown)


.travis.yml
```diff
@@ -1,7 +1,12 @@
-language: python
+# We set the language to c because python isn't supported on the MacOS X nodes
+# on Travis. However, the language ends up being irrelevant anyway, since we
+# install Python ourselves using conda.
+language: c
```
ejeschke: I was wondering about that, because it seemed like Travis was trying to install a system python, and I was pretty sure we were using miniconda's python.  I think this makes more sense to put `c` here (or maybe `bash`)?

ejeschke:
LGTM, @pllim!

.travis.yml
pllim: I would stick with C since it works for Astropy. :)

PR #704, opened 2018-11-07
Fix typo in setup.cfg 

pllim:
Probably an oversight when helpers changed the command from `build_sphinx` to `build_docs`.


pllim:
@ejeschke , this is uncontroversial, so I am going to go ahead and merge. Hope you don't mind.

PR #706, opened 2018-11-07
Avoid using masked array for iqcalc

pllim:
When using `Mosaic` plugin, I noticed this warning:
```
2018-11-07 16:04:20,707 | W | warnings.py:99 (_showwarnmsg) |
    .../numpy/core/fromnumeric.py:688: UserWarning: Warning:
    'partition' will ignore the 'mask' of the MaskedArray.
  a.partition(kth, axis=axis, kind=kind, order=order)
```

Then, I found numpy/numpy#7330,  numpy/numpy#7345, and numpy/numpy#8669. To illustrate the problem using Numpy 1.15.4:
```python
>>> import numpy as np
>>> data_np = np.array([1, np.nan, 3, 4, 5])
>>> mdata = np.ma.masked_array(data_np, np.isnan(data_np))
>>> np.median(mdata)
.../numpy/core/fromnumeric.py:688: UserWarning: Warning: 'partition' will ignore the 'mask' of the MaskedArray.
  a.partition(kth, axis=axis, kind=kind, order=order)
.../numpy/lib/function_base.py:3250: RuntimeWarning: Invalid value encountered in median
  r = func(a, **kwargs)
nan
>>> np.ma.median(mdata)
3.5
```

However, behavior of `np.ma.median` is unstable (probably fixed since Numpy 1.12 but you seem to support Numpy as old as 1.9). Even if you use the latest Numpy, when the whole array is NaN, `np.ma.median` (or `np.ma.mean`) returns "masked", which you cannot operate on anyway.

This PR avoids masked array altogether. The tests should show you the expected behaviors. If these are not the behaviors you want, please let me know.

### DIFF
--- a/ginga/util/iqcalc.py
+++ b/ginga/util/iqcalc.py
@@ -23,13 +23,32 @@ from ginga.misc import Bunch
 
 
 def get_mean(data_np):
-    mdata = np.ma.masked_array(data_np, np.isnan(data_np))
-    return np.mean(mdata)
+    """Calculate mean for valid values.
+
+    Parameters
+    ----------
+    data_np : ndarray
+        Input array.
+
+    Returns
+    -------
+    result : float
+        Mean of array values that are finite.
+        If array contains no finite values, returns NaN.
+
+    """
+    i = np.isfinite(data_np)
+    if not np.any(i):
+        return np.nan
+    return np.mean(data_np[i])
 
 
 def get_median(data_np):
-    mdata = np.ma.masked_array(data_np, np.isnan(data_np))
-    return np.median(mdata)
+    """Like :func:`get_mean` but for median."""
+    i = np.isfinite(data_np)
+    if not np.any(i):
+        return np.nan
+    return np.median(data_np[i])
 
 
 class IQCalcError(Exception):
--- a/ginga/tests/test_iqcalc.py
+++ b/ginga/tests/test_iqcalc.py
@@ -1,0 +1,20 @@
+import numpy as np
+import pytest
+
+from ginga.util import iqcalc
+
+
+@pytest.mark.parametrize(
+    ('arr', 'ans'),
+    [(np.arange(5), 2),
+     (np.array([1, np.inf, 3, np.nan, 5]), 3),
+     (np.arange(10).reshape(2, 5), 4.5)])
+def test_get_mean_median(arr, ans):
+    assert iqcalc.get_mean(arr) == ans
+    assert iqcalc.get_median(arr) == ans
+
+
+def test_get_mean_median_nan():
+    arr = np.array([np.nan, np.inf])
+    assert np.isnan(iqcalc.get_mean(arr))
+    assert np.isnan(iqcalc.get_median(arr))


ginga/util/iqcalc.py
```diff
@@ -23,13 +23,32 @@
 
 
 def get_mean(data_np):
-    mdata = np.ma.masked_array(data_np, np.isnan(data_np))
```
pllim: Note: Assumption is that `data_np` is *always* a Numpy array. If you want to pass in list, there needs to be extra processing.

pllim:
p.s. You might also want to revisit other usage of masked arrays in Ginga, like `cutout_shape`.

ginga/util/iqcalc.py
ejeschke: Hmm...it's on my TODO list to support `dask` arrays.

pllim:
Numpy is definitely not dropping support but those particular operations went through some stuff between 1.9 and 1.15. Backward compatibility is going to be a pain. Also masked array have extra overhead.

ginga/util/iqcalc.py
pllim: If `dask` array quacks like Numpy, this should be okay? I'm not familiar enough to say for sure.

ejeschke:
> Also masked array have extra overhead

Hmm, but so does creating an index array and then operating on it.  Be interesting to see a speed comparison.  The masked array approach allows it all to happen in C in one python call.

ginga/util/iqcalc.py
ejeschke: Apparently `dask` supports some `numpy` ufuncs, but not all.  It also only supports a subset of the `numpy` fancy indexing, some of which Ginga uses.

ejeschke:
```python
In [2]: import numpy as np

In [3]: def get_median1(data):
   ...:     mdata = np.ma.masked_array(data, np.isnan(data))
   ...:     return np.mean(mdata)
   ...: 
   ...: 
In [7]: arr = np.random.randint(0, 100000, (1024, 1024))

In [8]: %timeit get_median1(arr)
8.14 ms ± 1.64 ms per loop (mean ± std. dev. of 7 runs, 100 loops each)

In [9]: def get_median2(data):
   ...:     i = np.isfinite(data)
   ...:     if not np.any(i):
   ...:         return np.nan
   ...:     return np.median(data[i])
   ...: 
   ...: 

In [10]: %timeit get_median2(arr)
19 ms ± 6.31 ms per loop (mean ± std. dev. of 7 runs, 10 loops each)
```

ejeschke:
When I keep testing, the indexed version is basically 3X slower.

ejeschke:
They become closer as the array gets smaller, but still around 2X slower.

ejeschke:
Whups!!  comparing mean to median!  Let's try that again...

ejeschke:
```python
In [29]: def get_median1(data):                          
    ...:     mdata = np.ma.masked_array(data, np.isnan(data))
    ...:     return np.ma.median(mdata)
    ...:                      
    ...:                              
In [30]: %timeit get_median1(arr)
205 ms ± 55.4 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)

In [31]: %timeit get_median2(arr)
24.5 ms ± 6.19 ms per loop (mean ± std. dev. of 7 runs, 10 loops each)
```
Now it looks like the indexed version blows it away! (Of course I changed from `numpy.median` to `numpy.ma.median`.  So this is looking like a great PR. :smile:

pllim:
For completeness, more timing using your input and functions above and similar ones for mean:

```python
In [5]:  %timeit get_median1(arr)
112 ms ± 30.9 µs per loop (mean ± std. dev. of 7 runs, 10 loops each)

In [6]: %timeit get_median2(arr)
12 ms ± 23 µs per loop (mean ± std. dev. of 7 runs, 100 loops each)

In [13]: %timeit get_mean1(arr)
2.62 ms ± 2.56 µs per loop (mean ± std. dev. of 7 runs, 100 loops each)

In [14]: %timeit get_mean2(arr)
2.73 ms ± 6.47 µs per loop (mean ± std. dev. of 7 runs, 100 loops each)
```

What about when array is all NaN?

```python
In [18]: arr = np.empty((1024, 1024))

In [19]: arr[:] = np.nan

In [26]: %timeit get_median1(arr)
41.7 ms ± 7.82 ms per loop (mean ± std. dev. of 7 runs, 10 loops each)

In [27]: %timeit get_median2(arr)
498 µs ± 228 ns per loop (mean ± std. dev. of 7 runs, 1000 loops each)

In [28]: %timeit get_mean1(arr)
3.64 ms ± 169 µs per loop (mean ± std. dev. of 7 runs, 100 loops each)

In [29]: %timeit get_mean2(arr)
501 µs ± 3.11 µs per loop (mean ± std. dev. of 7 runs, 1000 loops each)
```

A case of 2D float array with one NaN:

```
In [43]: arr = np.arange(1024*1024, dtype=np.float).reshape((1024, 1024))

In [44]: arr[0, 0] = np.nan

In [45]: %timeit get_median1(arr)
34.4 ms ± 103 µs per loop (mean ± std. dev. of 7 runs, 10 loops each)

In [46]: %timeit get_median2(arr)
2.96 ms ± 3.77 µs per loop (mean ± std. dev. of 7 runs, 100 loops each)

In [48]: %timeit get_mean1(arr)
2.73 ms ± 1.28 µs per loop (mean ± std. dev. of 7 runs, 100 loops each)

In [49]: %timeit get_mean2(arr)
1.69 ms ± 1.83 µs per loop (mean ± std. dev. of 7 runs, 1000 loops each)
```

p.s. In my work in other plugins in the past, I did find that the performance increased noticeably when I switch away from using `np.ma` and instead extract what I needed from masked array returned by cutout method and used indexing instead.

ginga/util/iqcalc.py
pllim: I am definitely open to modifying these for `dask` integration in a future PR. 😄

PR #711, opened 2018-11-13
MNT: Remove PY2 support

pllim:
**THE POINT OF NO RETURN**

For #618


setup.py
```diff
@@ -109,8 +113,8 @@
 setup_requires = ['numpy>=1.9']
 
 # pretty much needed
-install_requires = ['numpy>=1.9', 'qtpy>=1.1', 'setuptools>=1.0',
-                    'astropy>=1.0']
+install_requires = ['numpy>=1.13', 'qtpy>=1.1', 'setuptools>=1.0',
+                    'astropy>=3']
```
pllim: Might as well? Is there a reason to drop PY2 support but retain compat with Astropy as old as v1? Numpy 1.13 bump goes with Astropy bump.

ejeschke:
@pllim, as we discussed, shall I branch off a 2.x and then merge this into master?

PR #726, opened 2018-12-18
Modify Pick plugin to be able to recenter shape on found object

ejeschke:
Enhancement to the Pick plugin to have shape track the pick.
    
This modifies the Pick plugin to have an additional option to center on the pick.  When this setting is enabled, the pick shape is recentered around the located center of the found object.  This is useful when images are updating in the channel and you wish to keep "tracking" the picked object.
    
Some additional changes in this PR:
- Box becomes the default shape, not Rectangle.  Boxes are drawn from the center of the object.  This should be more natural for ds9 users.
- Triangle has been removed as a shape for picks.  It's not a typical choice, and one can always use a Polygon to get a similar (or more useful) shape.


ejeschke:
I will add updated screen shots to the documentation before merging.

ejeschke:
@fmadec and @alefur, if you guys have any feedback after trying this out, let us know here.

ejeschke:
@fmadec and @alefur, any feedback?

alefur:
I've tested this new version and it's exactly what we had in mind.

ejeschke:
Rebased.

ejeschke:
Ok, going to merge after one more test on my end.

PR #728, opened 2019-01-08
BUG: Fix FBrowser drag-and-drop on Windows

### LINKED ISSUE #727: FBrowser drag-and-drop does not work on Windows
I suspect the hardcoding of `file://` here is not compatible with Windows (I was using Windows 10):

https://github.com/ejeschke/ginga/blob/24034716fba4d64375b0f280f5c97d2cb35d1cf5/ginga/rv/plugins/FBrowser.py#L221-L222

Drag-and-drop resulted in this error:
```
2018-12-22 19:15:20,154 | I | FBrowser.py:223 (item_drag_cb) | urls: ['file://C:\\Users\\Documents\\test_data\\jb5g05ubq_flt.fits']
2018-12-22 19:15:20,561 | I | Mosaic.py:416 (drop_cb) | files dropped: ['']
2018-12-22 19:15:20,564 | W | Control.py:579 (load_image) | error determining file type: Can't determine file type of ''; assuming 'image/fits'
2018-12-22 19:15:20,565 | I | Control.py:611 (load_image) | Loading object from kwargs={'numhdu': None, 'inherit_primary_header': True}
2018-12-22 19:15:20,565 | E | Control.py:617 (load_image) | Failed to load file'': File does not appear to be on disk: file://
2018-12-22 19:15:20,576 | E | Control.py:440 (error_wrap) | ControlError
Failed to load file '': File does not appear to be on disk: file://  File "C:\...\ginga\rv\Control.py", line 430, in error_wrap
    return method(*args, **kwargs)
```

As a result, I cannot drag-and-drop a file to create a mosaic via `Mosaic` plugin on Windows. When I used `get_fileinfo()` independently,

https://github.com/ejeschke/ginga/blob/24034716fba4d64375b0f280f5c97d2cb35d1cf5/ginga/util/iohelper.py#L65

it worked with filename set to `"C:\\Users\\Documents\\test_data\\jb5g05ubq_flt.fits"` but not `"file://C:\\Users\\Documents\\test_data\\jb5g05ubq_flt.fits"`.

pllim:
Fix #727

This uses `pathlib`, which is new in Python 3.4. For Ginga 3.0, this is not a problem because its minimum supported Python is 3.5.

### DIFF
--- a/doc/WhatsNew.rst
+++ b/doc/WhatsNew.rst
@@ -10,6 +10,7 @@ Ver 3.0.0 (unreleased)
 - Pick plugin enhanced with option to center on found object; also
   default shape changed to a box rather than a rectangle
 - Added support for ASDF and GWCS.
+- Fixed drag-and-drop functionality in FBrowser plugin on Windows.
 
 Ver 2.7.2 (2018-11-05)
 ======================
--- a/ginga/rv/plugins/FBrowser.py
+++ b/ginga/rv/plugins/FBrowser.py
@@ -31,6 +31,7 @@ directory if closed and then restarted.
 import glob
 import os
 import time
+from pathlib import Path
 
 from ginga.misc import Bunch
 from ginga import GingaPlugin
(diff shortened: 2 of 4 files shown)


ejeschke:
@pllim, is this a fix that we could also apply [here](https://github.com/ejeschke/ginga/blob/master/ginga/util/iohelper.py#L120) ?

pllim:
@ejeschke , `pathlib.Path` should be smart enough so we don't have to worry about OS. Yes, I can use it in other parts of repo too. I'll rebase and add commit when I get the chance.

ejeschke:
@pllim, FFTM when you are ready.

PR #719, opened 2018-11-15
Implement basic interface to ASDF and GWCS

pllim:
This might still need unreleased ASDF and GWCS. It definitely requires Astropy 3.1 or later. But it does work, I propose that we merge this at its current state and improve it in future PRs, as needed. (Limited support is better than zero support?)

Fix #327

cc @hcferguson , @nden , and @drdavella

**TODO:**

- [x] Test with some files: FITS with FITS WCS, ADSF with GWCS, ASDF-in-FITS with GWCS
- [x] Write tests.

**Open Questions:**

* `AstroImage.load_asdf()` -- How to let user specify where data and WCS are in ASDF?
* No way to load JWST data in a meaningful way without JWST data models hidden within JWST pipeline.

**Known issues (future PRs):**

* ~~`Compass` does not work.~~ 
* `WCSAxes` does not work. But this could be an issue with the test image rather than the plugin; need a better dataset to be sure.
* ~~Data cube not fully supported. `LineProfile` does not fully work.~~
* Very limited testing done.
* Documentation deferred to future PR when this is more stable.

PR #731, opened 2019-01-10
Fix wcspt_to_datapt bug for astropy_ape14

### LINKED ISSUE #729: Fix WCSAxes plugin in astropy_ape14
Follow up of #719. Need to investigate why WCSAxes is not drawing RA and DEC lines when WCSpkg is set to `astropy_ape14`. There is no traceback shown in terminal.

pllim:
Fix #729 . Turns out there was a bug in `wcspt_to_datapt` method in the `astropy_ape14` implementation from #719. 😅 

Also adds some logging to `WCSAxes` from the debugging process. No change log is needed as this is fixing something that is not released yet.

### DIFF
--- a/ginga/util/wcsmod/wcs_astropy_ape14.py
+++ b/ginga/util/wcsmod/wcs_astropy_ape14.py
@@ -191,7 +191,7 @@ class AstropyWCS(common.BaseWCS):
         try:
             args = [wcspt[:, i] for i in range(wcspt.shape[1])]
             # NOTE: Ignores system transformation.
-            datapt = np.asarray(self.wcs.world_to_pixel_values(*args))[:, :2].T
+            datapt = np.asarray(self.wcs.world_to_pixel_values(*args))[:2, :].T
         except Exception as e:
             self.logger.error(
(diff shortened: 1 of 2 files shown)


ginga/util/wcsmod/wcs_astropy_ape14.py
```diff
@@ -191,7 +191,7 @@ def wcspt_to_datapt(self, wcspt, coords='data', naxispath=None):
         try:
             args = [wcspt[:, i] for i in range(wcspt.shape[1])]
             # NOTE: Ignores system transformation.
-            datapt = np.asarray(self.wcs.world_to_pixel_values(*args))[:, :2].T
+            datapt = np.asarray(self.wcs.world_to_pixel_values(*args))[:2, :].T
         except Exception as e:
```
ejeschke: This looks like the critical fix.  The rest is just logging changes, I guess.

ejeschke: LGTM, AFAIAC.

pllim: Yes. This line is the bug fix.

Issue #311, opened 2016-03-28
Opening MEF image file


AlexaVillaume:
Hi all,

This is a very basic question but I can't figure out how to open a MEF image such that all the extensions are displayed at the same time. I'm thinking of something similar to "open other" in DS9. Right now when I load the image it just displays one extension. I'd appreciate it if somebody could help me out.

Thank you,
Alexa

pllim:
Use the `MultiDim` local plugin. Then choose the extension(s) you want to open.

AlexaVillaume:
Thanks @pllim for your reply but still a little confused. There's the drop down menu on the left where I can only select one extension at a time and there's the display on the right where, again, I can only go through one extension at a time. What's the secret handshake for getting all the extensions to display at once? 

<img width="1436" alt="screen shot 2016-03-28 at 9 53 29 am" src="https://cloud.githubusercontent.com/assets/5116488/14083613/835deb40-f4cb-11e5-8bd6-2c65dd8cf7e7.png">

pllim:
Good question. `File->Open` and `FBrowser` accepts wildcard in filename to open multiple files at once, but they do not support wildcard in the extension specification (the square brackets).

AlexaVillaume:
Does that mean I cannot open all the extensions of the file at once?

pllim:
There might be a way that I am not aware of. @ejeschke ? Or perhaps @eteq knows how to accomplish this using Jupyter notebook magic?

pllim:
Thinking about this a bit more, I am not sure if it is a desirable feature to have Ginga automatically open all extensions. While it might be convenient in your case, some files have mixed image/table/header format, which means Ginga will fail to load some of them. And also, what if someone provides a file with hundreds of extensions?

ejeschke:
Hi @AlexaVillaume !

It would be quite straightforward to add a button to [`MultiDim`](https://github.com/ejeschke/ginga/blob/master/ginga/misc/plugins/MultiDim.py) that would open all HDU's (ideally just image HDUs, I guess).  `MultiDim` uses `astropy.io.fits` to open the file and assigns this to the instance variable `self.fits_f`.  So the callback for the button would just iterate over then length of this and load all image HDUs.  There is even a function called `set_hdu()` which takes an index and loads the HDU corresponding to that index if it is not already loaded.

If would be great if you'd want to take a stab at a PR for that--it really would be just a few lines of python.  If not, LMK here and I'll do it.

[...]

PR #734, opened 2019-01-14
ENH: Load multiple files or multiple extensions from the same file

### LINKED ISSUE #410: FBrowser does not populate text box with selected filename
Selecting (highlighting) a file from `FBrowser` does not populate the "selected" text box with its name.

In other applications (e.g., Photoshop), the file open dialog automatically populates text box with selected filename. If multiple files are selected, then the text box displays each name separated by ~~comma~~ whitespace.

### LINKED ISSUE #559: Feature Request: Provide options to only auto-load certain extensions from multi-extension FITS
Requested by @jdavies-st . For example, perhaps one of the following (but not limited to these):

1. `ginga myfile.fits[SCI,*]`
2. Recognize `myfile.fits[SCI,*]` syntax in `FBrowser` plugin

Maybe @ejeschke knows of some tricks to do this, but right now such a thing *might* only be possible by coding up some pre-processor in `~/.ginga/ginga_config.py`.

pllim:
Fix #410 

* Update text entry when file is selected.
* Support opening multiple files from text entry box.

Fix #559 

* Enable `ginga filename[EXT,*]` command to open multiple named extensions for a given file.
* Enable opening `filename[EXT,*]` from `FBrowser` plugin.

Address part of #311 (enable opening multiple extensions at once but only if they are named the same). 

With this patch, crazy specifications like the following is now possible both from CLI and `FBrowser`:

```
file1.fits file2*.fits file3.fits[SCI,*] file4.fits[ERR]
```

However, `file.fits[*]` and `file*.fits[SCI,*]` are *not* supported (too crazy?). 

**Update:** The following error will go away if you use `gwcs` and `asdf` from latest `master`.

When testing this patch (Python 3.7.1, Numpy 1.15.4, Astropy 3.2.dev23734), the following combo gave me error (see spacetelescope/asdf#639) but I cannot reproduce it with a different combo nor from `FBrowser` 🤷‍♀️ 

```
$ ginga --loglevel=30 --stderr jb5g05ubq_flt.fits[SCI,*] imaging_wcs.asdf cropped_asdfinfits.fits[ASDF]
2019-01-14 16:07:08,833 | E | Control.py:577 (load_image) | Failed to load file'cropped_asdfinfits.fits': Can't handle 'tag:astropy.org:astropy/coordinates/frames/baseframe-1.0.0' as a file for mode 'r'
```

cc @jdavies-st and @AlexaVillaume

### DIFF
--- a/ginga/rv/Control.py
+++ b/ginga/rv/Control.py
@@ -587,6 +587,12 @@ class GingaShell(GwMain.GwMain, Widgets.Application):
         self.logger.debug("Successfully loaded file into object.")
         return image
 
+    def _dl_indicator(self, count, blksize, totalsize):
+        """Show download status."""
+        pct = float(count * blksize) / float(totalsize)
+        msg = "Downloading: {:.2%} complete".format(pct)
+        self.gui_do(self.show_status, msg)
+
     def get_fileinfo(self, filespec, dldir=None):
         """Break down a file specification into its components.
 
@@ -599,43 +605,32 @@ class GingaShell(GwMain.GwMain, Widgets.Application):
 
         Returns
         -------
-        res : `~ginga.misc.Bunch.Bunch`
+        res : list
+            A list of `~ginga.misc.Bunch.Bunch`.
 
         """
         if dldir is None:
             dldir = self.tmpdir
 
         # Get information about this file/URL
-        info = iohelper.get_fileinfo(filespec, cache_dir=dldir)
+        res = iohelper.get_fileinfo(filespec, cache_dir=dldir)
 
-        if isinstance(info, list):
-            for res in info:
-                if ((not res.ondisk) and (res.url is not None) and
-                        (not res.url.startswith('file:'))):
-                    raise NotImplementedError(
-                        'Cannot download data when wildcard given in extension')
-            return info
+        # Download the file if a URL was passed
+        for info in res:
+            if ((not info.ondisk) and (info.url is not None) and
+                    (not info.url.startswith('file:'))):
 
-        if ((not info.ondisk) and (info.url is not None) and
-                (not info.url.startswith('file:'))):
+                # Try to download the URL.  We press our generic URL server
+                # into use as a generic file downloader.
+                try:
+                    dl = catalog.URLServer(self.logger, "downloader", "dl",
+                                           info.url, "")
+                    filepath = dl.retrieve(info.url, filepath=info.filepath,
+                                           cb_fn=self._dl_indicator)
+                finally:
+                    self.gui_do(self.show_status, "")
 
-            # Download the file if a URL was passed
-            def _dl_indicator(count, blksize, totalsize):
-                pct = float(count * blksize) / float(totalsize)
-                msg = "Downloading: %%%.2f complete" % (pct * 100.0)
-                self.gui_do(self.show_status, msg)
-
-            # Try to download the URL.  We press our generic URL server
-            # into use as a generic file downloader.
-            try:
-                dl = catalog.URLServer(self.logger, "downloader", "dl",
-                                       info.url, "")
-                filepath = dl.retrieve(info.url, filepath=info.filepath,  # noqa
-                                       cb_fn=_dl_indicator)
-            finally:
-                self.gui_do(self.show_status, "")
-
-        return info
+        return res
 
     def name_image_from_path(self, path, idx=None):
         return iohelper.name_image_from_path(path, idx=idx)
@@ -686,9 +681,6 @@ class GingaShell(GwMain.GwMain, Widgets.Application):
             image_loader = self.load_image
 
         res = self.get_fileinfo(filepath)
-
-        if not isinstance(res, list):
-            res = [res]
 
         for info in res:
             filepath = info.filepath
@@ -2295,9 +2287,10 @@ class GingaShell(GwMain.GwMain, Widgets.Application):
         # Fancy load (first file only)
         # TODO: If load all the matches, might get (Qt) QPainter errors
         else:
-            info = iohelper.get_fileinfo(filename)
-            if isinstance(info, list):
+            res = iohelper.get_fileinfo(filename)
+            if len(res) != 1:
                 raise NotImplementedError('Wildcard in extension not supported')
+            info = res[0]
             ext = iohelper.get_hdu_suffix(info.numhdu)
             paths = ['{0}{1}'.format(fname, ext)
                      for fname in glob.iglob(info.filepath)]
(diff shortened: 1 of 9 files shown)


ginga/rv/Control.py
```diff
@@ -608,6 +608,14 @@ def get_fileinfo(self, filespec, dldir=None):
         # Get information about this file/URL
         info = iohelper.get_fileinfo(filespec, cache_dir=dldir)
 
+        if isinstance(info, list):
```
ejeschke: I am wondering if we should just make `get_fileinfo()` *always* return a list.  If there is only one file selected, it would return a list of one bunch.  I think this makes more sense than having to check the results everywhere that it is used.

ginga/rv/Control.py
```diff
@@ -676,47 +685,52 @@ def load_file(self, filepath, chname=None, wait=True,
         if image_loader is None:
             image_loader = self.load_image
 
-        info = self.get_fileinfo(filepath)
-        filepath = info.filepath
+        res = self.get_fileinfo(filepath)
```
ejeschke: Then that would simplify the following logic.

ginga/rv/plugins/FBrowser.py
```diff
@@ -210,20 +214,29 @@ def makelisting(self, path):
...
-        return path
+    def get_paths_from_item(self, res_dict):
+        return [info.path for info in res_dict.values()]
 
     def item_dblclicked_cb(self, widget, res_dict):
-        path = self.get_path_from_item(res_dict)
-        self.open_file(path)
+        paths = self.get_paths_from_item(res_dict)
+        if len(paths) > 0:
+            self.open_file(paths[0])
```
ejeschke: Hmm, maybe raise an error if `len(paths) > 0`?  There probably shouldn't be more than one if double clicking a file.

ginga/rv/plugins/FBrowser.py
```diff
@@ -250,25 +263,40 @@ def file_icon(self, bnch):
...
+            if path.endswith('*'):
+                path = path[:-1]
 
-        self.load_paths(paths)
-        return True
+            if os.path.isdir(path):
+                continue
+
+            self.logger.debug('Opening files matched by {0}'.format(path))
+            res = iohelper.get_fileinfo(path)
```
ejeschke: ditto on list comment--simplifies following logic to assume result is a list.

ginga/util/io_fits.py
```diff
@@ -189,6 +189,8 @@ def load_file(self, filespec, numhdu=None, dstobj=None, memmap=None,
     def open_file(self, filespec, memmap=None, **kwargs):
 
         info = iohelper.get_fileinfo(filespec)
+        if isinstance(info, list):
+            raise NotImplementedError('Wildcard in extension not supported')
```
ejeschke: so instead, something like:
```python
specs = iohelper.get_fileinfo(filespec)
if len(specs) > 0:
    raise NotImplementedError('Wildcard in extension not supported')
```

ginga/util/iohelper.py
```diff
@@ -87,10 +87,35 @@ def get_fileinfo(filespec, cache_dir='/tmp', download=False):
...
             hduname, extver = idx.split(',')
-            hduname = hduname.strip()
-            extver = int(extver)
-            idx = (hduname, extver)
-            name_ext = "[%s,%d]" % idx
+            hduname = hduname.strip().upper()
+
+            # User trying to match extver with wildcard!
+            # NOTE: Only use Astropy FITS; Only works on local file.
+            if extver == '*':
```
ejeschke: Can we do something with `io_fits` instead?  Something like:
```python
opener = io_fits.get_fitsloader(logger=logger)
try:
    opener.open_file(filespec)
    if opener.hdu_info is not None:
        for hduinfo in opener.hdu_info:
            ...
finally:
    opener.close()
```

Here is an example of the contents of `opener.hdu_info`:
```python
In [11]: opener.hdu_info
Out[11]: 
[{'index': 0, 'name': 'PRIMARY', 'extver': 1, 'htype': 'PrimaryHDU', 'dtype': ''},
 {'index': 1, 'name': 'SCI', 'extver': 1, 'htype': 'ImageHDU', 'dtype': 'float32'},
 {'index': 2, 'name': 'ERR', 'extver': 1, 'htype': 'ImageHDU', 'dtype': 'float32'},
 {'index': 3, 'name': 'DQ', 'extver': 1, 'htype': 'ImageHDU', 'dtype': 'int16'},
 {'index': 4, 'name': 'SCI', 'extver': 2, 'htype': 'ImageHDU', 'dtype': 'float32'},
 {'index': 5, 'name': 'ERR', 'extver': 2, 'htype': 'ImageHDU', 'dtype': 'float32'},
 {'index': 6, 'name': 'DQ', 'extver': 2, 'htype': 'ImageHDU', 'dtype': 'int16'},
 {'index': 7, 'name': 'WCSCORR', 'extver': 1, 'htype': 'BinTableHDU', 'dtype': '[40A, I, 1A, 24A, 24A, 24A, 24A, D, D, D, D, D, D, D, D, 24A, 24A, D, D, D, D, J, 40A, 128A]'},
 {'index': 8, 'name': 'WCSDVARR', 'extver': 1, 'htype': 'ImageHDU', 'dtype': 'float64'},
 {'index': 9, 'name': 'WCSDVARR', 'extver': 2, 'htype': 'ImageHDU', 'dtype': 'float64'},
 {'index': 10, 'name': 'WCSDVARR', 'extver': 3, 'htype': 'ImageHDU', 'dtype': 'float64'},
 {'index': 11, 'name': 'WCSDVARR', 'extver': 4, 'htype': 'ImageHDU', 'dtype': 'float64'}]
```

ejeschke: This keeps the abstraction of the fits opener.

ginga/rv/plugins/FBrowser.py
pllim: Is it possible for someone with touchscreen _and_  keyboard to "double click" two spots on the monitor at the same time? But regardless, raising exception makes sense here.

ginga/rv/Control.py
```diff
@@ -587,6 +587,12 @@ def load_image(self, filepath, idx=None, show_error=True):
         self.logger.debug("Successfully loaded file into object.")
         return image
 
+    def _dl_indicator(self, count, blksize, totalsize):
```
pllim: @ejeschke , there is no reason why this cannot be a proper class method, right? It was previously buried in another method below.

ejeschke:
@pllim, can I take a swing at editing this PR?  The conversation here is making me rethink how this is structured.  I think `get_fileinfo()` should only be returning information about a file specification, and not downloading anything and not opening any file--and probably not returning a list after all.  Now that we have a `loader` module, it makes more sense to me that we would do any downloading there and do any parsing of the HDU spec in the `io_fits` module.  Doing it that we we can probably support a wildcard as well.

pllim:
@ejeschke , go for it! Thanks!

pllim:
Superseded by #742

PR #735, opened 2019-01-14
ENH: Enable HDU sorting in MultiDim

### LINKED ISSUE #556: Feature Request: Add option to have MultiDim sort extensions by name
This is a feature request by @jdavies-st . Currently, the drop-down menu sort extensions by number. I think it will be easy to add an user configurable item in `plugin_MultiDim.cfg` to enable this.

### LINKED ISSUE #557: Feature Request: MultiDim to auto start if multi-extension FITS is loaded
As requested by @jdavies-st -- If user loads a multi-extension FITS file, the loader can automatically starts `MultiDim` local plugin.

p.s. An easier alternative is just to have `MultiDim` set to always start when a channel is started.

pllim:
Fix #556 

Close #557 

cc @jdavies-st

### DIFF
--- a/ginga/rv/plugins/MultiDim.py
+++ b/ginga/rv/plugins/MultiDim.py
@@ -67,6 +67,7 @@ class MultiDim(GingaPlugin.LocalPlugin):
     def __init__(self, fv, fitsimage):
         # superclass defines some variables for us, like logger
         super(MultiDim, self).__init__(fv, fitsimage)
+        self._toc_fmt = "%(index)4d %(name)-12.12s (%(extver)3d) %(htype)-12.12s %(dtype)-8.8s"
 
         self.curhdu = 0
         self.naxispath = []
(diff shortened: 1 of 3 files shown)


ginga/rv/plugins/MultiDim.py
```diff
@@ -91,7 +92,8 @@ def __init__(self, fv, fitsimage):
         # Load plugin preferences
         prefs = self.fv.get_preferences()
         self.settings = prefs.create_category('plugin_MultiDim')
-        self.settings.set_defaults(auto_start_naxis=False)
```
pllim: @Cadair , just want to double check that removing this `auto_start_naxis` option does not break anything that SunPy is doing. It is not used here in Ginga. If you use it on your side, I would recommend to subclass this plugin instead and add the option over there.

pllim: Speak now for forever hold your peace! :wink:

ejeschke:
FWIW, I have been thinking that we should add an icon to the toolbar for opening `MultiDim`, since it is an oft-used plugin like `Preferences` or `FBrowser`...

jdavies-st:
Tested this with some JWST files, and it looks good!

pllim:
As per offline conversation with @jdavies-st , this is good to merge. Thanks, James and Eric!

PR #737, opened 2019-01-17
Add buttons to Toolbar plugin

ejeschke:
- added buttons to launch MultiDim, Header and Zoom plugins


pllim:
@jdavies-st , would this new button be a good compromise if we forego autostart of `MultiDim`?

pllim:
@ejeschke , I merged #738, so if you rebase now, the failure would go away. FYI.

PR #738, opened 2019-01-18
TST: Fix import failure from astropy.tests

pllim:
This should fix the failure with astropy dev job. Please double check Travis CI results before merge.


pllim:
Since Ginga 3.0 only needs `astropy>=3`, we don't have to worry about backward compatibility for this PR. These new imports have been there for a while.

PR #740, opened 2019-01-27
Update maintainers info and copyright for year 2019

ejeschke:
Update maintainers info and copyright for year 2019.  

Also fixes a broken load of banner using Menu->About.


ejeschke:
@pllim, these changes ok with you?

ginga/util/loader.py
```diff
@@ -118,5 +118,6 @@ def add_loader(mimetype, loader):
     add_loader(mimetype, load_asdf)
 
 # ### RGB ###
-for mimetype in ['image/jpeg', 'image/png', 'image/tiff', 'image/gif']:
+for mimetype in ['image/jpeg', 'image/png', 'image/tiff', 'image/gif',
+                 'image/ppm', 'image/pnm', 'image/pbm', 'image/pnm']:
```
pllim: `pnm` was listed twice. Did you mean something else?
```suggestion
                 'image/ppm', 'image/pnm', 'image/pbm']:
```

pllim:
p.s. What is `icons/ginga.fits` used for?

ginga/util/loader.py
hcferguson: I think it's fine to include a mention of STScI there. Thanks!

ejeschke: Good catch, @pllim.  I think that is in there in error.  I'll update the PR.

ejeschke:
> p.s. What is icons/ginga.fits used for?

That's a good question!  It looks like it was committed quite a while ago.  My guess is that at some point I thought it would be nice to include one fits file with the package for testing. If we get around to adding more test files, we can properly stick it in a directory or get rid of it.

PR #741, opened 2019-01-27
Add buttons to Toolbar plugin to restore color map, pick color map

ejeschke:
Fairly simple update to Toolbar to allow the user to restore the colormap if they have rotated or inverted it.


pllim:
@ejeschke , not sure if I understand what it means to rotate or invert a colormap (and not the image).

ejeschke:
Well it may be a case of a solution looking for a problem, but we had one instance where a user typed `Shift+i` (capital i) in the window and inverted the colormap.  But he didn't realize that he had done so and could not figure out how to reset the color map.  So I thought maybe we could use a button somewhere to allow that rather simply.

Here are the current ways you can reset it:
- Right click in the color bar
- Go into "cmap" mode by pressing Space, 'y' and then pressing 'r'.  Press 'Esc' to exit cmap mode.

pllim:
That "launch colormap plugin" button is quite convenient. However, the "reset colormap" is a little misleading for me. When I see "reset", I think it was to go back to the default colormap (e.g., the gray scale that I started out with before I played with different options), but it is not the case here.

ejeschke:
> When I see "reset", I think it was to go back to the default colormap (e.g., the gray scale that I started out with before I played with different options), but it is not the case here.

That's a good point.  And in fact there is *that* kind of "reset"--when you press `Shift+y`.

Basically, you can do the following things:
1. Set a new color map, which resets everything related to the color map (BUT NOT the contrast)
2. Invert the color map, or rotate it, which results in a new color map (doesn't change contrast)
3. Do a contrast operation, which is essentially a shift+stretch of the color map

The question is what buttons should there be and what should they do?  We currently have the one button that is a "reset" of 3.  The proposed button "resets" 2 AND 3, resulting in a "sane state" of the color map, albeit the last one that you have selected from the list, not a default one.

ejeschke:
Maybe the button should reset everything, including setting the default color map and also the color distribution (linear, log, etc)?  A kind of "panic reset" button to set sane defaults for everything color?

pllim:
@ejeschke , they are all good questions but they are best answered by users.

ejeschke:
> they are all good questions but they are best answered by users

Right, but we should maybe narrow it down to a very short list of reasonable possibilities, or face the "too many choices" problem.

pllim:
I still think it should be simple and simply reset back to gray scale (the one I started out with). But I am only one data point...

ejeschke:
Rebased.

ejeschke:
Well, everything still works as described here, but maybe punt to release 3.2 ?  Give us a little more time to think about what kind of "color reset" buttons there should be and what they should do...

pllim:
@ejeschke , I agree and I took the liberty to update the milestone. Thanks!

ejeschke:
Rebased

ejeschke:
@pllim, I have thought about this a little bit and have decided to simply make it about adding the "launch colormap plugin" icon to the toolbar.  This will allow the user to quickly reset/restore the color map to whatever one they want. A caveat is that this doesn't restore the contrast (stretch and shift), so that one must also click the "Restore contrast" icon to reset that.

We can rethink this issue in a future PR, if needed.

pllim:
Sounds good, but looks like you need a rebase. 😸

ejeschke:
Rebased

pllim:
Ops, looks like there is another conflict.

ejeschke:
Rebased

ejeschke:
Closing without merge.  This can be revisited in the future if there is interest.

PR #761, opened 2019-03-12
WIP: Refactor Pan, Header and Info plugins not to stack widgets

ejeschke:
This refactors the Pan, Info and Header plugins to create one instance of widgets and update it when the channel changes, rather than creating a separate container full of widgets for each channel and stacking them on top of each other, popping one to the top when the channel changes.

The latter arrangement seems overly complex.  It does seem to provide a little bit of time savings when switching focus between channels, but the former approach is easier to understand and maintain.

We should measure the performance of the channel switch and if it is acceptable with this PR, it would be a worthwhile merge for maintainability.  The best test is doing a "blink channels" in tab view.


ejeschke:
@pllim, no hurry on this at all.  Sometime when you have a little extra time try to compare a blink channels between this PR and master.  See if you can notice any particular difference.

pllim:
```
ginga/rv/plugins/Info.py:45:1: F401 'ginga.misc.Bunch' imported but unused
1
```

ejeschke:
There is no hurry on this one.  Let's let it sit for a bit and consider after 3.0.

ejeschke:
Rebased

ejeschke:
Rebased

ejeschke:
Just tested Blink with some very short ~0.3 sec transitions. Definitely faster and smoother *without* this PR.  Not going to close it just yet though.

ejeschke:
Rebased.

PR #763, opened 2019-03-15
Use JSON-formatted layout and plugin files

ejeschke:
This PR changes the default format for the layout file to JSON.  It also introduces a JSON-formatted plugin file.  Between the two files users have a straightforward way to customize the reference viewer extensively.

In addition, the reference viewer can now be started with a `--basedir` option in which you can specify alternate directories to `$HOME/.ginga`.  In this way, a user can create different layouts and sets of plugins for different uses.


ejeschke:
@pllim, wait a bit before review. I want to add some documentation, as well as some example alternative layouts in the `ginga/examples` directory.

pllim:
Okay, I'll wait. Please let me know when this is ready.

ejeschke:
Rebased

ejeschke:
Rebased.

ejeschke:
Rebased.

ejeschke:
Rebased.

pllim:
To follow-up on offline discussion, I am awaiting feedback from @ejeschke on whether this changes any existing behavior or not. If this would break downstream packages (e.g., `stginga`), I would prefer to re-milestone to 3.2.

ejeschke:
@pllim, so all this PR does operationally is change the format of the layout file that is stored in the user's $HOME/.ginga area from Python to JSON.  It does not change or require the layout format used in ginga or stginga `main` to be JSON.  As such, it shouldn't have any effect on any layout or plugins that you define there.

To test with the PR, first remove any `layout.json` file in your $HOME/.ginga.  Then start the program as usual, and once it is up, resize it a bit to a new dimension and then close it normally. You should now see a new `layout.json` file in your $HOME/.ginga.  If you start the program again and it starts in the size and position that you closed it, it is working correctly.

ginga/examples/configs/general.cfg
```diff
@@ -96,10 +96,15 @@ icc_working_profile = "sRGB"
 # NOTE: overridden by channel setting of the same name, if any
 scrollbars = 'auto'
 
+# Name of a layout file in ~/.ginga that specifies the UI layout of the
+# program (default is 'layout.json').  This is only used if 'save_layout'
+# (see below) is True
+layout_file = 'layout.json'
```
pllim, line 102: Currently, there is already a `~/.ginga/layout` file. Does that mean `~/.ginga/layout` will be ignored with this patch? If not, what will Ginga do when it finds both `layout` and `layout.json`?

ginga/rv/main.py
```diff
@@ -231,6 +232,8 @@ def add_default_options(self, argprs):
             # newer ArgParse
             add_argument = argprs.add_argument
 
+        add_argument("--basedir", dest="basedir", metavar="NAME",
```
pllim, line 235: Does this also change where Ginga look for other things like `ginga_config.py` and `*.cfg` files?

ginga/rv/main.py
```diff
@@ -323,6 +329,8 @@ def main(self, options, args):
                               font_scaling_factor=None,
                               save_layout=True,
                               use_opengl=False,
+                              layout_file='layout.json',
+                              plugin_file='plugins.json',
```
pllim, line 333: `plugins.json` looks like a new feature but I cannot find enough documentation to understand what it does.

ginga/util/json.py
```diff
@@ -0,0 +1,54 @@
...
+
+from ginga.misc import Bunch
+
+
+class BunchEncoder(json.JSONEncoder):
+    """Custom encoder to serialize Ginga's Bunch.Bunch class.
+
+    Usage
+    -----
+      json.dumps(objs, indent=2, cls=BunchEncoder)
```
pllim, line 17:
```suggestion
    json.dumps(objs, indent=2, cls=BunchEncoder)
```

ginga/util/json.py
```diff
@@ -0,0 +1,54 @@
...
+            return d
+        return json.JSONEncoder.default(self, obj)
+
+
+def as_bunch(dct):
+    """Custom decoder to deserialize Ginga's Bunch.Bunch class.
+
+    Usage
+    -----
+      json.loads(buf, )
```
pllim, line 33: Is something supposed to happen after the comma?
```suggestion
    json.loads(buf, )
```

ginga/util/json.py
```diff
@@ -0,0 +1,54 @@
+#
+# json.py -- augment JSON parsing
+#
+# This is open-source software licensed under a BSD license.
+# Please see the file LICENSE.txt for details.
+#
+import json
+
+from ginga.misc import Bunch
+
```
pllim, line 10:
```suggestion

__all__ = ['BunchEncoder', 'as_bunch', 'dumps', 'loads']
```

pllim:
Would like some clarifications, as stated above, but otherwise seems to work. Thanks!

ginga/examples/configs/general.cfg
ejeschke: If there are both files, Ginga will prefer the `layout.json`.

ginga/rv/main.py
ejeschke: Yes, so you can have multiple "configurations" for Ginga just by creating different directories and using this option. Each one is completely separate.

ginga/rv/main.py
ejeschke: It's just the list of plugins, in JSON format. If present, it will override the built in list (in `main`). Currently there is nothing in Ginga to write out this file, but that would be something to possibly add in the future (e.g. allow users to easily customize their plugins). I am going to provide 2 or 3 example configurations with plugin lists.

ginga/util/json.py
ejeschke: Whups, yes. I'll add a commit to fix this.

ginga/util/json.py
```diff
@@ -0,0 +1,55 @@
...
+            return d
+        return json.JSONEncoder.default(self, obj)
+
+
+def as_bunch(dct):
+    """Custom decoder to deserialize Ginga's Bunch.Bunch class.
+
+    Usage
+    -----
+    json.loads(buf, )
```
pllim, line 34: Just comment was marked as resolved but I want to point out that stuff is still missing after the comma. 😸

ejeschke: At least one more commit coming with the fix!

ginga/examples/configs/general.cfg
ejeschke: > If there are both files, Ginga will prefer the layout.json

So, with this latest commit, I've reversed course on this one: the `layout` is preferred over `layout.json`. I just think that for this file it's easier to craft Python layouts.

ginga/rv/main.py
ejeschke: > `plugins.json` looks like a new feature but I cannot find enough documentation to understand what it does.

@pllim, can you check out the new documentation in the manual page on customization? See if it explains the layout and plugins files to your satisfaction.

ginga/examples/configs/general.cfg
pllim: Okay, but eventually there should only be one way to do this. Otherwise, it is confusing.

doc/manual/customizing.rst
```diff
@@ -49,49 +89,202 @@ inherit them.  You can also manually copy the example file from
...
+            "ptype": "global"
+       },
+       ...
+   ]
+    
+                 
+The keys for each object are defined as follows:
+
+* ``__bunch__``: should be present and set to ``true`` to force
+  deserialization to a ``~ginga.misc.Bunch.Bunch``.
```
pllim, line 236:
```suggestion
  deserialization to a `~ginga.misc.Bunch.Bunch`.
```

ginga/rv/main.py
pllim: So, the "standard layout" example in https://ginga--763.org.readthedocs.build/en/763/manual/customizing.html#format-of-the-layout-table is the Python one, right? Not JSON? Maybe should be clarified.

I also think example `plugin.json` should be formatted similarly to https://ginga--763.org.readthedocs.build/en/763/manual/customizing.html#customizing-the-set-of-plugins ; i.e., the line breaks. I like the extra line breaks in the doc example.

doc/manual/customizing.rst
```diff
@@ -49,49 +89,202 @@ inherit them.  You can also manually copy the example file from
...
+* ``module``: The name of the module in the ``$PYTHONPATH`` containing
```
pllim, line 237: Is it only PYTHONPATH? What about those in `~/.ginga/plugins`?

doc/manual/customizing.rst
```diff
@@ -49,49 +89,202 @@ inherit them.  You can also manually copy the example file from
...
+* ``tab``: the name that the plugin should appear as when opened in a
+  workspace (usually as a tab, but it depends on the type of
+  workspace). Often the same name as the class, but can be different.
+  If not present, defaults to the class or module name of the plugin.
+* ``workspace``: the name of the workspace defined in the layout file
+  (or default layout) where the plugin should be started (see section
+  below on workspace customization).
+* ``start``: ``true`` if the module is of the global type and should
+  be started at program startup. Defaults to ``false``.
+* ``hidden``: ``true`` if the plugin should be hidden from the
```
pllim, line 251: I wonder if `static` is a better name than `hidden`... But maybe that ship has sailed.

doc/manual/customizing.rst
```diff
@@ -49,49 +89,202 @@ inherit them.  You can also manually copy the example file from
...
+  "Operation" and "Plugins" menus. Often paired with ``hidden`` being
+  ``true`` for plugins that are considered to be a necessary part of
+  continuous operation from program startup to shutdown. Defaults to
+  ``false``. 
+* ``category``: an arbitrary organizational name under which plugins
+  are organized in the ``Operation`` and ``Plugins`` menus.
+* ``menu``: a name for how the plugin should appear under the category
+  in the menu structure.  The convention is to append " [G]" if it is
```
pllim, line 259:
```suggestion
  in the menu structure.  The convention is to append "[G]" if it is
```

doc/manual/customizing.rst
```diff
@@ -101,67 +294,12 @@ can be used to define a different viewer layout, add or remove plugins,
...
 
-Ginga has a flexible table-driven layout scheme for dynamically creating
-workspaces and mapping the available plugins to workspaces.  By changing
-a couple of tables via ``ginga_config.pre_gui_config()`` you can change
-the way Ginga looks and presents its content.
-
-If you examine the module `ginga.rv.main` you will find a layout table
-called ``default_layout``::
-
-    default_layout = ['seq', {},
```
pllim, line 116: So setting `default_layout` like this will no longer work?

ginga/rv/main.py
ejeschke: I'm ok with that. Makes it a lot longer though.

doc/manual/customizing.rst
ejeschke: @pllim, what is the guidance on using double vs. single backticks? I could not get a clear answer from the internet.

doc/manual/customizing.rst
ejeschke: `~/.ginga/plugins` is added to the  `PYTHONPATH`. I think this is explained a little further in the document.

doc/manual/customizing.rst
ejeschke: I don't understand "static" for being left out of the menus, but maybe there is a better keyword. We could consider a rename for a later update but "hidden" has been used for a couple of releases now ... :ship: :smile_cat:

doc/manual/customizing.rst
ejeschke: No, it should work as before, if you are modifying from Python code, and assign over that variable.  There is also an example at the end of the customization page showing a way to make a customized startup script where you set your own layout from Python without having to know the global variable names from the `main` module.

doc/manual/customizing.rst
pllim: Single to link. Double to use preformatted font without linking. :)

doc/manual/customizing.rst
pllim: From reading https://ginga--763.org.readthedocs.build/en/763/manual/customizing.html#customizing-the-layout , it is still unclear to me if I have `default_layout` defined as well as the JSON/Python, which one will take precedence.

doc/manual/customizing.rst
pllim: Yeah, okay. Let's table this one. Thanks for the clarification!

doc/manual/customizing.rst
pllim: Ah, okay. I see it at https://ginga--763.org.readthedocs.build/en/763/manual/customizing.html#custom-plugin-directory . Thanks!

doc/manual/customizing.rst
ejeschke: If there is no custom layout defined, it will use the default layout.  But note that every time you close the program it saves the layout which becomes a custom layout (unless, as the documentation notes: you have set `save_layout = False` in your `general.cfg` or started the program with `--norestore`). I'll try to clarify this in the docs.

ejeschke: @pllim, I think I've clarified this in the `customizing.rst`; can you read it over and LMK what you think.

ejeschke:
Latest commit should address most of the remaining points. @pllim, you should try the two example layouts in `.../examples/layouts/ds9ish` and `.../examples/layouts/twofer`.  Just fire up Ginga and add the option `--basedir=` with either of those folders.  I think it kind of shows off the flexibility provided by this PR.

ginga/examples/layouts/twofer/README.md
```diff
@@ -0,0 +1,14 @@
+### About
+
+This directory defines a set of files that provide a riff on the standard
+look and feel of the reference viewer layout called "twofer".  It provides
+a second large pane on the bottom to take some of the plugins that would
```
pllim, line 5: Maybe given an example or two of which plugin would utilize the second large pane? I tried `Histogram` but does not seem like that plugin uses it.

ejeschke: @pllim, see the `screenshot.png` in the `twofer` directory.  The plugins that will open there are `Pan`, `Info`, `Errors`, `Command`, `MultiDim`, `RC`, `SAMP`, `FBrowser` ([G]) and `Zoom`.

ejeschke: > Or is ~ expansion not supported?

I think that is a host shell issue, but I am not sure. Was this on Linux? Normally `bash` expands those when issued from the command line, so the program invoked doesn't see that unless it is dealing with a path typed directly into an input opened by the program. Not sure how that handled on Windows.

ejeschke:
@pllim, ok, I am finally merging this then. We can deal with follow-up PRs as needed. Thanks for your thorough review!

ginga/examples/layouts/twofer/README.md
pllim: > Was this on Linux?

Yes. I don't think Windows understand `~` natively.

PR #775, opened 2019-06-28
Add splitter to PlotTable plugin, similar to other plugins with plots

ejeschke:
This adds a splitter to the PlotTable plugin, so that the user can adjust the height of the plot (they can already adjust the width). This is similar to what the other plotting plugins do.


ejeschke:
@pllim, when you have time.  Not urgent.

ejeschke:
Good point!  Do you want to make it with the FITS data that was used to make the existing plot? I think you can just push a commit to this PR.  If not I will make a plot with new data.

pllim:
@ejeschke , feel free to update the plot with your own data. I think I just used a Vega spectrum and zoom into one of the absorption lines; nothing too fancy.

ejeschke:
@pllim, I updated the screen shot.  Let me know what you think.  We don't deal with too many tables here so I don't have a lot of examples available.

pllim:
In that case, I can submit an alternate screenshot, but can't do it today. 🤔

ejeschke:
No worries.  Take a look at the example I submitted.  I think your data looks more interesting, but I'm ok with either.  I included the channel viewer in the screenshot (here, a `TableViewer`) showing the columns plotted.  I think that is more in line with what we have been doing with the screenshots lately.

pllim:
@ejeschke , if I push directly to this PR, I would need to squash your image out. Would you mind?

ejeschke:
Thanks!

PR #779, opened 2019-07-12
Add statistics calculation to Histogram plugin

ejeschke:
This adds a line of simple statistics calculation to the Histogram plugin.  The line appears just below the plot. We include a new screenshot showing the line, explanation in the docstring, and example settings in the configuration file.

This was added based on a in-person request from a user.


ejeschke:
@pllim, when you have time.  This is a simple one, I think.

ejeschke:
histogram-plugin

ejeschke:
Updated the `WhatsNew` document.  We can squash when merging.

ejeschke:
Updated screen shot to be smaller and squash rebased into a single commit.

PR #764, opened 2019-03-30
WIP: (DO NOT MERGE) Refactor io handlers

ejeschke:
A first step to refactoring the IO (file) handling.  The main idea is to prevent the increasing complexity of AstroImage/AstroTable classes (due to supporting more and more formats) by moving the file opening and handling code exclusively to the individual modules that deal with those formats (e.g. io_fits, io_rgb).

Code should no longer use:
```python
from ginga import AstroImage
image = AstroImage()
image.load_file(path)
```
But instead:
```python
from ginga.util.loader import load_file
image = load_file(path)
```
Or, if you know that the file is a FITS file, for example:
```python
from ginga.util.io_fits import load_file
image = load_file(path)
```

PR #781, opened 2019-07-27
EXP: Support JWST ASDF-in-FITS in reference viewer

pllim:
~Need to discuss how to best generalize the ASDF loading in reference viewer. This might change depending on how #764 goes (currently does not depend on it).~ Will just roll with this and follow up in a separate PR after #764 is rebased on this and merged.

These are the changes necessary for Python in Astronomy 2019 demo using Ginga standalone GUI (reference viewer) to view ASDF-in-FITS data file generated by JWST. Follow up of #719. Also see spacetelescope/stginga#155

cc @jdavies-st (in case you are interested), @hcferguson , and @eteq 

~Figure out why compass in `Pan` appears weird for the ASDF extension, but not FITS. Maybe spacetelescope/gwcs#241 ?~

**Future work:**

- Ginga should not emit error when RA or Dec is `NaN`, but instead silently format the string to something equivalent (the default "BAD WCS" is probably a little misleading in this case). See #782 
- Figure out a better way to support ASDF loader customization.

### DIFF
--- a/ginga/util/wcsmod/wcs_astropy_ape14.py
+++ b/ginga/util/wcsmod/wcs_astropy_ape14.py
@@ -226,6 +226,10 @@ class AstropyWCS(common.BaseWCS):
         """
         if self.coordsys == 'raw':
             raise common.WCSError("No usable WCS")
+        elif self.coordsys == 'world':
+            coordsys = 'icrs'
+        else:
+            coordsys = self.coordsys
 
         if system is None:
             system = 'icrs'
@@ -233,8 +237,7 @@ class AstropyWCS(common.BaseWCS):
         # Get a coordinates object based on ra/dec wcs transform
         wcspt = self.datapt_to_wcspt(datapt, coords=coords,
                                      naxispath=naxispath)
-        frame_class = coordinates.frame_transform_graph.lookup_name(
-            self.coordsys)
+        frame_class = coordinates.frame_transform_graph.lookup_name(coordsys)
         ra_deg = wcspt[:, 0]
         dec_deg = wcspt[:, 1]
         coord = frame_class(ra_deg * u.degree, dec_deg * u.degree)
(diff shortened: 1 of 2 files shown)


ginga/util/wcsmod/wcs_astropy_ape14.py
```diff
@@ -226,15 +226,18 @@ def datapt_to_system(self, datapt, system=None, coords='data',
         """
         if self.coordsys == 'raw':
             raise common.WCSError("No usable WCS")
+        elif self.coordsys == 'world':
```
pllim: This is necessary for `WCSAxes` plugin to work for GWCS.

ginga/AstroImage.py
```diff
@@ -815,10 +815,10 @@ def info_xy(self, data_x, data_y, settings):
                 # log traceback, if possible
                 (type_, value_, tb) = sys.exc_info()
                 tb_str = "".join(traceback.format_tb(tb))
-                self.logger.error("Traceback:\n%s" % (tb_str))
+                self.logger.warning("Traceback:\n%s" % (tb_str))
             except Exception:
                 tb_str = "Traceback information unavailable."
-                self.logger.error(tb_str)
+                self.logger.warning(tb_str)
```
ejeschke: @pllim, any particular reason for changing the logging of this traceback to a warning?

pllim: Yeah, this was a leftover from my attempt to silence the bad WCS calculations, as GWCS returns `nan` for RA and Dec for pixels outside of image bounds. It didn't really work as I thought it would anyway, so I will go ahead and revert this change here. Thanks for catching it.

pllim:
@ejeschke , given your approval above, I am going to take the liberty to merge this in. I have removed the unnecessary changing from `error` to `warning` for bad WCS. Thanks!

PR #783, opened 2019-07-31
Remove Gtk 2 support

### LINKED ISSUE #780: Deprecate GTK2 support
Since GTK2 is not supported for Python 3, and Python 2 support has been removed from Ginga, we can transitively posit that we should remove support for GTK2 from dev version of Ginga.

ejeschke:
This removes Gtk 2 support, since that only works under Python 2 and Ginga is now Python 3 only.

EDIT: Fix #780


ejeschke:
No urgency, @pllim.  This just removes some stuff and updates docs.  But I recommend we merge before releasing 3.0, since 3.0 will be Python 3 only.

ejeschke:
Fixes #780

ejeschke:
> You might also want to remove GTK2 option from ginga/examples/configs/general.cfg.

Good catch. Tnx!

PR #790, opened 2019-08-09
Refactor the way viewers initialize renderer 

ejeschke:
This refactors the way viewers initialize their initial renderer. It allows for a flattening of the image viewer class hierarchy.  Some redundant code is removed.  Dynamic switching of the renderer in a live viewer is supported via the `choose_renderer` method.

The `cairo` renderer is made available to the Pg and Jupyter viewers, if the user has the `cairo` (`pycairo`?) package installed.  This renderer provides the most supported features (e.g. rotated text) and nicest text rendering for those viewers.

### DIFF
--- a/ginga/ImageView.py
+++ b/ginga/ImageView.py
@@ -3155,7 +3155,8 @@ class ImageViewBase(Callback.Callbacks):
 
     def reschedule_redraw(self, time_sec):
         """Reschedule redraw event.
-        This must be implemented by subclasses.
+
+        This should be implemented by subclasses.
 
         Parameters
         ----------
@@ -3163,7 +3164,9 @@ class ImageViewBase(Callback.Callbacks):
             Time, in seconds, to wait.
 
         """
-        self.logger.warning("Subclass should override this abstract method!")
+        # subclass implements this method to call delayed_redraw() after
+        # time_sec
+        self.delayed_redraw()
 
     def set_cursor(self, cursor):
         """Set the cursor in the viewer widget.
@@ -3270,14 +3273,24 @@ class ImageViewBase(Callback.Callbacks):
         """
         self.set_cursor(self.cursor[cname])
 
+    def configure_surface(self, width, height):
+        """Reconfigure the renderer for a new size, then reconfigure
+        our viewer for the same.
+
+        This can be overridden by subclasses.
+        """
+        self.renderer.resize((width, height))
+
+        self.configure(width, height)
+
     def get_image_as_array(self):
         """Get the current image shown in the viewer, with any overlaid
         graphics, in a numpy array with channels as needed and ordered
         by the back end widget.
 
-        This should be implemented by subclasses.
+        This can be overridden by subclasses.
         """
-        raise ImageViewError("Subclass should override this abstract method!")
+        return self.renderer.get_surface_as_array(order=self.rgb_order)
 
     def get_image_as_buffer(self, output=None):
         """Get the current image shown in the viewer, with any overlaid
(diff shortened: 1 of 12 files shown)


ejeschke:
@pllim, no hurry on this.

ejeschke:
@mwcraig, would you mind testing this with the `AstroWidget`? I'm especially interested if you could test it with the `cairo` package from conda installed. The rendering of text into the widget should be nicer than with the previous default renderer.

mwcraig:
@ejeschke -- sure, should be able to give it a try by the end of the week

ejeschke:
Rebased.

mwcraig:
Trying it out this morning...

mwcraig:
Took a quick look and everything appears to work as expected.

ejeschke:
Thanks, @mwcraig .

ginga/ImageView.py
```diff
@@ -3155,15 +3155,18 @@ def render_image(self, rgbobj, dst_x, dst_y):
...
         Parameters
         ----------
         time_sec : float
             Time, in seconds, to wait.
 
         """
-        self.logger.warning("Subclass should override this abstract method!")
+        # subclass implements this method to call delayed_redraw() after
+        # time_sec
+        self.delayed_redraw()
```
pllim: Do you really mean this, from the comment?
```suggestion
        self.delayed_redraw(time_sec)
```

ginga/aggw/ImageViewAgg.py
```diff
@@ -20,56 +20,11 @@ def __init__(self, logger=None, rgbmap=None, settings=None):
...
-                                                              format=format,
-                                                              quality=quality)
-
-    def get_rgb_image_as_bytes(self, format='png', quality=90):
-        # TO BE DEPRECATED: DO NOT USE
-        return self.renderer.get_surface_as_rgb_format_bytes(format=format,
-                                                             quality=quality)
-
-    def save_rgb_image_as_file(self, filepath, format='png', quality=90):
-        # TO BE DEPRECATED: DO NOT USE
```
pllim: Does this mean deprecation is no longer planned?

ginga/web/jupyterw/ImageViewJpw.py
```diff
@@ -46,36 +46,43 @@
...
-        ImageView.__init__(self, logger=logger,
-                           rgbmap=rgbmap,
-                           settings=settings)
+        ImageView.ImageViewBase.__init__(self, logger=logger,
+                                         rgbmap=rgbmap,
+                                         settings=settings)
 
+        self.t_.set_defaults(renderer='cairo')
+
+        self.rgb_order = 'RGBA'
```
pllim: You had a comment before that said `# Format 'png' is ok with 'RGBA', but 'jpeg' only works with 'RGB'`. Is that no longer the case?

ginga/ImageView.py
ejeschke: No, `time_sec` is set elsewhere.  I will update the comment.

ginga/aggw/ImageViewAgg.py
ejeschke: Yes, at least for now, and the comment is removed where the method migrated to.

ginga/web/jupyterw/ImageViewJpw.py
ejeschke: This is now handled correctly in the `ginga.pilw.CanvasRenderPil` module, and we no longer need the comment or the restriction.

ginga/ImageView.py
ejeschke: Changing the comment to:
```python
# subclass implements this method to call delayed_redraw() after
# time_sec.  If subclass does not override, redraw is immediate.
```

ejeschke:
Thanks for the review, @pllim.  I will merge if the build is successful.

pllim:
I tested it with the examples notebooks in this repo and have `pycairo` installed. Nothing broke. 👍

PR #791, opened 2019-08-13
Enhancements to text drawing

ejeschke:
- Enhances the basic `Text` canvas type (and other types that draw text) to have more fluidity when scaling fonts.  The font size can be fixed or vary according to the viewers scale, with the ability to set a base size at scale 1.0 and have it vary up or down from there.
- API is preserved.  If the `fontsize` was specified as None, then the font is scaled as before with a base size of 10.0 points at scale 1.0. If the `fontsize` is specified as a number, and `fontscale` is True, then that size is used as the base at 1.0 and scaled accordingly. If the `fontsize` is specified as a number and `fontscale` is False, then this indicates a fixed size irrespective of scale.
- `Text` objects on the canvas when selected in edit mode now show rotation and scale adjustment handles.  By moving the former you can rotate the text; by moving the latter you can scale the font.
- It is possible to set the minimum and maximum size of the font with `fontsize_min` and `fontsize_max` in a `Text` object.
- This PR also fixes a bug in the rotation of text with the agg back end.


ejeschke:
@pllim, when you have time.  I'll probably add a little more to this PR in the way of documentation, but it is ready for testing.  The biggest thing you might notice that some of the text sizes vary more with the viewer scale in the plugins that don't fix the size of the font that they draw on the canvas.

ejeschke:
Rebased

ejeschke:
Rebased

ejeschke:
> I can't see any difference. I guess that is a good thing.

Open the `Drawing` plugin, select `Text` as the type of thing to draw, and click in the window to place the text.  Now press the `Edit` button for the canvas and click on the lower left corner of the text to select it. You might have to try a few times, as the selection area for text is currently just a single point with a small radius (it helps to zoom in sometimes).

Once the text object is selected, you should see three control edit points.  The anchor moves the object, the rotation one rotates the text, and the scale one scales the font size.

pllim:
@ejeschke ,  thanks for the instructions. They do appear to work as advertised, so feel free to merge.

The "select lower left to edit" is a little non-intuitive and hard to find, but I guess it is out of scope for this PR. I was expecting the anchor points and dotted box to pop up when I click anywhere on the text.

ejeschke:
> I was expecting the anchor points and dotted box to pop up when I click anywhere on the text.

That would be my choice as well, and I hope to eventually get there, but it turns out that text is tricky compared to other forms.  In particular, finding the bounding box isn't as straightforward as it should be across widget sets and will require some more refactoring outside of this PR.

As a workaround for the present, I'm trying to make text work similarly to a `Point`, which lets you click basically anywhere within it's radius to select it.  The question is finding a good default radius for a `Text` object since it doesn't carry a radius attribute (and I don't want to add one because eventually you should just be able to click anywhere in the text).

ejeschke:
Rebased.

ejeschke:
@pllim, I was able to make it so that you can click on the text anywhere to select it for editing.  However, it still shows the single-radius style control points for scaling and rotating.  Might be able to fix that in a future PR.

pllim:
Sounds good. Thanks, @ejeschke!

PR #794, opened 2019-08-22
Protections against possible problems from numpy 1.17.0 clip()

ejeschke:
`numpy` changed the behavior of `clip()` in 1.17.0 where it returns a float when operating on an int array.  This PR provides workarounds for potential code sections that could be affected.

### DIFF
--- a/ginga/ColorDist.py
+++ b/ginga/ColorDist.py
@@ -35,7 +35,7 @@ class ColorDistBase(object):
     def hash_array(self, idx):
         # NOTE: data could be assumed to be in the range 0..hashsize-1
         # at this point but clip as a precaution
-        idx = idx.clip(0, self.hashsize - 1)
+        idx = idx.clip(0, self.hashsize - 1).astype(np.uint, copy=False)
         arr = self.hash[idx]
         return arr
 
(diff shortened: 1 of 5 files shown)


pllim:
For some reason conda `defaults` still doesn't have numpy 1.17, so we need to test this locally with installation from `conda-forge`.

ejeschke:
> Should we add a test to catch that edge case that caused the segfault?

I'm trying to think how that would work.  We could of course test that `clip` returns a float.  But that would fail if users are using an older version of `numpy`. And with this and the previous fix we already put in workarounds for the new behavior that are backward compatible.  I'm open to suggestions, but not sure what the test should look like.

pllim:
Re: test -- I am not sure either, but it is apparent that the tests did not catch the segfault. Anyway, it should not hold this up. Feel free to merge. Thanks for the quick fix!

ejeschke:
> it is apparent that the tests did not catch the segfault.

Yes, but I'm not sure we would have had a test for that.  It's like putting a test in for 
```python
type(int(x)) != float
```
You just don't expect that is necessary. Who would have expected that a `clip()` operation on an integer array (with `int` parameters) would produce a float!

PR #798, opened 2019-08-30
TST: Use RTD PR builder

### LINKED ISSUE #797: TST: Enable ReadTheDocs PR build
As per https://blog.readthedocs.com/building-docs-for-pull-requests/ , we can take doc build job off Travis CI and let ReadTheDocs (RTD) do it instead. The advantages are that the documentation would be built exactly as it would after merge (no surprises caused by infrastructure difference) and you get a HTML preview *before* merge.

I already have this beta feature enabled for this repository. However, I cannot access the admin settings here, so @ejeschke would have to go in and make sure the RTD webhook is listening to "pull requests".

p.s. You can see it in action at spacetelescope/synphot_refactor#214 .

pllim:
Fix #797


pllim:
@ejeschke , this is ready for review. You can see that there is an extra "read-the-docs" check right above "travis-ci" now. And if you click on "Details", you will see a *rendered* document if successful and traceback if failed. And the PR build log will appear as the same place as `master` etc at https://readthedocs.org/projects/ginga/builds/ .

pllim:
Merged. Thanks for the review!

PR #799, opened 2019-08-31
DOC: Fix ico mimetype warning

pllim:
Follow up of #798 . Not sure why we didn't see the RTD warning (which now causes failure for the PR check) back in #798. Maybe cache was not being cleared until after merge or something. 🤷‍♀ 

```
sphinx.errors.SphinxWarning: unknown mimetype for _static/astropy_logo.ico, ignoring

Warning, treated as error:
unknown mimetype for _static/astropy_logo.ico, ignoring
```


pllim:
Merging now so I can see if this really fixes the build on `master`. 🤞

PR #806, opened 2019-10-30
Refactor Zoom plugin

ejeschke:
This small refactoring changes the way the Zoom plugin operates. Instead of cutting a part of the data and updating a viewer, it shares the canvas for the viewer that it is currently being used with, and simply sets the pan position to the point of the cursor in the main viewer.

The advantages are:
- simplifies conceptual operation of the plugin
- makes the plugin more efficient
- allows the zoomed view to see items that are drawn on the canvas


ejeschke:
@pllim, any objections to changing `Zoom` in this fashion?

ginga/examples/configs/plugin_Zoom.cfg
```diff
@@ -3,21 +3,10 @@
...
-zoom_radius = 30
-
 # default zoom level
 zoom_amount = 3
 
 # refresh interval (sec)
 # NOTE: usually a small delay speeds things up
 refresh_interval = 0.02
 
-# rotate the zoom image if the channel image is rotated?
```
pllim: Hmm, so all the removed settings are basically features that are no longer supported?

ejeschke: Yes.  There is actually no reason that the new version of the plugin could not support those features.  Since they were turned off by default I don't know how many people were using them anyway.  If we get an issue that someone misses one of those features it would be easy enough to add it back in.  But the plugin is simpler without them.

ejeschke:
> I really like being able to see the drawing in Zoom too. Thanks!

That's a nice bonus, yes, and one of the reasons for the update.

Issue #328, opened 2016-04-28
How does Ginga handle image distortion?


pllim:
Is there a way to un-distort a distorted image (using its WCS) and display the un-distorted version on Ginga? This was also asked by @mcara.

pllim:
c/c @stscieisenhamer in case he secretly implemented this for 3D printing project...

mcara:
I think this would be a great feature. I do not think DS9 is capable of doing this either.

ejeschke:
I think if you are using `astropy` wcs there is *some* support for distortion correction, no?  And I think that even some images that have distortion information in auxiliary tables in the header can be supported.  But this probably just means that the coordinate reporting and graphing should be correct.  There is no actual warping of the image.  I wonder if that would be possible to do fairly straightforwardly using the astropy `reproject` module?

pllim:
Yes, `reproject` was briefly discussed. The solution could be to use that to create a "corrected" image in memory and display it, and then use `SaveImage` to save it out if desired. Or to modify `Mosaic` to work with `reproject`. Of course, solution like this is prone to memory being all used up, so it is not perfect.

ejeschke:
@pllim, revisiting this issue.  Do you have some examples of FITS files with significant distortion and screenshots of corrected images?  Would like to try out some support for correction in a branch I am working on.

pllim:
How about ACS on HST? http://www.stsci.edu/hst/acs/analysis/distortion

You can request all its images at http://archive.stsci.edu/hst/search.php . The Wide Field Camera in ACS is what you want.

pllim:
@ejeschke , are you still working on the support or should I just close this?

pllim:
This came up again... cc @eteq , @hcferguson , and @larrybradley

ejeschke:
> Yes, reproject was briefly discussed. The solution could be to use that to create a "corrected" image in memory and display it ...

I think this would be rather trivial to implement from a working example of morphing one `ndarray` to a corrected one.  If the resulting image is not rectangular, has holes, etc. this could be handled by either `NaN`s, numpy masks or an alpha mask.  The other thing that would probably need to be done is to change the WCS so that overplotting and coordinate readout under the cursor would work correctly--once you correct the image the "correction" has to be taken out of the WCS.

ejeschke:
See #813 for a strawman approach

ejeschke:
[...]

PR #813, opened 2019-12-19
Add `Reproject` plugin for reference viewer

ejeschke:
- This is a plugin that basically provides a GUI for doing WCS reprojections using the AstroPy "reproject" module.
- Provided mostly as a basis for further discussion on #328


ejeschke:
This is mostly just to advance the discussion in #328.  I think probably the support needs to be baked in a little deeper than just in a plugin, but this is a strawman to get ideas flowing.

To test this you need to have the AstroPy `reproject` module installed.

To use: start the `Reproject` plugin from the Operations->Utils menu (or Plugins->Utils).  Load an image into the channel that you want to use as the reprojecting WCS (e.g. by drag drop).  Click "Set WCS".  Now load any image containing data that you want to reproject and click "Reproject".  The image will be reprojected and added to the channel.  You can manipulate the reprojection parameters using the GUI.

Caveats:
* currently, the reprojected image does not contain a WCS.
* the reprojected image may be clipped.  I cannot see a way around this using the current `reproject` module (even though it should be quite possible)
* the reprojected image is currently shown with `NaN`s replacing unused pixels
* I have not updated any documentation yet in this PR

ejeschke:
ping @pllim, @eteq, @hcferguson, @larrybradley

ejeschke:
Added a WCS to the reprojected image.

ejeschke:
Screenshot, using 2MASS example in `reproject` documentation:
Screenshot from 2019-12-19 09-45-14

pllim:
@ejeschke , two things:

1. I have one big image uncorrected for distortion, I want to un-distort it for display but only the displayed portion of it. This plugin does not seem to handle that use case? If it does, I failed to figure it out from the docstring.
2. Having the entire big image (2k by 4k) displayed and clicking the reproject button, the viewer got stuck while the reprojection algorithm churns away. If we want to go this route with using plugin, we probably need some sort of progress bar display like the `Mosaic` plugin. Or simply refuse to process a too-big array.

I also encountered segfault at some point but I don't know how to reproduce it.

ejeschke:
Rebased.

ejeschke:
> I have one big image uncorrected for distortion, I want to un-distort it for display but only the displayed portion of it. This plugin does not seem to handle that use case? If it does, I failed to figure it out from the docstring.

Well, that's what I mean when I said (above):
> probably the support needs to be baked in a little deeper than just in a plugin, but this is a strawman to get ideas flowing

Ideally, the distortion should be corrected at image load time.  That way, you only pay for it once, especially if it is compute intensive.  

But, no, this plugin does not do just the displayed portion. We might be able to refactor it to work on only the displayed area, I would think.

ejeschke:
> Having the entire big image (2k by 4k) displayed and clicking the reproject button, the viewer got stuck while the reprojection algorithm churns away. If we want to go this route with using plugin, we probably need some sort of progress bar display like the Mosaic plugin. Or simply refuse to process a too-big array.

I thought that too, but I'm not sure offhand how to monitor the `reproject` call so that it would report its progress.

ejeschke:
We could run the reproject call in another thread--that would be easy.  The viewer would still be responsive.  I'll see if I can push up a fix for that.

ejeschke:
> We could run the reproject call in another thread--that would be easy. The viewer would still be responsive. I'll see if I can push up a fix for that.

@pllim, if you try the plugin now you will see that it remains responsive during the reprojection. I also added a little status message in the plugin.

ejeschke:
@pllim, does drizzle do what you want for STScI images? Has any put together a drizzle plugin for ginga?

ejeschke:
Also, can I get a copy of your large image and also your "corrected one", to test with this?  You could put them in the Box drive, if you folks are still using that...

hcferguson:
@pllim @ejeschke There was an effort started some time ago to put drizzle into reproject. That's going to get some attention again over the next few months by Mihai Cara and others.

pllim:
TBH, I still think it is a bad idea to have a GUI do the drizzling/reprojection on-the-fly for the whole image. It is computationally intensive and the image can be pre-processed outside the GUI anyway. I just don't see the point.

I think @hcferguson said one thing that can be useful is to use this as a way to explore which are the optimal parameters to use for the algorithm, but then the parameters that are exposed depend on the use case. So, I don't see much value in that either but maybe Harry can clarify if I misunderstood.

hcferguson:
The main use case is for viewing multiple images with the same pixel scale and rotation in different windows. Generally one can ignore higher-order distortions for doing that. And for speed, one could do these all on just the portion viewed. Also one could in principle do this only for the bit depth viewed (i.e. on 16 or 32 bit integer data rather than floats). And there are in principle methods to use GPUs for such transformations, although I don't know how portable they are. 

I suppose one relatively common use-case that stresses doing the whole image would be to overlay a mask (e.g. cosmic rays) from an unrectified image onto the rectified image. There it's necessary to do the full reprojection, and one generally wants a method that does some nearest-neighbor interpolation so the reprojected mask is still binary.

pllim:
Thanks for the clarification, @hcferguson . Can you please upload some test images (maybe 3 would be sufficient?) to https://stsci.app.box.com/folder/3381606676 for @ejeschke ?

ejeschke:
Rebased.

ejeschke:
> the reprojected image may be clipped. I cannot see a way around this using the current reproject module (even though it should be quite possible)

I fthink I've figured this out.  With the latest commit and example images it now can get the fully rotated result.

ejeschke:
> The main use case is for viewing multiple images with the same pixel scale and rotation in different windows. Generally one can ignore higher-order distortions for doing that. And for speed, one could do these all on just the portion viewed.

Isn't that covered pretty well with the `WCSMatch` plugin?

Screenshot from 2020-01-09 17-42-10

ejeschke:
There is another candidate PR (#816) that I think works a little more closely to the workflow @pllim was describing.  You don't need two WCSes and it doesn't use `reproject`, instead it tries to remap the pixels using the SIP WCS in the image.

I'm still a little unclear if this is the kind of "undistortion" being described in this issue, but it seems to do something akin to barrel distortion type adjustment to a couple of the HST images I tried.

hcferguson:
> > The main use case is for viewing multiple images with the same pixel scale and rotation in different windows. Generally one can ignore higher-order distortions for doing that. And for speed, one could do these all on just the portion viewed.
> 
> Isn't that covered pretty well with the `WCSMatch` plugin?
> 

Yes.

hcferguson:
> There is another candidate PR (#816) that I think works a little more closely to the workflow @pllim was describing. You don't need two WCSes and it doesn't use `reproject`, instead it tries to remap the pixels using the SIP WCS in the image.
> 
> I'm still a little unclear if this is the kind of "undistortion" being described in this issue, but it seems to do something akin to barrel distortion type adjustment to a couple of the HST images I tried.

Right now the JWST images don't have SIP coefficients in their FITS headers. Perry Greenfield has been investigating that as an option, since ds9 supports SIP re-mapping (somehow). But I don't think there is yet a definite plan to put SIP in the JWST headers.  At some level it seems like a step backwards when we have all the sophisticated gWCS machinery. But on the other hand, for visualization, speed often trumps accuracy.

ejeschke:
@hcferguson, how do you enable that in ds9? When I load HST images, they look the same as they do in ginga.

hcferguson:
@ejeschke I'm not sure myself. There's some discussion back on https://github.com/astropy/astropy/issues/4761. I'm not sure if it's (a) ignoring SIP, (b) only using it to convert pixel positions to RA & Dec, or (c) doing a reprojection. I would suspect (b).

ejeschke:
I've made one more modification to this plugin. It now has *both* Reproject and Undistort buttons. The Undistort button creates a derivative WCS from the current image without the SIP components, and simply reprojects the current image to that (using `reproject`). Not having inspected the internals of `reproject`, I'm sure if this is a viable approach, but you can compare the result to externally corrected images to compare and see how well it works.

I think if this comes close to doing what you want it is preferable to #816, since it can also be used as a general reprojection GUI for two separate images, exposing the features of the `reproject` module (and thus should eventually be able to do drizzling, etc. if plans work out as @hcferguson describes above).

pllim:
@ejeschke , thanks! I tried the updated plugin with HST/ACS FLT image (calibrated but not drizzled) but the "undistort" button remains grayed out. Under what condition is the button usable?

I don't think I have much to test this on until @hcferguson can provide clear use cases and accompanying data with his requests.

ejeschke:
> Under what condition is the button usable?

At the moment, only if the image contains a WCS with SIP distortion information.  I tested with HST images, so I know that some of them contain this info.  Basically, when creating an `astropy.wcs.WCS` object from the header (e.g. `obj`), the value for `obj.sip` should not be `None`.

Is it possible that the calibrated image has had the keywords stripped?

pllim:
> calibrated image has had the keywords stripped?

I am not sure. I think we should just wait for Harry to provide some realistic data to play with.

pllim:
> the reprojected image may be clipped. I cannot see a way around this using the current reproject module (even though it should be quite possible)

@ejeschke , here's a reply from @astrofrog , "reproject will reproject images to an exact header so if it's clipped it's because the final header is too small, so making the NAXIS values larger and recentering it with CRPIX should work"

ejeschke:
I figured that out. The current state of the PR does not clip (AFAIK).

hcferguson:
> Thanks for the clarification, @hcferguson . Can you please upload some test images (maybe 3 would be sufficient?) to https://stsci.app.box.com/folder/3381606676 for @ejeschke ?

Sorry for the delay. I've uploaded 3 images...from ACS. 
 - jbt7a3020_drz.fits (drizzled, rectified image)
 - jbt7a3k7q_flc.fits (undrizzled, distorted image)
 - jbt7a3kaq_flc.fits (undrizzled, distorted image)

If I load these into ds9 and Frame->lock->WCS and zoom in to center on a particular star (say, the one at 11:15:58.73, 1:30:51.12), what I see is that this star is centered in all three images, but the ones around it are not lined up. If I cursor over to the stars that are not centered and just center the cursor on those individually, it does read back the same RA & Dec, even though they are not pixel aligned.

So basically ds9 is not doing a reprojection (as I expected), but it is using SIP properly compute the center of the displayed image for each of the locked frames, and it is using it to do the conversions from x,y to RA, Dec.

ejeschke:
> I've uploaded 3 images...from ACS.

Thanks, @hcferguson, but I don't see the images in that Box drive. I see a lot of other files, but not those.
 
If you can make them visible we can test with the plugin.

hcferguson:
Sorry about that @ejeschke. They are there now.

pllim:
p.s. This needs a rebase to fix the `ginga` command line.

pllim:
>  (say, the one at 11:15:58.73, 1:30:51.12)

For my own future reference, this is the one at `X=748 Y=479` in the DRZ image.

pllim:
Using the three files that @hcferguson provded in https://github.com/ejeschke/ginga/pull/813#issuecomment-584735515 ...

1. WCS match after load without any other manipulations. DRZ is the reference frame. DRZ is on upper left.

Screenshot from 2020-02-17 17-06-55

2. I use DRZ to set WCS, then reprojected one of the FLCs with the "Reproject" button. Then I move the reprojected result to another channel so I can use `WCSMatch`. DRZ and reproject result do not align at all. DRZ is on the left.

Screenshot from 2020-02-17 17-16-26

3. Still using the `Reproject` plugin in Step 2 above (probably still has DRZ as reference WCS from clicking that "Set WCS" above), I now press the "Undistort" button for one of the FLC. Then again, I move the undistorted FLC to another channel for `WCSMatch` to work. Much better but if you look carefully, the alignments are not perfect (i.e., things shift a little when you blink them). DRZ is on the left.

Screenshot from 2020-02-17 17-21-08

ejeschke:
Rebased.

ejeschke:
Rebased

PR #814, opened 2019-12-26
Remove astropy-helpers

pllim:
Fix #758 

Also tagged 3.1.dev as part of this work (see tags in this repo), as `setuptools_scm` requires dev tag after a release for proper version tracking.

PR #815, opened 2020-01-04
Unify the coordinate API for canvas types

ejeschke:
This adds alternative classes for many canvas types so that *all* types use an API where points are specified as single parameters.  This will be beneficial if we ever want to make a common change to coordinate handling across all canvas objects.

To illustrate, before:
```python
>>> canvas.add(Rectangle(x1, y1, x2, y2, ...))
```
after:
```python
>>> canvas.add(RectangleP(p1, p2, ...))
```
Subclasses are used to keep API compatible versions of the old class constructors available, and the new classes are not exposed as drawing catalog members yet (although you can explictly import them and use them. So there is no API change visible to any older code.

### DIFF
--- a/ginga/canvas/types/astro.py
+++ b/ginga/canvas/types/astro.py
@@ -27,7 +27,7 @@ __all__ = ['Ruler', 'Compass', 'Crosshair', 'AnnulusMixin', 'Annulus',
            'Annulus2R', 'WCSAxes']
 
 
-class Ruler(TwoPointMixin, CanvasObjectBase):
+class RulerP(TwoPointMixin, CanvasObjectBase):
     """
     Draws a WCS ruler (like a right triangle) on a DrawingCanvas.
     Parameters are:
@@ -86,14 +86,14 @@ class Ruler(TwoPointMixin, CanvasObjectBase):
 
     @classmethod
     def idraw(cls, canvas, cxt):
-        return cls(cxt.start_x, cxt.start_y, cxt.x, cxt.y, **cxt.drawparams)
+        return cls((cxt.start_x, cxt.start_y), (cxt.x, cxt.y), **cxt.drawparams)
 
-    def __init__(self, x1, y1, x2, y2, color='green', color2='skyblue',
+    def __init__(self, pt1, pt2, color='green', color2='skyblue',
                  alpha=1.0, linewidth=1, linestyle='solid',
                  showcap=True, showplumb=True, showends=False, units='arcmin',
                  font='Sans Serif', fontsize=None, **kwdargs):
         self.kind = 'ruler'
-        points = np.asarray([(x1, y1), (x2, y2)], dtype=np.float)
+        points = np.asarray([pt1, pt2], dtype=np.float)
         CanvasObjectBase.__init__(self, color=color, color2=color2,
                                   alpha=alpha, units=units,
                                   showplumb=showplumb, showends=showends,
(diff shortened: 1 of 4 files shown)


ejeschke:
@pllim, when you have time.  No hurry on this.

pllim:
#814 will fix the failure -- FYI

ginga/canvas/types/astro.py
```diff
@@ -656,19 +709,19 @@ def __init__(self, x, y, radius, width=None,
         coord = kwdargs.get('coord', None)
 
         klass = get_canvas_type(atype)
-        obj1 = klass(x, y, radius, color=color,
+        obj1 = klass(pt[0], pt[1], radius, color=color,
```
pllim: Why not use the new class that takes point here?

ginga/canvas/types/astro.py
```diff
@@ -800,13 +871,14 @@ def get_params_metadata(cls):
...
+        return cls((cxt.start_x, cxt.start_y), (xradius, yradius),
+                   **cxt.drawparams)
 
-    def __init__(self, x, y, xradius, yradius, xwidth=None,
+    def __init__(self, pt, radii, xwidth=None,
                  ywidth=None, atype='ellipse', color='yellow',
                  linewidth=1, linestyle='solid', alpha=1.0,
                  rot_deg=0.0, **kwdargs):
-
+        xradius, yradius = radii
```
pllim: Is this used anywhere?

ginga/canvas/types/astro.py
```diff
@@ -247,7 +247,25 @@ def draw(self, viewer):
...
+
+    @classmethod
+    def idraw(cls, canvas, cxt):
+        return cls(cxt.start_x, cxt.start_y, cxt.x, cxt.y, **cxt.drawparams)
+
+    def __init__(self, x1, y1, x2, y2, color='green', color2='skyblue',
+                 alpha=1.0, linewidth=1, linestyle='solid',
+                 showcap=True, showplumb=True, showends=False, units='arcmin',
+                 font='Sans Serif', fontsize=None, **kwdargs):
+        RulerP.__init__(self, (x1, y1), (x2, y2), color=color, color2=color2,
```
pllim: Why not use `super`?

ginga/canvas/types/astro.py
ejeschke: I think because we are looking up the class based on the string in
```python
klass = get_canvas_type(atype)
```
and these classes are not yet added to the drawing dictionaries.  I wanted to put in the support but not actually modify the drawing (class) dictionaries yet.

ginga/canvas/types/astro.py
ejeschke: Yes, just below:
```python
xwidth = 0.15 * xradius
```

ginga/canvas/types/astro.py
ejeschke: It might work, but the parent class is derived from mixin classes and uses explicit class calls for it's constructor.  So I'm just trying to be consistent here.

ejeschke:
> I like the concept and I appreciate you keeping this backward compatible! However, the definition of points should be clarified in documentation, so it is clear that pt is (x, y) and not (y, x).

Good point (pun intended!). We should mention this somewhere.  But currently all points for polygons, paths, bezier curves, etc all are ordered (x, y, [z]).

> I think get_params_metadata also needs to be updated?

Perhaps.  Fundamentally, all coordinates (already) are represented at a higher level by points (this was from an earlier refactor a year or two ago).  But things like rectangle or line gained and exposed properties (e.g. x1, y1, x2, y2) that work to set the components of the points.  As the `get_params_metadata` is mostly used to construct GUIs for the properties of the objects, it might make sense to leave them as is so that things like `Drawing` continue to work without any surprises.

PR #818, opened 2020-01-16
BUG: Fix CLI entry points

pllim:
Oversight in #814 . My bad! This adds the `ginga` and `ggrc` CLI commands back.


pllim:
The RTD time out might be a fluke, given that it passed in the first commit, but if you are worried about it, I can do a force push to trigger a rerun. Please let me know, @ejeschke !

PR #822, opened 2020-01-31
MNT: Move coveragerc to setup.cfg

pllim:
Follow-up of #814


setup.cfg
```diff
@@ -81,3 +80,43 @@ console_scripts =
...
+    ginga/conftest*
+    ginga/cython_version*
+    ginga/setup_package*
+    ginga/*/setup_package*
+    ginga/*/*/setup_package*
+    ginga/tests/*
+    ginga/*/tests/*
+    ginga/*/*/tests/*
+    ginga/version*
+    */ginga/_astropy_init*
```
pllim: Note: Tom R said this repetition is for `tox`. So might as well, in case you want to use `tox` in the future.

PR #840, opened 2020-05-08
MNT: Ignore import error in pytest collection

pllim:
This is an automated update made by the `batchpr` tool :robot: to future-proof test collection behavior `pytest-doctestplus` - feel free to close if it doesn't look good or isn't relevant! You can report issues to @pllim.

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -3,7 +3,7 @@ minversion = 3
 testpaths = "ginga" "doc"
 norecursedirs = build doc/_build
 astropy_header = true
-addopts = -p no:warnings
+addopts = -p no:warnings --doctest-ignore-import-errors
 
 [bdist_wheel]
 universal = 1


ejeschke:
@pllim, deferring to you on this one...

pllim:
Well, I am obviously biased since I submitted the PR. Do you use `doctest`? Do you use or plan to use `pytest-doctestplus`? If yes to either or both, you probably want this. If not, then you don't need it.

ejeschke:
I have not used `doctest`, but if that is something that runs tests embedded in documentation, it sounds useful.

pllim:
In that case, should we close this without merge? Seems like it won't be used for now anyway?

ejeschke:
Unless you think it would cause problems, I think we should merge it.  I could see writing some doctests in the future. I haven't before, mostly because I wasn't really aware of the tooling.

pllim:
Okay, merging then. You can always take it back out if you don't like it. I cannot find any good documentation for this option. But you can see https://docs.pytest.org/en/latest/_modules/_pytest/doctest.html .

`doctest` is good for non-GUI code examples. I don't think it is meant for GUI testing. FYI.

PR #845, opened 2020-05-28
Ginga pipeline

ejeschke:
This is a new attempt at a generalized `Pipeline` plugin for Ginga.  A `Pipeline` class allows pipelines to be constructed of `Stage`s. A `Pipeline` plugin is provided which offers a GUI for building and controlling a general image processing pipeline.  A beginning set of stages are provided.  I expect some interesting stages could be developed for quick look type applications like flat fielding, bias/dark/sky subtraction, binning, etc. based on `astropy` modules.

In addition to the full-image data pipelines, the standard pixel renderer is refactored to use a pipeline for generating the output to standard backends. This makes the rendering process much easier to understand and provides future opportunities for interesting modular hacks by inserting stages in the pipeline instead of having to carefully modify an algorithm embedded in the renderer itself.


pllim:
Interesting idea! Hard for me to look through 3k lines of changes, so I'll just ask: Does this use [cddproc](https://github.com/astropy/ccdproc/) to process the image? Does user have an option to plug in their own pipelines into this GUI?

ejeschke:
> Does this use cddproc to process the image? 

Currently there are no CCD-like operation stages, but I certainly imagine we would use ccdproc or some other astropy ecosystem packages to write some.

> Does user have an option to plug in their own pipelines into this GUI?

The interesting thing about this approach is that if you write to the `Stage` API, your stage can be inserted and moved up and down the pipeline GUI like any other stage.  And if you have experience writing Ginga a plugin, a stage is *very* similar, right down to the API (there are some minor differences).

pllim:
Todos from demo meeting (thanks for the demo!):

* Tutorial-like doc to demonstrate to users how this works.
* When PR is ready, ping Matt Craig and STScI TEL for possible reviews.

Link of interest: https://mwcraig.github.io/ccd-as-book/00-00-Preface.html

ejeschke:
@pllim, as mentioned in #867, I'd like to merge this PR before the next release.  As per your comments above, this needs some documentation and a good example use case for an astronomy pipeline.

But, since an important secondary purpose is that it is used as the main image display pipeline, I'd like to get that part in now so that future internal mods can make use of the easier to understand pipeline architecture.  I've commented out the `Pipeline` entry from the plugins so that the plugin doesn't show up for regular users.  I'll target an example and better docs, with the plugin enabled in the following release.

Any objection to merging?

pllim:
Sounds good. Feel free to merge. Thanks!

PR #847, opened 2020-07-07
Fix for problem with PixTable plugin when used with Zoom

### LINKED ISSUE #836: PixValue not reporting correct value when used with Zoom
First reported in https://github.com/ejeschke/ginga/pull/812#issuecomment-618754188 but then I realized it is already that way on `master` even without #812 patch. Release 3.0.1 worked fine. I tried rolling back the OpenGL PR #830 but that didn't help either.

UPDATE: Caused by  #806 . With current `master`, if you set the zoom amount in `Zoom` to, say `1/18x`, `PixTable` reports correct values!

ejeschke:
Fix for github issue #836.  Fixes a issue where if PixTable and Zoom are both launched on the same channel PixTable will report two sets of values in succession. The second set will in most cases not be correct.

This PR fixes this particular issue with PixTable, but it may be possible for a future plugin to suffer the same problem, so a more permanent fix should be implemented, preferably by refactoring how the `cursor-changed` callback is implemented.

EDIT: Fix #836

### DIFF
--- a/ginga/rv/plugins/PixTable.py
+++ b/ginga/rv/plugins/PixTable.py
@@ -119,7 +119,7 @@ class PixTable(GingaPlugin.LocalPlugin):
 
         self.dc = self.fv.get_draw_classes()
         canvas = self.dc.DrawingCanvas()
-        canvas.set_callback('cursor-changed', self.cursor_cb)
+        self.fitsimage.set_callback('cursor-changed', self.cursor_cb)
         canvas.enable_draw(True)
         canvas.set_drawtype('point', color='cyan', linestyle='dash')
         canvas.set_callback('draw-event', self.draw_cb)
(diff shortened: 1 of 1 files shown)


ejeschke:
@pllim, thanks for the pointers, found the issue.  Not related to ssh at all, but only whether `PixTable` and `Zoom` are active at the same time.  Please test with this PR and LMK if LGTM. :smile_cat:

pllim:
Oh! I must not have Zoom active when I tested locally on the PC then. Sorry for throwing you off the loop there for a while. I'll test this and let you know. Thank you!

pllim:
`PixTable` works now. Thanks!

And looks like this is the only plugin that used `canvas.set_callback` as far as I can grep. Perhaps add a developer doc that warns about using `canvas.set_callback` for the future?

pllim:
(Just a minor comment about whether we need to document the gotcha. Otherwise, feel free to merge!)

ejeschke:
> Perhaps add a developer doc that warns about using canvas.set_callback for the future?

It's not `set_callback` or even using callbacks in general. The problem is only with the `cursor-changed` callback--needs to be redesigned. But currently the way this callback is being used by the current set of plugins is ok. Are there any stginga only plugins that use this callback?

pllim:
> Are there any stginga only plugins that use this callback?

I don't think so.

PR #850, opened 2020-07-07
MultiDim should stack vertically

### LINKED ISSUE #849: MultiDim layout is wrong
multidim_horizontal

The second box should have been stacked horizontally below the first box.

pllim:
Fix #849

### DIFF
--- a/ginga/gw/Widgets.py
+++ b/ginga/gw/Widgets.py
@@ -16,20 +16,20 @@ elif tkname == 'pg':
 
 # MODULE FUNCTIONS
 
-def get_orientation(container):
+def get_orientation(container, factor=1.0):
     if not hasattr(container, 'size'):
         return 'vertical'
     (wd, ht) = container.size
     # wd, ht = container.get_size()
     # print('container size is %dx%d' % (wd, ht))
-    if wd < ht:
+    if wd < (ht * factor):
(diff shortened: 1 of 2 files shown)


ejeschke:
@pllim, I feel like this should be rather addressed as a plugin setting (preference).  Some keyword where default is like `orientation='vertical'`.  Then code in `build_gui` looks something like:
```
orientation = self.settings.get('orientation', 'vertical')
if orientation is None:
    # orientation should follow window shape
    vbox, sw, orientation = Widgets.get_oriented_box(container, scrolled=True)
else:
    vbox = Widgets.Box(orientation=orientation)
    sw = Widgets.ScrollArea()
    sw.set_widget(vbox)
self.orientation = orientation
...
```

ejeschke:
You will find that there are a number of plugins that auto-orient like this.  Another possibility is to just change the `get_oriented_box()` to need a "more" horizontal shape before returning a horizontally-oriented box.  This is fairly easy to do by altering the check of the aspect ratio in that method.

pllim:
While I like the preferences, I feel like it is not the right hammer for this nail. I have made the aspect ratio check more forgiving for `MultiDim`. Is this okay?

pllim:
Superseded by #851

PR #851, opened 2020-07-07
Make get_oriented_box() return more sane results

### LINKED ISSUE #849: MultiDim layout is wrong

ejeschke:
Fixes the `get_oriented_box()` function to take an `aspect` parameter which can be used to force it to make more sane decisions about vertical or horizontal orientation.

EDIT: Fix #849

### DIFF
--- a/ginga/gw/Widgets.py
+++ b/ginga/gw/Widgets.py
@@ -16,20 +16,24 @@ elif tkname == 'pg':
 
 # MODULE FUNCTIONS
 
-def get_orientation(container):
+def get_orientation(container, aspect=1.0):
     if not hasattr(container, 'size'):
         return 'vertical'
     (wd, ht) = container.size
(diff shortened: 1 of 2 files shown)


ejeschke:
@pllim, an alternative to #850 that addresses all plugins that use this feature. An advantage is that we can tune the auto-orientation on a plugin-by-plugin basis by simply changing the `aspect` parameter to `get_oriented_box()`. I set the default to 2.0, meaning that a window needs to be *over* twice as wide as it is tall to get the horizontal layout.  And for `MultiDim`, I set it to 3.0.  

Can you try this out and let me know what you think?  Remember, you can always close `MultiDim`, resize your windows/panels and then open it again to see the effect--no need to restart the reference viewer over and over.

PR #852, opened 2020-07-07
Remove cursor-changed callback from canvases (fix #848)

### LINKED ISSUE #848: cursor-changed callback needs refactor
As mentioned in #847

ejeschke:
- removes the cursor-changed callback from canvases. It only works correctly as coded when invoked from a viewer.

EDIT: Fix #848


ejeschke:
@pllim, this should remove the issue for future plugins, because you can only register for this callback now from a viewer (where it works ok).

ejeschke:
I will add some comments to `WhatsNew` before we merge...

pllim:
Actually, I'll just merge now, so I can test it as part of the "release candidate".

PR #853, opened 2020-07-08
BUG: Fix moving annulus

pllim:
One of the refactoring broke moving of an existing annulus. This should fix it.


ejeschke:
Hmm, it would be nice if derived types of this sort didn't have to manage moving all their own objects.  Let me take a quick look...

ejeschke:
@pllim, fixed in commit f33e409b77f251a7ef90e1bdf81fa361f7fd0855.  The `super` call had not been updated correctly in the refactor.

ejeschke:
Thanks for catching this!

ejeschke:
Sorry, hang on, let me merge the commit...

pllim:
Thanks, @ejeschke , but that commit is not on `master`. Where is it?

PR #854, opened 2020-07-08
Fixed a bug where Annulus objects could not be moved

ejeschke:
Alternative to #853.  Fixes the `super` calls to use the right subclass.

EDIT: Close #853


ejeschke:
@pllim, I think you will find this fixes the problem with moving the annulus objects.

pllim:
Not critical but I noticed that you made this change on your fork's master branch.

PR #856, opened 2020-07-08
Fix WCSAxes (fix #855)

### LINKED ISSUE #855: BUG: WCSAxes plugin is broken
Using the latest `master`, it no longer works. Maybe related to the recent refactoring associated with shapes/canvas. I used HST/ACS calibrated fullframe image but I don't think the image matters as long as it has a valid WCS.

broken_wcsaxes

To reproduce:

1. Open image with WCS in Ginga.
2. Start WCSAxes plugin.

ejeschke:
- remove print statement
- change to correctly get set of images plotted

EDIT: Fix #855

### DIFF
--- a/ginga/canvas/types/astro.py
+++ b/ginga/canvas/types/astro.py
@@ -1253,13 +1253,10 @@ class WCSAxes(CompoundObject):
         # see if we need to recalculate our grid
         canvas = viewer.get_canvas()
         vip = viewer.get_vip()
-        images = set(vip.get_images([], canvas))
+        cvs_imgs = vip.get_images([], canvas)
+        images = set([cvs_img.get_image() for cvs_img in cvs_imgs])
         diff = images.difference(self._cur_images)
-        print('diff', diff)
-        update = False
-        if len(diff) > 0:
-            # new image loaded
-            update = True
+        update = len(diff) > 0
 
         cur_swap = viewer.get_transforms()[2]
         if cur_swap != self._cur_swap:
@@ -1282,7 +1279,6 @@ class WCSAxes(CompoundObject):
             # only expensive recalculation of grid if needed
             self.ra_angle = None
             self.dec_angle = None
-            image = viewer.get_image()
             self.objects = self._calc_axes(viewer, images, cur_rot, cur_swap)
 
         super(WCSAxes, self).draw(viewer)
--- a/ginga/canvas/types/astro.py
+++ b/ginga/canvas/types/astro.py
@@ -1080,7 +1080,7 @@ class WCSAxes(CompoundObject):
         # for keeping track of changes to image and orientation
         self._cur_rot = None
         self._cur_swap = None
-        self._cur_limits = None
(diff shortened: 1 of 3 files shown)


ejeschke:
@pllim, this fixes some of the issues at least.  Now the WCSAxes updates correctly when I switch thumbs, rotate, or swap axes, so long as the WCS is working.

ejeschke:
@pllim, I get some issues with testing with HST images that have to do with the WCS, and not `WCSAxes`:
- 2020-07-08 12:15:25,570 | E | wcs_astropy.py:48 (load_header) | Error making WCS object: Cannot cast array data from dtype('>f8') to dtype('float32') according to the rule 'safe'
- fter 20 iterations, the solution is diverging at least for one input point.
2020-07-08 12:19:17,251 | W | astro.py:1159 (_get_path) | WCSAxes failed: 'WCS.all_world2pix' failed to converge to the requested accuracy.
After 20 iterations, the solution is diverging at least for one input point.

ejeschke:
> Now the WCSAxes updates correctly when I switch thumbs

I have to retract this part.  It seems that the limits are set after the draw(), so that it doesn't pick up the limits correctly.  I'll push another commit.

ejeschke:
Ok, now responding correctly to limits change.  Works nicely with `Collage`.

pllim:
@ejeschke , thanks, I'll test with this PR.

And just for the record, I was using dev version of astropy (4.2.dev). What version of astropy were you using?

pllim:
This fix does work but I needed to patch it a bit more with naojsoft/ginga#7 . Thanks!

ginga/canvas/types/astro.py
```diff
@@ -1274,6 +1275,11 @@ def draw(self, viewer):
             # and update all the text objects in self.objects
             update = True
 
+        cur_limits = viewer.get_limits()
+        if np.any(np.isclose(cur_limits, self._cur_limits)):
```
pllim, line 1279: I thought it's like this?
```suggestion
        if not np.isclose(cur_limits, self._cur_limits):
```

ejeschke: @pllim, doesn't work, because `np.isclose` returns an array, so you have to wrap with the `np.any`. I think it works nicely and is understandable.

pllim, line 1279: In that case, this is more correct?
```suggestion
        if not np.all(np.isclose(cur_limits, self._cur_limits)):
```

ejeschke: Right you are!

ejeschke:
@pllim, does this fix `WCSAxes` for you?

ejeschke:
What do you think about the Travis error:
```python
ImportError: erfa version 1.7 or later must be installed to use Astropy. erfa is not installed.
```
?

pllim:
Yeah, you can ignore that. Hopefully astropy/ci-helpers#451 would fix it.

PR #857, opened 2020-07-09
Downgrade rgb_cms warnings to info

pllim:
I keep seeing this warning when I start Ginga but I don't think I need to do anything about it, so I think info would have been more appropriate.
```
2020-07-09 13:50:47,065 | W | rgb_cms.py:63 (profile_to_working_pil) | No working profile defined; leaving image unprofiled.
```


pllim:
Test failure is unrelated (see astropy/astropy#10329), so merging. Thanks for the approval!

Issue #837, opened 2020-04-24
Help button segfaults


pllim:
I think this is a side effect of #830 even though I started Ginga *without* `--opengl` option. I see this on Linux:

1. Start Ginga: `ginga --loglevel=30 --stderr`
2. Start any plugin
3. Click "Help"

Ginga crashed with segfault:
```
QOpenGLShader::compile(Vertex): failed
QOpenGLShader::compile(Fragment): failed
Could not link shader program:
 ""
Segmentation fault
[14890:14890:0100/000000.705366:ERROR:broker_posix.cc(41)] Invalid node channel message
```

ejeschke:
@pllim, I think I may have fixed this issue in commit 882653935bac2395ca7131e16fa3a074abc36aeb. Could you try to recreate the error using master on the platform that you got this error?

pllim:
Unfortunately, it still segfaults with latest `master` with that commit you mentioned:
```
2020-05-04 15:15:03,925 | W | rgb_cms.py:64 (profile_to_working_pil) | No working profile defined; leaving image unprofiled.
QOpenGLShader::compile(Vertex): failed
QOpenGLShader::compile(Fragment): failed
Could not link shader program:
 ""
Segmentation fault
[28479:28479:0100/000000.119132:ERROR:broker_posix.cc(41)] Invalid node channel message
```

I did a clean install.

ejeschke:
@pllim, I will need your help because I cannot reproduce this issue on any machine so far.  Could you comment out the following lines in `ginga.qtw.ImageViewQt.py`:
```python
## try:
     from ginga.opengl.CanvasRenderGL import CanvasRenderer as OpenGLRenderer
     from ginga.opengl.GlHelp import get_transforms
     from ginga.opengl.glsl import req

     have_opengl = True
 except ImportError:
     pass
```
so that it looks like:
```python
## try:
##     from ginga.opengl.CanvasRenderGL import CanvasRenderer as OpenGLRenderer
##     from ginga.opengl.GlHelp import get_transforms
##     from ginga.opengl.glsl import req

##     have_opengl = True
## except ImportError:
##     pass
```
Ginga only tries to build an opengl program if the `--opengl` switch is passed, so I think this error message must be coming from the Qt web widget or somewhere else in Qt. These lines (above) are the only thing that I can imagine would be affecting things and all they do is cause `pyopengl` to be imported--no opengl programs are compiled just importing them.

pllim:
[...]

PR #861, opened 2020-07-18
Fix Qt/PySide help crash

### LINKED ISSUE #841: Segfault when closing Ginga Help window (Qt, PySide backends)
Platform: Linux with PySide2 as the Qt environment

When exiting Ginga, there is a segmentation fault, but only if a Help button has been pressed during the session AND the Help window is still open when quitting Ginga.

ejeschke:
(Fix #841)

### DIFF
--- a/doc/WhatsNew.rst
+++ b/doc/WhatsNew.rst
@@ -27,6 +27,8 @@ Ver 3.1.0 (unreleased)
   wrong results to be displayed in PixTable
 - New ability to specify alternative Ginga home directories, with custom
   layouts and plugin configurations (--basedir option)
+- Fix for a bug that caused a crash when closing the Help window with
+  Qt/PySide backend
 
 Ver 3.0.0 (2019-09-20)
 ======================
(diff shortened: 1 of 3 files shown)


ejeschke:
@pllim, this prevents crash when closing Help window in Qt/PySide.  There is still a crash on program exit in PySide if you leave the Help window open until program exit, but I think that moves the problem back into lower priority. Qt backend is fine whether you close the Help window right away or leave it open until program exit. It seems the QWebView widget is a bit sensitive.

pllim:
I cannot test this myself due to #837 , so I guess I'll just have to trust you on this one. FFTM

ejeschke:
> I cannot test this myself due to #837

Sorry, @pllim, don't understand this comment. Meaning you can't reproduce the crash when you press the Close button on `Help`?

pllim:
Sorry for the confusion. I meant, I still get segfault even if this patch, but I think it is a different problem?

ejeschke:
Ah, right.  Reviewing that issue, I think this will solve the problem @rendinam saw with `PySide`.  Ok, merging.

PR #860, opened 2020-07-18
Downgrade channel_Image.cfg not found from warning to info

pllim:
I don't like the warning because even if I am okay with defaults (in the even of that file missing), I have to create the file to silence the warning. Therefore, I propose to downgrade it from warning to info. What do you think?

*Before* this patch:
```
2020-07-18 15:37:57,505 | W | Control.py:1429 (add_channel) |
    no saved preferences found for channel 'Image':
    Error opening settings file (/home/username/.ginga/channel_Image.cfg):
    [Errno 2] No such file or directory: '/home/username/.ginga/channel_Image.cfg'
```

With this patch:
```
2020-07-18 15:59:25,384 | I | Control.py:1429 (add_channel) |
    no saved preferences found for channel 'Image', using default:
    Error opening settings file (/home/username/.ginga/channel_Image.cfg):
    [Errno 2] No such file or directory: '/home/username/.ginga/channel_Image.cfg'
```

PR #865, opened 2020-07-20
MNT: Downgrade layout warning to info

pllim:
Direct follow up of #763 . Same thought process as #860.

### DIFF
--- a/ginga/gw/Desktop.py
+++ b/ginga/gw/Desktop.py
@@ -685,8 +685,7 @@ class Desktop(Callback.Callbacks):
                 layout = self.read_layout_conf(lo_file)
 
             except Exception as e:
-                self.logger.warning("Error reading saved layout: %s" % (str(e)))
-                self.logger.info("reverting to default layout")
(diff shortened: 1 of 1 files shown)


pllim:
Does assigning it to me means you are okay with this? 😹

ejeschke:
> Does assigning it to me means you are okay with this?

@pllim, I don't really know what that assign business is good for with this particular project, but I just figure "your" PRs are yours and "my" PRs are mine and "our" PR are "ours". :smile:

ejeschke:
Are you seeing the warning over and over after deleting your layout files and letting Ginga regenerate them?

pllim:
> This warning shouldn't normally be issued

I only see it when I manually delete the file in my `~/.ginga` to get a fresh copy. I feel like it shouldn't be a warning because it is a no-op for a user. If you disagree, we can close this without merging.

ejeschke:
@pllim, see #866 as an alternate that avoid issuing a warning if the file does not exist.

PR #872, opened 2020-07-30
TST: Use GitHub Actions

### LINKED ISSUE #771: TST: Port CI to GitHub Actions?
Maybe? Only if you are interested, @ejeschke .

pllim:
Fix #771

Uses https://github.com/jurplel/install-qt-action/ to install Qt.

- [x] Fixed failures reported in https://github.com/pllim/ginga/actions (this will show up under "Actions" tab above after merge)
- [x] Check results against existing Travis CI; e.g., https://travis-ci.org/github/ejeschke/ginga/builds/713146934


pllim:
Hmm... Actions didn't kick off.

pllim:
I think the new setup is pretty close(?) to the old one, except that I have to bump Numpy 1.15 to 1.16, which is reasonable given that the stable Numpy is now 1.19. I even added OSX job as a bonus!

With Actions, there is no more `conda` nor `ci-helpers` involved. Actions also does not respect `[ci skip]` for a pull request, but given that this is a relatively low-traffic package, should not matter. You can cancel jobs manually via the Actions tab if you really have to kill them.

The ginga version reported at https://github.com/pllim/ginga/actions  is wrong but that is because it is pulling tags from my fork (which has none). Once this is merged, it will pull tag from this repo and it will be correct again.

ginga/colors.py
```diff
@@ -781,7 +781,7 @@ def lookup_color(name, format='tuple'):
 def resolve_color(color):
     if isinstance(color, str):
         r, g, b = lookup_color(color)
-    elif isinstance(color, collections.Sequence):
+    elif isinstance(color, collections.abc.Sequence):
```
pllim, line 784: This emits deprecation warning in Python 3.8. Even though we currently have a setting under `ginga/conftest.py` to turn all deprecation warnings to errors, it is not happening on Travis. I moved `conftest.py` to root dir here and it seems to work again.

pllim: Ah, wait, maybe Travis never used Python 3.8 in the first place! No matter, `conftest.py` should go to root dir anyway.

pllim:
What do you think, @ejeschke ? You like Actions?

.github/workflows/ci_workflows.yml
```diff
@@ -0,0 +1,99 @@
...
+  pull_request:
+  schedule:
+    # Weekly Monday 6AM build
+    # * is a special character in YAML so you have to quote this string
+    - cron: '0 6 * * 1'
+
+jobs:
+  pep:
+    runs-on: ubuntu-16.04
+    steps:
```
ejeschke, line 14: Ubuntu 16.04 is the latest they have? Seems like 20.04 has been out a few months, I'd hope they'd at least have 18.04...

ejeschke:
@pllim, just wondering whether we should add a job ("action"?) for a miniconda based test. I'm just guessing, but I'll bet most of our users are running on some kind of conda environment, rather than using system Python. Or is the conda installation "behind the scenes" somehow?

.github/workflows/ci_workflows.yml
pllim: For PEP, I guess we can use `ubuntu-latest`. Good catch!

pllim:
Yes, this PR moves everything to pip or system install and does not use conda at all.

To use conda, I think we might need some custom action from their "marketplace" (it is free even though they say "market"). I'll have to ask around if there is a recommended and secure way to do this.

pllim:
Not using conda makes the CI runs so much faster though. 😉

ejeschke:
> Not using conda makes the CI runs so much faster though.
I wonder if there is any way to preserve the conda environment across tests, so it doesn't have to download conda and then have conda re-download everything.

Wonder if we could set up something like a docker environment with all the miniconda install already done...

pllim:
But... what does using conda in Actions really test? Even if things pass, it does not guarantee Ginga's next release to conda-forge would work, because conda-forge uses its own recipe to build Ginga. What value does this add?

ejeschke:
> But... what does using conda in Actions really test?

Just testing in the environment that the vast (?) majority of users are using. Do you know anyone that installs their astronomical packages in the system python environment? That way lies madness... :smile_cat:

ejeschke:
@pllim, is `astropy` planning to migrate to Actions?

ejeschke:
I see from some quick searches that adding a conda setup adds a huge time penalty.

ejeschke:
> I see from some quick searches that adding a conda setup adds a huge time penalty.

I wonder if that is true even if you [enable caching](https://github.com/marketplace/actions/setup-miniconda#caching).

tepickering:
since @pllim tagged me in on slack and i've been hacking on github actions for a while, i'll add my €0.02. the ubuntu vm that actions uses does provide `miniconda`: https://github.com/actions/virtual-environments/blob/main/images/linux/Ubuntu2004-README.md. so the `conda` command is there for you to use in your workflows however you like. the 3rd party action linked above might help simplify the process of using `conda`, but is not required. however, as you note, using `conda` does add a time penalty. this is why many of us just use `tox` and `pip` for installing dependencies for testing. setting up a docker container with your testing environment is another option, but i believe that limits you to linux containers in addition to being more complicated.

the one use i've seen for testing with `conda` is to test/benchmark against the MKL libraries. when i tried to install those via `pip` a while back, it didn't work with python 3.8.x. i ended up not using MKL in production so didn't follow up further with that.

tepickering:
see https://github.com/actions/virtual-environments/issues/1003 for some more discussion on using the built-in `conda` vs the `setup-miniconda` action.

pllim:
Thanks, @tepickering !

Re: astropy's plans -- It won't happen so soon, but there are discussions at astropy/astropy#8445 and astropy/astropy#10388

tepickering:
actually, i should point out that ALL of the the vm's that github actions uses provide miniconda. so it covers windows and mac as well as ubuntu.

pllim:
Even those miniconda seems to be provided, setting it up is another story. LoL. I think for our purpose here, just one Linux job for it is good enough. I hope the code from actions/starter-workflows#197 still works.

tepickering:
the implication from that thread i linked to was that `conda` should now be linked into `/usr/local/bin`, but maybe that's not the case? at any rate, using `$CONDA` should always work and for a single test using the base env is probably fine, too. saves the hassle of dealing with how `activate` interacts with the shell.

ejeschke:
@tepickering, @pllim thanks for the comments.  After thinking about this for a bit, it seems to me that for the purposes of running unit tests, it's probably sufficient to run them with the system versions of Python. But I still think that if there is a not too unruly way to do a miniconda env test on at least one platform (preferably Linux) that would be "reasonably" efficient, it would be really useful, since it is so widely used in this ecosystem.  But I am ok to add that later (@pllim, maybe put on the roadmap if we merge this without a miniconda test env ?)

pllim:
I added one conda job and it seems to work as far as I can tell from the log,see https://github.com/pllim/ginga/actions/runs/209033803 . It is actually not much slower. Maybe because we don't have a lot of dependencies and avoided overhead of creating a new env.

pllim:
I think this is ready for final review.

pllim:
Adding `pillow` can be a future PR, though if it is not used in tests, not sure what value that adds.

pllim:
First run on `master` passed. 😸 

https://github.com/ejeschke/ginga/actions

PR #873, opened 2020-07-30
Ginga will remember sizes and placement of windows in MDI workspaces

ejeschke:
This allows Ginga to remember the size and placement of the windows in workspaces that have been configured as MDI. If you close the reference viewer with some MDI windows opened and positioned, then start it again and open windows with the same names, they will be opened at the locations and sizes that they were when the program was closed.


ejeschke:
No hurry, @pllim. This is a minor enhancement that only affects MDI windows.

pllim:
FFTM

PR #874, opened 2020-08-13
ENH: Add copy cut feature

### LINKED ISSUE #820: Feature Request: Make new cut from copy in Cuts
Just something that crossed my mind while trying out `Cuts` in #812. I made a cut line and I wanted to create an exact copy of the same cut line but apply it to another region, so I can compare similar cuts across different regions in the image. Making them exact is hard to do by hand.

This is not high priority but maybe would be nice to have?

pllim:
Fix #820 

**TODO**

- [x] How to copy existing object? Encountered `pickle.PicklingError('logger cannot be pickled')` -- Wait for #875 to be merged and then use it.
- [x] Test the following use cases:
    * Copy from one
    * Copy from copy
    * Copy from another


ginga/rv/plugins/Cuts.py
```diff
@@ -444,6 +456,33 @@ def set_cutsdrawtype_cb(self, w, index):
...
+
+        new_index = self._get_new_count()
+        new_tag = "cuts{}".format(new_index)
+
+        old_obj = self.canvas.get_object_by_tag(old_tag)
+        self.logger.debug("copying {}".format(old_obj.objects[0]))
+
+        # FIXME: deepcopy breaks because logger cannot be pickled
+        new_cut = self._create_cut_obj(
+            new_index, deepcopy(old_obj.objects[0]), color='cyan')
```
pllim, line 474: @ejeschke , I can't get `deepcopy` to work. And the existing API does not appear to have a `copy` method. Any advise? Thanks!

ginga/rv/plugins/Cuts.py
```diff
@@ -93,6 +93,12 @@
 Use the "Save" button to save the ``Cuts`` plot as as image and
 data as a Numpy compressed archive.
 
+**Copying Cuts**
+
+To copy a cut, select its name from the "Cut" dropdown and click the
+"Copy Cut" button. A new cut will be created from it. You can then manipulate
+the new cut indepedently.
```
pllim, line 100: Ops, typo.
```suggestion
the new cut independently.
```

ginga/rv/plugins/Cuts.py
ejeschke: @pllim, recommend adding a `copy()` method to `~ginga.canvas.CanvasObject`, so that we can benefit from it in the future (then just call `copy()` on the object here).

For implementing `copy()` I think you can so something like:
```python
from copy import deepcopy
...
def copy(self): 
    logger = self.logger
    try:
        self.logger = None
        obj = deepcopy(self)
        obj.logger = logger
        return obj
    finally:
        self.logger = logger
```
There might be some other things that should not be shared--we can figure those out later.

ejeschke:
You want the new cut to appear exactly in the same position as the old one?  In the imagined use case, does the user immediately move the cut to a new location?  If so, then might be nice to default to offsetting the cut by a few pixels in X and Y.

ejeschke:
> If so, then might be nice to default to offsetting the copy by a few pixels in X and Y

Since objects have a `move_delta_pt(off_pt)` method, you could do something (after the copy and add to canvas) like `obj.move_delta_pt((off_x, off_y))`.

ginga/rv/plugins/Cuts.py
ejeschke: See code comment under "Files changed"

ejeschke: Hmm, just tried this and now it fails trying to deepcopy `threading.RLock` objects.

ejeschke:
@pllim, have a look at #875. I think the functionality of that `copy()` method might be good enough for use in this PR.

pllim:
#875 works fine for the purpose of this PR. I left a review there as well. Thanks!

As for the slight offset, yes, that was I wanted but I just haven't gotten around to it, but looks like your PR has that too, so that's great.

I'll wait for #875 to get merged before proceeding here.

pllim:
@ejeschke , this is ready for final review. Also, I added change logs for #875 here as well.

pllim:
> Do you plan to update the screenshot?

@ejeschke , I can but the image will be different from what is it the doc right now. I don't have that particular 2D spectrum image.

pllim:
@ejeschke , I updated the screenshot and also fixed a documentation bug. How does this look? https://ginga--874.org.readthedocs.build/en/874/manual/plugins_local/cuts.html

ejeschke:
Looks good!

pllim:
Thanks! Do you want to merge? 😸

ejeschke:
FFTM!

pllim:
Thanks!

PR #875, opened 2020-08-14
Add copy() method to canvas objects

ejeschke:
Adds a `copy()` method to Ginga canvas objects.

This makes a shallow copy, which may not always work according to the intended purposes.  The `Drawing` plugin is modified to be able to use this to copy a selected object in edit mode.

In support of #874

### DIFF
--- a/ginga/canvas/CanvasObject.py
+++ b/ginga/canvas/CanvasObject.py
@@ -6,6 +6,7 @@
 #
 import numpy as np
 from collections import namedtuple
+import copy
 
 from ginga.misc import Callback, Bunch
 from ginga import trcalc, colors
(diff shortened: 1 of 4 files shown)


ejeschke:
@pllim, try `Drawing` plugin to see how this works with various objects.  It might be good enough for use in #874.

ejeschke:
Works for most basic objects. Won't work for `Annulus` type implementation without an explicit method override and implementation of `copy`. Probably could just add a method that raises a `NotImplemented` error or something like that.

`CompoundObject`, etc. might work ok, but it is a shallow copy, so all constituent objects would be shared.

ejeschke:
> As for compound shapes, if it loops through all its self.objects and creates a copy for each and then returns a new compound with those copies, it is essentially a deep copy?

I added a method for compound objects that does the above. It seems to support `Annulus` just fine (as far as `Drawing` is concerned). I think for now let's not muddle the waters too much by defining shallow vs. deep for this method. A copy of a canvas object is (for all practical purposes) a new one that can be plotted, changed to have different attribute values, etc.  I think that intuitively speaks to a deep copy, but if we later encounter a situation where a shallow one might be useful, we can add a keyword option.

ejeschke:
Bonus: added unit tests module for canvas.  Lots of things we can add to that in the future!

pllim:
Hmm, does this need an entry in `doc/WhatsNew.rst`?

PR #876, opened 2020-08-14
TST: Add security audit job

pllim:
Follow up of #872 . This uses [bandit](https://bandit.readthedocs.io/en/latest/) package to do simple security audit. If you think this is not necessary for Ginga, we can close without merge.

### DIFF
--- a/ginga/gtk3w/GtkHelp.py
+++ b/ginga/gtk3w/GtkHelp.py
@@ -401,9 +401,9 @@ class MDIWidget(Gtk.Layout):
         # pick a random spot to place the window initially
         rect = self.get_allocation()
         wd, ht = rect.width, rect.height
-        x = random.randint(self.cascade_offset,
+        x = random.randint(self.cascade_offset,  # nosec
                            max(self.cascade_offset + 10, wd // 2))
-        y = random.randint(self.cascade_offset,
+        y = random.randint(self.cascade_offset,  # nosec
                            max(self.cascade_offset + 10, ht // 2))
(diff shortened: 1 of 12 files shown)


ginga/gtk3w/GtkHelp.py
```diff
@@ -401,9 +401,9 @@ def append_page(self, widget, label):
         # pick a random spot to place the window initially
         rect = self.get_allocation()
         wd, ht = rect.width, rect.height
-        x = random.randint(self.cascade_offset,
+        x = random.randint(self.cascade_offset,  # nosec
```
ejeschke, line 404: Hmm, it flags use of `random` module due to seed not being truly random enough?

pllim: No, it's actually not that relevant unless we doing crypto. See https://bandit.readthedocs.io/en/latest/blacklists/blacklist_calls.html#b311-random

pllim:
> do you think the security audit is "comprehensive" enough to be useful?

I am not sure, but I think it doesn't hurt. For a serious security audit though, if necessary, it should be the _only_ tool, but it is a good start.

> how much time does it add to our CI?

The log says it added 27 seconds.

> Your call

I would merge it unless you have objection. We can always remove it in the future if you change your mind. Do you agree?

ejeschke:
Yikes, 27 seconds.  We are quickly ramping back up to longer CI.  But let's go ahead and merge.  If the build becomes too long it might be one of first to get commented out.

PR #877, opened 2020-08-20
Use np instead of np.ma for iqcalc

pllim:
Is there a reason to use `np.ma.` functions instead of `np.` ones? Do you pass in masked arrays?

### DIFF
--- a/ginga/util/iqcalc.py
+++ b/ginga/util/iqcalc.py
@@ -40,7 +40,7 @@ def get_mean(data_np):
     i = np.isfinite(data_np)
     if not np.any(i):
         return np.nan
-    return np.ma.mean(data_np[i])
+    return np.mean(data_np[i])
 
 
 def get_median(data_np):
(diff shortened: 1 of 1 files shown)


pllim:
Failing job is unrelated and being addressed at #878

ejeschke:
Masked arrays are created by the `cutout_shape()` method. I switched to the 'ma' versions because I was getting a numpy warning about masked elements being ignored (in the very functions you have changed back here).

ejeschke:
Try your PR doing a `Pick` with any shape other than a rectangle or box.

pllim:
Okay, never mind then. It isn't clear by just looking at the unit tests. Sorry for the noise.

ejeschke:
Maybe add some unit tests for other shapes and some comments in the code about why we are doing a masked array operation?  What do you think, @pllim?

ejeschke:
I can do that.

pllim:
Thanks!

PR #878, opened 2020-08-20
MNT: Compatibility with astropy.coordinates

pllim:
This deprecation warning started showing up and failing astropy-dev job:

```
E           astropy.utils.exceptions.AstropyDeprecationWarning: Transforming a frame instance to a frame class
(as opposed to another frame instance) will not be supported in the future.  Either explicitly instantiate the
target frame, or first convert the source frame instance to a `astropy.coordinates.SkyCoord` and
use its `transform_to()` method.
```

This is caused by astropy/astropy#10475 .

This PR proposes a fix for the astropy-dev job. It also clarifies that minimum version of `astropy` that we really support and added that to one of the jobs for testing.

### DIFF
--- a/.github/workflows/ci_workflows.yml
+++ b/.github/workflows/ci_workflows.yml
@@ -69,7 +69,7 @@ jobs:
     - name: Install and build
       run: |
         python -m pip install --upgrade pip setuptools
-        python -m pip install PyQt5 numpy==1.16
+        python -m pip install PyQt5 numpy==1.16 astropy==3.2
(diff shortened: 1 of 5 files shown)


pllim:
> maybe we should just keep around an empty instance of each kind of frame, so that we can just reuse them

Hmm, perhaps @mhvk has recommendations?

ejeschke:
>  I added some code to cache the instance. What do you think?

This looks good.

pllim:
Thanks for the review!

PR #880, opened 2020-08-20
New iqcalc_astropy module to use astropy to fit FWHM

### LINKED ISSUE #754: Feature Request: Add Lorentz function to Pick plugin
Someone asked me if Pick can handle Gaussian, Moffat, and Lorentz, but only first two are supported right now.

pllim:
This fixes #754 by implementing Lorentz fitting, but it is only available if `iqcalc_astropy` is used over `iqcalc`.

**TODO**

- [x] Add tests for new stuff.
- [x] Compare results between `iqcalc` and `iqcalc_astropy`. (This can be part of the tests too.)
- [x] Rebase after #881 is merged.
- [x] Figure out a way for Ginga to swap between the two modules: New entry in ~`general.cfg`~ `plugin_Pick.cfg`?
- [x] See if `Pick` still works with both.
- [x] Update change log entry when more stuff above is complete.

**Screenshot**

ginga_iqcalc_astropy

```python
import numpy as np           

from ginga.util.iqcalc_astropy import IQCalc

arr1d = np.array([0., 9., 0., 0., 0., 34., 25., 60., 196., 602., 
                  1117., 1003., 413., 135., 29., 0., 3., 0., 4., 3.])
x = list(range(arr1d.size))

iqcalc = IQCalc() 
res_gaussian = iqcalc.calc_fwhm_gaussian(arr1d)
res_moffat = iqcalc.calc_fwhm_moffat(arr1d)
res_lorentz = iqcalc.calc_fwhm_lorentz(arr1d)

fit_gaussian = res_gaussian.fit_fn(x, res_gaussian.fit_args)
fit_moffat = res_moffat.fit_fn(x, res_moffat.fit_args)
fit_lorentz = res_lorentz.fit_fn(x, res_lorentz.fit_args)

fig = plt.figure()
ax1 = fig.add_subplot(311)
ax2 = fig.add_subplot(312)
ax3 = fig.add_subplot(313)

ax1.plot(arr1d)
ax2.plot(arr1d)
ax3.plot(arr1d)

ax1.plot(fit_gaussian, 'r--')
ax2.plot(fit_moffat, 'r--')
ax3.plot(fit_lorentz, 'r--')

ax1.text(0.05, 0.95, f'Gaussian FWHM: {res_gaussian.fwhm:.4f}', transform=ax1.transAxes, verticalalignment='top')
ax2.text(0.05, 0.95, f'Moffat FWHM: {res_moffat.fwhm:.4f}', transform=ax2.transAxes, verticalalignment='top')
ax3.text(0.05, 0.95, f'Lorentz FWHM: {res_lorentz.fwhm:.4f}', transform=ax3.transAxes, verticalalignment='top')

plt.show()
```

### DIFF
--- a/ginga/util/iqcalc_astropy.py
+++ b/ginga/util/iqcalc_astropy.py
@@ -1,0 +1,315 @@
+"""Module to handle image quality calculations using ``astropy``."""
+
+import numpy as np
+from astropy.modeling import models, fitting
+
+from ginga.misc import Bunch
+
+# Reuse shared items from the old module until we can merge old and new together.
+from ginga.util.iqcalc import IQCalcError, get_median, have_scipy
+from ginga.util.iqcalc import IQCalc as _IQCalc
+
+__all__ = ['IQCalc']
+
+
+# TODO: Use photutils for source finding.
+class IQCalc(_IQCalc):
+    """This is `ginga.util.iqcalc.IQCalc` that uses ``astropy``.
+
+    This subclass has an extra ``self.fitter`` attribute for ``astropy``
+    fitting.
+
+    """
+
+    def __init__(self, *args, **kwargs):
+        super().__init__(*args, **kwargs)
+        self.fitter = fitting.LevMarLSQFitter()
+
+    # FWHM CALCULATION
+
+    def _prep_for_fitting(self, arr1d, medv):
+        if not have_scipy:
+            raise IQCalcError("Please install the 'scipy' module "
+                              "to use this function")
+
+        N = len(arr1d)
+        X = np.array(list(range(N)))
+        Y = np.asarray(arr1d)
+
+        # Fitting works more reliably if we do the following
+        # a. subtract sky background
+        if medv is None:
+            medv = get_median(Y)
+        Y -= medv
+        maxv = Y.max()
+        # b. clamp to 0..max (of the sky subtracted field)
+        Y = Y.clip(0, maxv)
+
+        return N, X, Y, maxv
+
+    def gaussian(self, x, p):
+        """Evaluate Gaussian function in 1D.
+
+        Parameters
+        ----------
+        x : array-like
+            X values.
+
+        p : tuple of float
+            Parameters for Gaussian, i.e., ``(mean, stddev, amplitude)``.
+
+        Returns
+        -------
+        y : array-like
+            Y values.
+
+        """
+        g = models.Gaussian1D(amplitude=p[2], mean=p[0], stddev=p[1])
+        return g(x)
+
+    def calc_fwhm_gaussian(self, arr1d, medv=None, **kwargs):
+        """FWHM calculation on a 1D array by using least square fitting of
+        a Gaussian function on the data.
+
+        Parameters
+        ----------
+        arr1d : array-like
+            1D array cut in either X or Y direction on the object.
+
+        medv : float or `None`
+            Median of the data. If not given, it is calculated from ``arr1d``.
+
+        kwargs : dict
+            Not used; for backward-compatible API call only.
+
+        Returns
+        -------
+        res : `~ginga.misc.Bunch.Bunch`
+            Fitting results.
+
+        Raises
+        ------
+        IQCalcError
+            Fitting failed.
+
+        """
+        N, X, Y, maxv = self._prep_for_fitting(arr1d, medv)
+
+        # Gaussian model with initial guess
+        m_init = models.Gaussian1D(amplitude=maxv, mean=0, stddev=(N - 1))
+
+        # NOTE: without this mutex, optimize.leastsq causes a fatal error
+        # sometimes--it appears not to be thread safe.
+        # The error is:
+        # "SystemError: null argument to internal routine"
+        # "Fatal Python error: GC object already tracked"
+        with self.lock:
+            try:
+                m = self.fitter(m_init, X, Y)
+            except Exception:
+                raise IQCalcError("FWHM Gaussian fitting failed")
+
+        # Now that we have the sdev from fitting, we can calculate FWHM
+        fwhm = m.fwhm
+        # Some routines choke on numpy values and need "pure" Python floats
+        # e.g. when marshalling through a remote procedure interface
+        mu = m.mean.value
+        sdev = m.stddev.value
+        maxv = m.amplitude.value
+
+        self.logger.debug('mu={} sdev={} maxv={}'.format(mu, sdev, maxv))
+
+        res = Bunch.Bunch(fwhm=fwhm, mu=mu, sdev=sdev, maxv=maxv,
+                          fit_fn=self.gaussian, fit_args=[mu, sdev, maxv])
+        return res
+
+    def moffat(self, x, p):
+        """Evaluate Moffat function in 1D.
+
+        Parameters
+        ----------
+        x : array-like
+            X values.
+
+        p : tuple of float
+            Parameters for Moffat, i.e., ``(x_0, gamma, alpha, amplitude)``,
+            where ``x_0`` a.k.a. mean and ``gamma`` core width.
+
+        Returns
+        -------
+        y : array-like
+            Y values.
+
+        """
+        m = models.Moffat1D(amplitude=p[3], x_0=p[0], gamma=p[1], alpha=p[2])
+        return m(x)
+
+    def calc_fwhm_moffat(self, arr1d, medv=None, **kwargs):
+        """FWHM calculation on a 1D array by using least square fitting of
+        a Moffat function on the data.
+
+        Parameters
+        ----------
+        arr1d : array-like
+            1D array cut in either X or Y direction on the object.
+
+        medv : float or `None`
+            Median of the data. If not given, it is calculated from ``arr1d``.
+
+        kwargs : dict
+            Not used; for backward-compatible API call only.
+
+        Returns
+        -------
+        res : `~ginga.misc.Bunch.Bunch`
+            Fitting results.
+
+        Raises
+        ------
+        IQCalcError
+            Fitting failed.
+
+        """
+        N, X, Y, maxv = self._prep_for_fitting(arr1d, medv)
+
+        # Moffat model with initial guess
+        m_init = models.Moffat1D(amplitude=maxv, x_0=0, gamma=(N - 1), alpha=2)
+
+        # NOTE: without this mutex, optimize.leastsq causes a fatal error
+        # sometimes--it appears not to be thread safe.
+        # The error is:
+        # "SystemError: null argument to internal routine"
+        # "Fatal Python error: GC object already tracked"
+        with self.lock:
+            try:
+                m = self.fitter(m_init, X, Y)
+            except Exception:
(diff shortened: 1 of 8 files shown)


ejeschke:
@pllim, concurrently I pushed a couple of tests for `calc_fwhm_gaussian()` and `calc_fwhm_moffat()` (along with tests for masked data with `get_mean()` and `get_median()`. I would be interested to see how the results compare for the `astropy` fitting measurements...

ejeschke:
> I added some plots above. I did have to adjust how the initial parameters are set because astropy is more sensitive to them. What do you think?

@pllim, do you mean that the default plot for `Pick` cannot handle the plots generated by the new fittings? They are just points on a curve, right?

pllim:
> https://github.com/ejeschke/ginga/pull/880#issuecomment-678888402

No, I haven't hooked it up to `Pick` yet. I mean in the `iqcalc_astropy` itself, the algorithm for initial guess is a little different.

pllim:
Update: I implemented an option in Pick config to switch between `iqcalc` and `iqcalc_astropy`. Since this is a widely used plugin, you might want to play around with it carefully before merging.

When I try fitting Lorentz, sometimes I get this:
```
2020-08-27 13:55:52,143 | E | Pick.py:1587 (update_pick) | Error calculating quality metrics: No object matches selection criteria
2020-08-27 13:55:52,144 | E | Pick.py:1594 (update_pick) |   File ".../ginga/rv/plugins/Pick.py", line 1511, in update_pick
    raise Exception(msg)

```

Hard to tell if Lorentz is simply not a good choice, or if it is a bug of my implementation. Any idea to debug would be appreciated.

Moffat would sometimes raise warnings too, but I see that with both `iqcalc` and `iqcalc_astropy`. Perhaps it is just not a very good choice for stars. So, some sample images and expected results would also be nice to test this out.

ejeschke:
> Any idea to debug would be appreciated.

@pllim, that message means that according to the settings it could not detect sources that matched the parameters as defined under the `Settings` tab.

In such cases you may wish to tweak the ellipticity, edge, min fwhm, max fhwm, radius or threshold parameters.

ejeschke:
`iqcalc` does fhwm measurements as a part of checking all the possible sources, and if it can't get a fit with Lorentz, or the fit results in a measurement that puts it outside the range of min/max fhwm, then that may definitely result in the above message.

ejeschke:
@pllim, is it ready for testing?

pllim:
> is it ready for testing?

Yes, please. Thanks!

ejeschke:
When testing, it looks like the cutout array is side-effected:
iqcalc_astropy

and this affects the plots (and fitting):
iqcalc-astropy-contour

ginga/util/iqcalc_astropy.py
```diff
@@ -0,0 +1,321 @@
...
+
+        N = len(arr1d)
+        X = np.array(list(range(N)))
+        Y = np.asarray(arr1d)
+
+        # Fitting works more reliably if we do the following
+        # a. subtract sky background
+        if medv is None:
+            medv = get_median(Y)
+        Y -= medv
```
ejeschke, line 46: If I change this to:
```python
Y = Y - medv
```
then it solves the problem of side-effecting the cutout array.

ejeschke:
> Hard to tell if Lorentz is simply not a good choice, or if it is a bug of my implementation. Any idea to debug would be appreciated.

If I lower the `min FWHM` to 1.0 (from 2.0) then Lorentz seems to work fine.  We can adjust this default, but we don't want to go so low that we are picking up artefacts like detector defects...

ejeschke:
The "astropy-ifed" `iqcalc` does not seem noticeably slower. That's good news.

ejeschke:
Needs a rebase.

ejeschke:
Lorentz looks like a higher, narrower, "pointier" fit than either Gaussian or Moffat. I think this why the FWHM numbers tend to skew smaller.

pllim:
> cutout array is side-effected

Oh, no! How can that be? I only touched the FWHM stuff. I'll investigate on Monday.

ginga/util/iqcalc_astropy.py
pllim: Oh, weird... Maybe its `isub` isn't behaving as expected.

ejeschke: I suspect that it depends on what kind of array you are using it with. If the cutout ends up just being a view into the cutout, then doing `-=` probably drops the hatchet on the cutout.

pllim: Good to know! Sorry about that.

pllim:
@ejeschke , I fixed the `-=` bug and rebased. I think the cutout bug is gone, but you should double check.

Using `astropy` for fitting, there is an extra warning like this but it still shows results, so not sure if we should fail it or not:

```
2020-08-29 13:12:03,191 | W | warnings.py:109 (_showwarnmsg) |
  .../astropy/modeling/functional_models.py:2230:
  RuntimeWarning: invalid value encountered in sqrt
    return 2.0 * np.abs(self.gamma) * np.sqrt(2.0 ** (1.0 / self.alpha) - 1.0)

2020-08-29 13:12:03,322 | W | warnings.py:109 (_showwarnmsg) |
  .../astropy/modeling/fitting.py:919: AstropyUserWarning:
  The fit may be unsuccessful; check fit_info['message'] for more information.
```

For the "invalid value encountered in sqrt", it would happen to any formula with `sqrt`, not just that one function.

pllim:
> We can adjust this default, but we don't want to go so low

How about 1.5?

> Lorentz looks like a higher, narrower

Yes. You can see how they compare by looking at [Gaussian1D](https://docs.astropy.org/en/stable/api/astropy.modeling.functional_models.Gaussian1D.html#astropy.modeling.functional_models.Gaussian1D), [Moffat1D](https://docs.astropy.org/en/stable/api/astropy.modeling.functional_models.Moffat1D.html#astropy.modeling.functional_models.Moffat1D), and [Lorentz1D](https://docs.astropy.org/en/stable/api/astropy.modeling.functional_models.Lorentz1D.html#astropy.modeling.functional_models.Lorentz1D) (scroll down a bit for the plot). Lorentz profile is usually used in spectroscopy. I forgot why the scientist requested it all that time ago. Maybe for some galaxies? 🤷

ejeschke:
> How about 1.5?

That sounds reasonable. Does it give you acceptable results for your typical stellar images?

ejeschke:
>  I think the cutout bug is gone, but you should double check.

I confirmed the cutout bug is vanquished.

ejeschke:
```python
astropy/modeling/fitting.py:894: AstropyUserWarning: The fit may be unsuccessful; check fit_info['message'] for more information.
  AstropyUserWarning)
```
This warning could get annoying; I seem to get it sometimes even when there is a fit. Any way to either trap it, or quantify it somehow to see if we can weed out the spurious results?  I guess if it happens mostly with Lorentz then it's not such a big problem. We could say it is a "feature". :confused:

pllim:
> Does it give you acceptable results for your typical stellar images?

As far as I can test with one image, which isn't much.

> This warning could get annoying

I can do something like this below, but `Pick` fails with `Error calculating quality metrics: No object matches selection criteria` before it gets there. I cannot seem get that fit warning to trigger today to test this. 🤷 How should I proceed?

```diff
diff --git a/ginga/util/iqcalc_astropy.py b/ginga/util/iqcalc_astropy.py
index 68d09673..42856f5b 100644
--- a/ginga/util/iqcalc_astropy.py
+++ b/ginga/util/iqcalc_astropy.py
@@ -1,7 +1,10 @@
 """Module to handle image quality calculations using ``astropy``."""
 
+import warnings
+
 import numpy as np
 from astropy.modeling import models, fitting
+from astropy.utils.exceptions import AstropyUserWarning
 
 from ginga.misc import Bunch
 
@@ -267,9 +270,16 @@ class IQCalc(_IQCalc):
         # "Fatal Python error: GC object already tracked"
         with self.lock:
             try:
-                m = self.fitter(m_init, X, Y)
-            except Exception:
-                raise IQCalcError("FWHM Lorentz fitting failed")
+                with warnings.catch_warnings(record=True) as warning_lines:
+                    m = self.fitter(m_init, X, Y)
+                for w in warning_lines:
+                    if w.category == RuntimeWarning:
+                        raise ValueError(f'{str(w.message)}: {m.fwhm}')
+                    elif w.category == AstropyUserWarning:
+                        raise ValueError(f'{str(w.message)}: {self.fitter.fit_info["message"]}')
+
+            except Exception as e:
+                raise IQCalcError(f"FWHM Lorentz fitting failed: {repr(e)}")
 
         # Some routines choke on numpy values and need "pure" Python floats
         # e.g. when marshalling through a remote procedure interface
```

pllim:
Sometimes, I also see this error but with the `except` catch-all, I cannot find where it came from.

```
2020-09-01 16:24:40,740 | E | Pick.py:1346 (_redo) | Error calculating quality metrics: 'float' object cannot be interpreted as an integer
```

ejeschke:
> Sometimes, I also see this error but with the except catch-all, I cannot find where it came from.

Can you add the `exc_info=True` kwarg to the logging statement so that you can see the stack dump?

ejeschke:
> How should I proceed?

If you are not seeing the warning too much, and mostly with Lorentz, then maybe we can just treat it as a possible future issue if it gets reported again.

I am mostly wondering about fits that succeed, but astropy is not convinced and wants to tell us that maybe it did not succeed.  If we can quantity that uncertainty, instead of a vague error message we could do the plot in a different (warning) color perhaps--to indicate that it is a "rough fit".

pllim:
In trying to trap the warnings, I sometimes now see:

https://github.com/ejeschke/ginga/blob/c3b80e71624ea0955762e6dfcf96f695275e9fa9/ginga/util/iqcalc.py#L381-L383

Not sure why that is happening.

pllim:
@ejeschke , after some investigation, turns out sometimes the RuntimeWarning appears because the fitter tries some invalid values while fitting. Also see astropy/astropy#10708 . So, the warning is harmless if the fit converges because you get back a real FWHM anyway.

If it does not converge, I think `Pick` already handles that situation anyway. So, I think adding more logic now is probably overkill. I propose we "treat it as a possible future issue" as you suggested.

So, is there anything else I need to do for this PR before it is good to merge? Thanks.

ejeschke:
@pllim, did you ever manage to reproduce this one: "Error calculating quality metrics: 'float' object cannot be interpreted as an integer" ?

pllim:
Re: https://github.com/ejeschke/ginga/pull/880#issuecomment-686336613

@ejeschke , actually, no. I suspect maybe that was a result of me playing with warning trapping in `iqcalc_astropy.py`. 🤷

ejeschke:
Let me give it one last round of testing today. But I think it is very close.

ejeschke:
@pllim, FFTM.  I think you are planning to do some more updates anyway in a future PR, so if additional issues come up we can address them there.

pllim:
Thanks! I am hoping to see if we can take advantage of photutils next. 😸

PR #882, opened 2020-08-27
TST: Verbose pytest and use capsys

pllim:
This PR makes pytest output more verbose, so we see exactly which test fails when failure happen. It also shows any screen output.

This PR also fixes the following screen output in the test log in the verbose mode set above (it currently does not fail the test but should be captured and tested anyway):

```
ginga/misc/tests/test_Callback.py::TestCallbacks::test_make_callback_raises_no_exception Error making callback 'test_name': test_callback_function() takes 0 positional arguments but 1 was given
Traceback:
  File ".../ginga/misc/Callback.py", line 156, in _do_callbacks
    res = method(*cb_args, **cb_kwargs)

PASSED
ginga/misc/tests/test_Callback.py::TestCallbacks::test_make_callback_raises_no_exception_completes_all_callbacks Error making callback 'test_name': test_callback_function() takes 0 positional arguments but 1 was given
Traceback:
  File ".../ginga/misc/Callback.py", line 156, in _do_callbacks
    res = method(*cb_args, **cb_kwargs)

```

### DIFF
--- a/.github/workflows/ci_workflows.yml
+++ b/.github/workflows/ci_workflows.yml
@@ -53,7 +53,7 @@ jobs:
     - name: Install Qt
       uses: jurplel/install-qt-action@8ffcd60
     - name: Run tests
-      run: pytest --pyargs ginga doc
+      run: pytest --pyargs ginga doc -sv
 
   older_deps_tests:
     runs-on: ubuntu-16.04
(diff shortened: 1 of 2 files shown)


ejeschke:
So `capsys` is just a magical parameter to a test function?

pllim:
It is provided by pytest. 😸 https://docs.pytest.org/en/stable/capture.html

PR #887, opened 2020-08-27
Change FWHM plot colors, add to colors.py functionality

### LINKED ISSUE #883: FWHM plot: Change X and Y colors?
https://github.com/ejeschke/ginga/blob/bdcc46004c0e7710cc75392b89f7a2356bbd9720/ginga/util/plots.py#L468-L476

Blue and green don't have good contrast, and probably not colorblind-friendly for some as well. Currently, as seen in `Pick`:

pick_fwhm_plot

pllim:
Fix #883 . What do you think? Derived loosely from https://colorbrewer2.org/#type=diverging&scheme=RdBu&n=3

I don't think this needs a change log but let me know if you want one. Thanks.

**Updated plots**

ginga_pick_fwhm

ginga_pick_radial

### DIFF
--- a/ginga/colors.py
+++ b/ginga/colors.py
@@ -1,3 +1,4 @@
+"""Module to handle colors supported by Ginga."""
 #
 # colors.py -- color definitions
 #
@@ -8,7 +9,8 @@
 import re
 import collections
 
-import numpy as np
+__all__ = ['recalc_color_list', 'lookup_color', 'resolve_color', 'add_color',
+           'remove_color', 'get_colors', 'scan_rgbtxt', 'scan_rgbtxt_buf']
 
 color_dict = {
  'aliceblue': (0.9411764705882353, 0.9725490196078431, 1.0),  # noqa
@@ -751,6 +753,7 @@ color_list = []
 
 
 def recalc_color_list():
+    """Recalculate ``ginga.colors.color_list``."""
     global color_list
     color_list = list(color_dict.keys())
     color_list.sort()
@@ -763,8 +766,8 @@ def lookup_color(name, format='tuple'):
         Color name (e.g., ``'red'``) or hash (e.g., ``'#ff0000'``).
         Color name is case-sensitive.
 
-    format : {'tuple', 'hash', 'name'}
-        Desired output to be an RGB tuple, hash, or name.
+    format : {'tuple', 'hash'}
+        Desired output to be an RGB tuple or hash.
 
     Returns
     -------
@@ -780,7 +783,7 @@ def lookup_color(name, format='tuple'):
         Invalid format.
 
     """
-    supported_formats = ('tuple', 'hash', 'name')
+    supported_formats = ('tuple', 'hash')
     if format not in supported_formats:
         raise ValueError(f'format needs to be one of {supported_formats}')
 
@@ -792,22 +795,13 @@ def lookup_color(name, format='tuple'):
         rgb = (int(name[:2], 16) / 255.0,
                int(name[2:4], 16) / 255.0,
                int(name[4:6], 16) / 255.0)
-        if format == 'tuple':
-            return rgb
 
-        color = _reverse_lookup(rgb)
-        if color is None:
-            raise KeyError(f'No color name for resolved RGB {rgb}')
-
-        return color
+        return rgb
 
     # must be a name
 
     elif name not in color_dict:
         raise KeyError(f'{name} does not exist in color_dict')
-
-    if format == 'name':
-        return name  # no-op
 
     rgb = color_dict[name]
     if format == 'tuple':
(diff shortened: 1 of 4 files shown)


ejeschke:
@pllim, how about [this one](https://colorbrewer2.org/#type=diverging&scheme=PRGn&n=3)? It also says colorblind-friendly.

pllim:
@ejeschke , that works too. Can you please tell me the proper names in your color DB that correspond to what that site is showing? Thanks.

pllim:
I added functionality to `colors.py` to see if I can reverse-lookup the color palette you suggested, and changed the colors. I updated the screenshot above.

Turns out, I could simply look up the colors by eye at https://matplotlib.org/3.1.0/gallery/color/named_colors.html . 🤦  I kept my code addition anyway, in case it is useful.

What do you think? Please review. Thanks!

ejeschke:
@pllim, you can specify colors for Ginga canvas objects by HTML hash triplets... e.g. `'#af8dc3'` and `'#7fbf7b'`

Also, the area under the curve that is shaded, to show the "width" of the FWHM, I think that was matched to the line colors in X and Y if I'm not mistaken.  Has that been updated as well? Hard to tell from this particular screenshot...

pllim:
> you can specify colors for Ginga canvas objects by HTML hash triplets

Not sure if Matplotlib will honor that.

> the area under the curve that is shaded

Probably not updated yet. Where is that code?

pllim:
Re: under the curve -- Oh wait, if it is `color3`, then should be updated already?

ejeschke:
> Not sure if Matplotlib will honor that.

I think matplotlib also handles hex triplets.

pllim:
> I think matplotlib also handles hex triplets.

Yes, I noticed too late (after I did all that code change prior). I put the hex directly in the calls now. Please review. Thanks!

ginga/util/plots.py
```diff
@@ -467,12 +467,12 @@ def plot_fwhm(self, x, y, radius, image, cutout_data=None,
 
             self.logger.debug("xarr=%s" % (str(xarr)))
             fwhm_x = self._plot_fwhm_axis(xarr, iqcalc, skybg,
-                                          'blue', 'blue', 'skyblue',
+                                          '#af8dc3', '#af8dc3', 'thistle',
```
ejeschke, line 470: Try `#7570b3`

ginga/util/plots.py
```diff
@@ -467,12 +467,12 @@ def plot_fwhm(self, x, y, radius, image, cutout_data=None,
...
                                           fwhm_method=fwhm_method)
 
             self.logger.debug("yarr=%s" % (str(yarr)))
             fwhm_y = self._plot_fwhm_axis(yarr, iqcalc, skybg,
-                                          'green', 'green', 'seagreen',
+                                          '#7fbf7b', '#7fbf7b', 'aquamarine',
```
ejeschke, line 475: Try `#1b9e77`

These are also suggested by the color brewer as safe for color blind. I think it is bolder and easier to read. What do you think?

Once we decide on the color, we should also modify [here](https://github.com/ejeschke/ginga/blob/master/ginga/util/plots.py#L400) with the same substitutions for `blue` and `green`. This will fix a colorblind issue for the `Radial` plot as well.

ginga/colors.py
```diff
@@ -755,27 +757,95 @@ def recalc_color_list():
...
+    if format == 'name':
+        return name  # no-op
+
+    rgb = color_dict[name]
+    if format == 'tuple':
+        return rgb
+
+    return f'#{int(rgb[0] * 255):02x}{int(rgb[1] * 255):02x}{int(rgb[2] * 255):02x}'
+
+
```
ejeschke, line 818: @pllim, do you think the reverse color lookup is necessary?  I feel like since we can specify hex triplets to both ginga and matplotlib there is no real reason for it.  The color list in `colors` is obviously a small subset of the 24M possible colors.  It is not like we need to find the closest, non-perfect match when you can specify an exact triplet.  If it is easier to read the code with names, can do something like
```python
green, purple = '#1b9e77', '#7570b3'
```
inside a specific module (e.g. `ginga.util.plots`) and then use the variables in the code.

I don't see much of a use case for it, but if you really like it I'm ok to leave it in...

pllim: Heh heh. I guess not. It was a nice programming exercise regardless... 😬

pllim:
@ejeschke , I think I have addressed all your comments. I have also updated the screenshots above. And as a bonus, I threw in API doc for `colors`: https://ginga--887.org.readthedocs.build/en/887/ref_api.html#module-ginga.colors

ejeschke:
Thanks, @pllim! Looks good! FFTM.

PR #888, opened 2020-08-28
Remove support for matplotlib < 2.0

ejeschke:
This removes support for matplotlib versions < 2.0. 

Since the "current" version of matplotlib seems to be 3.x I think this is reasonable.

### DIFF
--- a/doc/WhatsNew.rst
+++ b/doc/WhatsNew.rst
@@ -10,6 +10,7 @@ Ver 3.2.0 (unreleased)
 - Added an option to make a copy of existing shape in Drawing plugin
 - Added an option to make a copy of existing cut in Cuts plugin
 - Fixed a bug where certain plots were not cleared in Pick plugin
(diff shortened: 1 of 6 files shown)


ejeschke:
Need to add a note to `WhatsNew`...will do so before a merge.

ejeschke:
Good catch!

ejeschke:
@pllim, what do you think should be the minimum version of `matplotlib` going forward?

pllim:
astropy has `>=2.1`, so maybe we'll just copy astropy?

https://github.com/astropy/astropy/blob/679a74a463794829d1dcd02570708c21695397aa/setup.cfg#L72

ejeschke:
> astropy has >=2.1, so maybe we'll just copy astropy?

Sounds fine.

ejeschke:
Rebased

ejeschke:
@pllim, ok to merge? Updated `setup.cfg` and rebased.

PR #889, opened 2020-08-28
Redraw plot after a clear operation

### LINKED ISSUE #884: Pick plugin: Not obvious when FWHM fitting failed
In the `Pick` plugin, currently user can switch between Gaussian and Moffat (and with #880, Lorentz too). When I switch "FWHM fitting" algorithm choice in Settings and click Redo Pick, it is hard to tell if the fit was actually successful or not without looking at the terminal output.

On the terminal, I would see things like `RuntimeWarning` or (with #880 and using `astropy.modeling`) `The fit may be unsuccessful`. However, there is no indication of any problem back at the GUI. At worst, the numbers don't update and one might mistake result from a previous selection as result for the new algorithm.

Given the already crowded GUI, perhaps the simplest solution is to clear the plot before attempting a re-fit. Then user will know something is wrong if a new plot does not appear. What do you think, @ejeschke ?

ejeschke:
Fix #884 

Fixes an issue where a plot is not redrawn automatically when calling `clear()` on it.

### DIFF
--- a/ginga/util/plots.py
+++ b/ginga/util/plots.py
@@ -94,6 +94,7 @@ class Plot(Callback.Callbacks):
         self.ax.cla()
         self.xdata = []
         self.ydata = []
+        self.draw()
 
     def draw(self):
         self.fig.canvas.draw()
--- a/doc/WhatsNew.rst
+++ b/doc/WhatsNew.rst
@@ -9,6 +9,7 @@ Ver 3.2.0 (unreleased)
 - Canvas shapes can now be copied
 - Added an option to make a copy of existing shape in Drawing plugin
 - Added an option to make a copy of existing cut in Cuts plugin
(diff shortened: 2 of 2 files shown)


ejeschke:
Not sure this warrants a mention in the `WhatsNew`.

ginga/util/plots.py
```diff
@@ -94,6 +94,7 @@ def clear(self):
         self.ax.cla()
         self.xdata = []
         self.ydata = []
+        self.draw()
```
pllim, line 97: I thought I tried this... but I'll test this PR to be sure.

pllim:
> Not sure this warrants a mention

Usually a bug warrants a mention but since no one else has complained about it, I'll leave that to your good judgement. 😸

ejeschke:
I added a comment to the `WhatsNew`

PR #891, opened 2020-08-29
MNT: Remove deprecated mpl usage

### LINKED ISSUE #890: MatplotlibDeprecationWarning: The global colormaps dictionary is no longer considered public API
This deprecation warning appears from [this line](https://github.com/ejeschke/ginga/blob/master/ginga/cmap.py#L13317) in `ginga.cmap`.  It looks like `matplotlib` is trying to [deprecate accessing the global colormaps dictionary](https://github.com/matplotlib/matplotlib/blob/master/lib/matplotlib/cm.py#L75-L118).  

So far I have not found a replacement API for getting the same information using a public API.

pllim:
Seems to work with matplotlib 2.1, so I don't think we need any version check.

Fix #890, fix spacetelescope/imexam#214 . Thanks to @jchavez1992 for reporting.

Caused by matplotlib/matplotlib#16991

### DIFF
--- a/ginga/cmap.py
+++ b/ginga/cmap.py
@@ -13306,6 +13306,7 @@ def add_matplotlib_cmap(cm, name=None):
 def add_matplotlib_cmaps(fail_on_import_error=True):
     """Add all matplotlib colormaps."""
     try:
+        import matplotlib.pyplot as plt
         from matplotlib import cm as _cm
         from matplotlib.cbook import mplDeprecation
     except ImportError:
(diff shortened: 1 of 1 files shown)


ginga/cmap.py
```diff
@@ -13314,7 +13315,8 @@ def add_matplotlib_cmaps(fail_on_import_error=True):
         # silently fail
         return
 
-    for name in _cm.cmap_d:
+    # NOTE: Update if matplotlib has new public API for this.
```
pllim, line 13318: Is this comment sufficient, @ejeschke ?

ejeschke:
@pllim, FFTM.

PR #896, opened 2020-09-04
Incorporate photutils into iqcalc_astropy and Pick

### LINKED ISSUE #895: ENH: Use photutils for source finding in Pick
This follows #880 as possible future enhancement.

pllim:
Fix #895 

**TODO**

- [x] Onto "EVALUATION ON A FIELD" section in `iqcalc.py` next.
   * Add/improve docstring
   * Add test in `test_iqcalc.py`
   * Implement alternative using photutils in `iqcalc_astropy.py` and make sure child test class runs it
- [x] See if anything needs changing in `Pick.py` and make sure it works with both "native" and "astropy" -- See https://github.com/ejeschke/ginga/pull/896#issuecomment-692832388
- [x] Make sure CI passes
- [x] Check rendered API doc

### DIFF
--- a/doc/WhatsNew.rst
+++ b/doc/WhatsNew.rst
@@ -9,7 +9,8 @@ Ver 3.2.0 (unreleased)
 - Canvas shapes can now be copied
 - Added an option to make a copy of existing shape in Drawing plugin
 - Added an option to make a copy of existing cut in Cuts plugin
-- Added new iqcalc_astropy module to handle FWHM fitting using astropy
+- Added new iqcalc_astropy module to handle FWHM fitting and source finding
+  using astropy and photutils
 - Added new calc_fwhm_lib configuration item to let Pick switch between
   iqcalc and iqcalc_astropy
 - Fixed a bug where certain plots were not cleared in Pick plugin
--- a/.github/workflows/ci_workflows.yml
+++ b/.github/workflows/ci_workflows.yml
@@ -52,7 +52,7 @@ jobs:
     - name: Install and build
       run: |
         python -m pip install --upgrade pip setuptools
-        python -m pip install PyQt5 scipy
+        python -m pip install PyQt5 scipy photutils
         python -m pip install -e .[test]
     - name: Install Qt
       uses: jurplel/install-qt-action@8ffcd60
--- a/ginga/tests/test_iqcalc.py
+++ b/ginga/tests/test_iqcalc.py
@@ -1,19 +1,13 @@
 import logging
+
 import numpy as np
 import pytest
+from numpy.testing import assert_allclose, assert_array_equal
 
+from ginga.AstroImage import AstroImage
 from ginga.util import iqcalc, iqcalc_astropy
-try:
-    from scipy import optimize  # noqa
-    HAS_SCIPY = True
-except ImportError:
-    HAS_SCIPY = False
-
-try:
-    from scipy import optimize  # noqa
-    HAS_SCIPY = True
-except ImportError:
-    HAS_SCIPY = False
+from ginga.util.iqcalc import have_scipy  # noqa
+from ginga.util.iqcalc_astropy import have_photutils  # noqa
 
 
 @pytest.mark.parametrize(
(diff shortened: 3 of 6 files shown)


doc/WhatsNew.rst
```diff
@@ -9,7 +9,8 @@ Ver 3.2.0 (unreleased)
 - Canvas shapes can now be copied
 - Added an option to make a copy of existing shape in Drawing plugin
 - Added an option to make a copy of existing cut in Cuts plugin
-- Added new iqcalc_astropy module to handle FWHM fitting using astropy
+- Added new iqcalc_astropy module to handle FWHM fitting and source finding
+  using astropy
```
pllim, line 13: Note to self: Fix this
```suggestion
  using astropy and photutils
```

pllim:
Update: Photometry algorithm is different enough from what I can find in photutils that I can't really integrate that one this round, but I did add a bunch of docstrings and tests. 😬

pllim:
Based on the test results, I don't think this will break `Pick` but still would be nice to test it out, @ejeschke .

pllim:
A simple test using https://stsci.box.com/s/c1rs5hqbv4mc6qlmw6fa9qedcina9p07 produced the following results when comparing `Pick` using `iqcalc.py` and `iqcalc_astropy.py`. Results reflect changes done in #880 *and* this PR:

Native | Astropy | Diff
:-------------------------:|:-------------------------:|:-------------------------:
ginga_native_pick_tab1 | ginga_astropy_pick_tab1 |<b>Background</b><br/>0.178 -> 0.192<br/><br/><b>Sky level</b><br/>40.187 -> 40.202<br/><br/><b>Brightness</b><br/>17.494 -> 17.486<br/><br/><b>FWHM X</b><br/>4.891 -> 4.883<br/><br/><b>FWHM Y</b><br/>8.258 -> 8.248<br/><br/><b>FWHM</b><br/>6.786 -> 6.777<br/><br/><b>Star size</b><br/>3.944 -> 3.939
ginga_native_pick_tab2 | ginga_astropy_pick_tab2 | None
ginga_native_pick_tab3 | ginga_astropy_pick_tab3 | None, <b>but</b> reported FWHM values do not match those in Readout above!<br/><br/><b>Above</b><br/>FWHM X: 4.9<br/>FWHM Y: 8.3<br/><br/><b>Here:</b><br/>FWHM X: 3.74<br/>FWHM Y: 5.09
ginga_native_pick_tab4 | ginga_astropy_pick_tab4 | None

p.s. `Cuts` tab is empty for both.

ejeschke:
@pllim, are you sure the setting got changed when you did the comparison the second time? I would expect the answers to be close but not exact, similar to the first result you got. Actually, that one made me pretty comfortable that we had replicated things properly, but to have them be identical seems odd.  I will have a test go at it ASAP.

pllim:
> are you sure the setting got changed when you did the comparison the second time?

I checked that "Lorentz" is an option in one but not the other. Not fool-proof though. Let me purposely fail something in `iqcalc_astropy.py` and see... I will update *this* comment when that is done.

pllim:
p.s. There is a chance that `astropy.modeling.fitting.LevMarLSQFitter()` is slightly non-deterministic. Either that or I am going crazy.

pllim:
p.p.s. @karllark suggested that the results are different because maybe I didn't click on the exact same pixel as the last run, affected the actual data in the cutout array, and hence slightly different best fit values.

ejeschke:
> p.p.s. @karllark suggested that the results are different because maybe I didn't click on the exact same pixel as the last run, affected the actual data in the cutout array, and hence slightly different best fit values.

Except that if the cutout was more or less the same, the peak should have been detected similarly and then the final fit done on the pixel identified by `evaluate_peaks()`. In that sense, this should be pretty deterministic.  If the modeled fit is not deterministic, then all bets are off...

ejeschke:
> p.s. Cuts tab is empty for both.

This is normal unless you have "Quick mode" checked.

pllim:
Re: https://github.com/ejeschke/ginga/pull/896#issuecomment-692989506

I am not sure. Could be `photutils` instead of `astropy.modeling` too. I have to try to find time to isolate this problem.

Do you see the same non-deterministic behavior when trying to measure the same object in different Ginga sessions?

ejeschke:
> Do you see the same non-deterministic behavior when trying to measure the same object in different Ginga sessions?

If I define a box region around an object, and there are no other star-like peaks in the box, I get the same result no matter where in the box the object is, unless the object gets really close to an edge (because then it might fail the edge setting in the `Settings` tab). With the "legacy" `iqcalc` it should be reasonably deterministic.

ejeschke:
Note: above comment is for the legacy `iqcalc`. I have to try the PR...will report back.

ejeschke:
With the asotry `iqcalc` it seems to behave deterministically for me, as far as moving the object around inside a box.  That's good news.

pllim:
Today, in 3 different Ginga sessions, `iqcalc_astropy` reported the exact same values as the "classic" `iqcalc` using `astropy` 4.1rc2.

I also tried with `astropy` 4.2.dev687+g214a599c1. Same result.

I am stumped! If you don't find anything wrong with this, I propose we merge it as is. Thanks for reviewing!

ejeschke:
Using the "Record Picks automatically" option in the `Report` tab makes it pretty easy to check repeated results against each other. You can repeatedly press the "Redo Pick" button in the `Settings` tab to repeat the exact same calculation with the exact same box selection.  Using this I did not see any variability in the `astropy` version `iqcalc` results.  By moving the box a little (instead of pressing "Redo Pick"), I can see that the values change slightly for each successive calculation, but this may be expected as the cutout is a little different each time.

ejeschke:
> Today, in 3 different Ginga sessions, iqcalc_astropy reported the exact same values as the "classic" iqcalc using astropy 4.1rc2.

> I also tried with astropy 4.2.dev687+g214a599c1. Same result.

> I am stumped! If you don't find anything wrong with this, I propose we merge it as is. Thanks for reviewing!

I find this very odd, and would like to understand why, but if you are satisfied with it I am not opposed to merging. FFTM.

pllim:
> I find this very odd, and would like to understand why

Me too, but I am not sure how to proceed. Any ideas?

ejeschke:
> Me too, but I am not sure how to proceed. Any ideas?

@pllim, is there a testing file you can share with me on Box?  I get different answers when I run with the different `iqcalc`s with the data files I've tried.

pllim:
@ejeschke , file is at https://stsci.box.com/s/c1rs5hqbv4mc6qlmw6fa9qedcina9p07 ; you should have permission already. For the coordinates, please see https://github.com/ejeschke/ginga/pull/896#issuecomment-692832388 . Thanks!

ejeschke:
@pllim, are you using the `astropy` WCS or the `astropy_ape14` WCS?  Using that FITS file (s12_ch4_v0.30_sci[PRIMARY, 1]) and the `Settings` plugin to set the exact RA/DEC shown in the `Pick` screenshots, it centered me on the large extended source at the top of the image, not a point source.  To encompass that I'd need to use a much large box radius on the order of 200x200 px.  Confused. :confused:

pllim:
Hmm. Looks like I have `WCSpkg = 'astropy'` set in my `~/.ginga/general.cfg`. I didn't even think about testing it with `astropy_ape14`!

ejeschke:
In any case, testing with this image I tend to get slight variations in calculations depending on the box cut-out.  I think this is not too unusual.  I tried picking the exact same pixel of a point source using the two different `iqcalc`s and got two different results. 

I'm ok to merge and follow up later on issues found "down the road". I expect there may be more work on this plugin.

ejeschke:
> I'm ok to merge and follow up later on issues found "down the road". I expect there may be more work on this plugin.

I would like to know about the coordinates! That is very odd!

ejeschke:
@pllim, did you pick a point source *near* that large extended source?

ejeschke:
> Hmm. Looks like I have WCSpkg = 'astropy' set in my ~/.ginga/general.cfg. I didn't even think about testing it with astropy_ape14!

I am testing with `astropy`, not `astropy_ape14`.

pllim:
No, I actually pick the extended source on purpose, because it is big and shiny. 😸

pllim:
I tried it again on a star this time, instead of big shiny thing, using the same image as provided above. I set the preferences on `general.cfg` for WCS (between `astropy` and `astropy_ape14`) and `plugin_Pick.cfg` (between `classic` and `astropy`).

Only **astropy_ape14 WCS + classic Pick** has slight differences; the rest are exactly the same as far as GUI-visible results are concerned. Even the difference isn't concerning. But if I run the **astropy_ape14 WCS + classic Pick** again (not shown), it shows the same results as the rest. I guess that is acceptable given that I probably clicked on slightly different starting pixel? 🤷 

| | classic Pick | astropy Pick |
| --- | --- | --- |
| **astropy WCS** | ginga_pick_classic_astropy_wcs | ginga_pick_photutils_astropy_wcs |
| **astropy_ape14 WCS** | ginga_pick_classic_astropy_ape14_wcs | ginga_pick_photutils_astropy_ape14_wcs |

ejeschke:
> the rest are exactly the same as far as GUI-visible results are concerned

I understand--the GUI results are rounded to 3 decimal places or so.  If you use the "Automatically record Picks" feature, it shows the full values for fwhm X and Y.  Can you ascertain that these are different when using the old vs new iqcalc?

In any case, I think we are "splitting hairs" now! LOL

pllim:
Here are the results when I save the pick into a file. It does seem the "native" algorithm is more stable, but the one using `astropy.modeling` and `photutils` doesn't produce differences at the level that people care about either. 🤷 

What do you think, @ejeschke ?

| | WCS: astropy<br/>Pick: native | WCS: astropy<br/>Pick: astropy | WCS: astropy_ape14<br/>Pick: native | WCS: astropy_ape14<br/>Pick: astropy |
| --- | --- | --- | --- | --- |
| **x** | 1135.8739052317208 | 1135.8738308871 | 1135.8739052317208 | 1135.8739059488003 |
| **y** | 1218.7036197475268 | 1218.7036375439445 | 1218.7036197475268 | 1218.7036196699014 |
| **fwhm** | 3.488110405684747 | 3.484472387457708 | 3.488110405684747 | 3.4881128040594405 |
| **fwhm_x** | 3.3940852887810866 | 3.3906405660278764 | 3.3940852887810866 | 3.3940884059797356 |
| **fwhm_y** | 3.579666668416521 | 3.5758428642652356 | 3.579666668416521 | 3.5796683868873616 |
| **starsize** | 2.0921255871634665 | 2.0899450290921133 | 2.0921255871634665 | 2.0921270378643135 |
| **ellipse** | 0.9481567987117843 | 0.9482073722846839 | 0.9481567987117843 | 0.9481572143421381 |
| **background** | 0.0007347411301452667 | 0.0008672954863868654 | 0.0007347411301452667 | 0.0007347411301452667 |
| **skylevel** | 40.00077147818665 | 40.00091066026071 | 40.00077147818665 | 40.00077147818665 |
| **brightness** | 0.19423084421331266 | 0.19414176863493565 | 0.19423084421331266 | 0.19423078177673536 |
| **ra_deg** | 53.117970132803016 | 53.11797014681092 | 53.117970132803016 | 53.11797013266791 |
| **dec_deg** | -27.802054981084684 | -27.80205497811911 | -27.802054981084684 | -27.802054981097626 |

ejeschke:
I agree. At least we can see that there is some expected (small) differences. FFTM when ready.

pllim:
Thanks for the review!

PR #899, opened 2020-09-14
Add bicubic and bilinear interpolation methods to OpenGL backend

ejeschke:
Adds bicubic and bilinear interpolation methods to the OpenGL backend, which up till now only supported "nearest". No change to non-opengl backends, which already have these options.


ejeschke:
@pllim, care to review? If not, no worries.  This is just a small enhancement to the OpenGL renderer.

pllim (commented):
Unfortunately, if I remember correctly, I think my Linux machine is not compatible with your OpenGL implementations (my graphics card too old or something).

If there is specific thing you want me to test on the Windows side, please let me know. Otherwise, I don't think anyone I support at STScI specifically uses OpenGL capability (my direct customers also don't use Windows for production on their side).

PR #903, opened 2020-09-23
MNT: Replace pep517 with python-build

pllim:
See pypa/pep517#82


ejeschke:
@pllim, the description is a bit unclear to me just from following the link.

Can you give me the 2 minute overview?

pllim:
PyPA moved some CLI out of `pep517` package into `python-build`. Just future-proofing here before they start the deprecation period. 😸

ejeschke (approved):
Looks ok to me, as far as I understand it.  FFTM, @pllim

PR #905, opened 2020-10-07
PixTable: Add box sum, fix index bug

### LINKED ISSUE #904: Adding total brightness inside the box in PixTable
Hi all, 
This is a request to add the total sum of the pixel values as one of the outputs in PixTable. This would be in addition to the existing output, Min, Max, Avg and Rms. I think the total surface brightness within the NxN box would be a very useful feature for a quick assessment of a PSF brightens when trying to identify best observing strategies. 

There may be other ways to obtain the total brightness inside a NxN box inside Ginga. For example, I am wondering if the total flux of a 3x3 box can be obtained via Pick by changing the Box size, however, currently the lowest setting allowed seems to be 5. 

However, having this total brightens functionality - within a NxN box - seems to be appropriate for the PixTable plugging. 

Thanks!

pllim:
* Fix bug, where indices are flipped
* Add box sum, fix #904 
* Add median (as per review comment)
* Move statistics display out
* Use f-string for some string formatting
- [x] Saw the comment about `np.vectorize`, ~lemme see...~ but "notes" section for [numpy.vectorize](https://numpy.org/doc/stable/reference/generated/numpy.vectorize.html) says, "The `vectorize` function is provided primarily for convenience, not for performance. The implementation is essentially a for loop." Therefore, implementing it will make the code less readable, while providing little return. I also have not noticed any performance issue myself, so I went ahead and removed the comment from code. **Update:** Implemented it anyway after review comment.

cc @obi-wan76


ejeschke:
Looks good, @pllim.  Thought I'd already fixed that indexing issue once.  It's also slightly confusing if the image is flipped in X, Y or swapped axes.  Not saying that the `PixTable` display should follow that, but it could.  Should be a separate PR, but something to think about.

I wonder whether we shouldn't relabel "Avg" to "Mean" and also add the "Median".  That is also a really useful number for this plugin (think sky level) and would then fill up the labels to an even number.

Also, what do you think about having six labels in two rows, instead of three rows?  I think it would fit fine.  But I'm not particular about this.

ejeschke:
Re: `np.vectorize`: any call to a C function that can perform a nested (double) loop and calling back into a Python function will be faster than two `for` loops in Python calling a Python function. I don't think we need to prematurely optimize, and when you consider the speed of just a single plugin in isolation it doesn't look like much, but you have to remember that for any plugin that is active (and even some that aren't) code gets run when you move the cursor around, and it is the *sum total* of all the callback times that we should keep an eye on.  If someone has `Info`, `Zoom`, `PixTable`, `Cursor` all open there are 4 different callbacks being invoked when you move from one pixel to another. I'm probably forgetting some too.

pllim:
@ejeschke , I think I applied your review comments. Please have another look.

> if the image is flipped in X, Y or swapped axes.

Yes, I agree that is a separate PR. If the transform info is stored somewhere, should not be hard to apply it to the PixTable values.

pllim:
In regards to performance improvement using `np.vectorize`, I am still not convinced. Consider this:

```python
import numpy as np

data = np.arange(12).reshape(4, 3)
fmt_cell = '> 8.4g'

class Text:
    def __init__(self):
        text = ''

x = np.array([[Text(), Text(), Text()],
              [Text(), Text(), Text()],
              [Text(), Text(), Text()],
              [Text(), Text(), Text()]])

def _vecfunc(val, out):
    if not np.isscalar(val):
        val = np.average(val)
    out.text = f'{val:{fmt_cell}}'

func = np.vectorize(_vecfunc)

def oldfunc(data, height, width, out):
    for i in range(height):
        for j in range(width):
            val = data[i][j]
            if not np.isscalar(val):
                val = np.average(val)
            out[i][j].text = f'{val:{fmt_cell}}'
```
```python
In [x]: %timeit func(data, x)                                                                       
22.9 µs ± 122 ns per loop (mean ± std. dev. of 7 runs, 10000 loops each)

In [x]: %timeit oldfunc(data, data.shape[0], data.shape[1], x)
18 µs ± 52.9 ns per loop (mean ± std. dev. of 7 runs, 100000 loops each)
```

ejeschke:
How about with 9x9?  I imagine that at some point there is a break-even time and then the vectorize becomes faster.  Perhaps with the sizes of the cutouts we are dealing with we never quite reach that threshold.

ejeschke:
@pllim, changes look good.  One last thing: can you keep the labels from constantly adjusting position depending on the length of the value digit length (based on the data you are mousing over)?  I think you can use a trick like I did with `maxdigits` and `fmt_cell` a little bit higher in the code that keeps the numbers in the table lined up.  LMK if you need more clarification.

pllim:
9 x 9
```python
In [x]: %timeit func(data, x)                                                                      
74 µs ± 174 ns per loop (mean ± std. dev. of 7 runs, 10000 loops each)

In [x]: %timeit oldfunc(data, data.shape[0], data.shape[1], x)                                     
106 µs ± 2.79 µs per loop (mean ± std. dev. of 7 runs, 10000 loops each)
```

1000 x 1000
```python
In [x]: %timeit func(data, x)                                                                      
875 ms ± 2.91 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)

In [x]: %timeit oldfunc(data, data.shape[0], data.shape[1], x)                                     
1.43 s ± 2.58 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)
```

Perhaps the vectorize has slight performance improvement for larger arrays but not by much.

pllim:
> can you keep the labels from constantly adjusting position depending on the length of the value digit length (based on the data you are mousing over)?

I thought I did. How is it constantly adjusting for you? I am not sure I am seeing what you are saying.

ejeschke:
It depends on the data.  It seems to keep triggering for me when the "Sum" changes between a small integer (e.g. 4 digits) and a larger number that is shown with the exponential notation.

Do you have any `uint16` arrays with some large values? It should show it.

pllim:
Sorry, @ejeschke , I seem to have a hard time understanding this problem. What exactly is jumping and how do you envision it working properly? I am currently using the same cell formatting that you had before (`> 8.4g`).

Screenshot 2020-10-09 102318

Screenshot 2020-10-09 102340

ejeschke:
> What exactly is jumping and how do you envision it working properly?
@pllim, I uploaded a [video to the shared Box folder](https://stsci.box.com/s/dqfwqib4ezesgndq68si1izobhw3tymm).

Hopefully you can play it.  It shows the issue.

ejeschke:
The problem seems to stem from the fact that the length of the label in the sum changes the number of digits shown, depending on the value, rather than padding the end with spaces or zeros, so that the field width is constant.

pllim:
I wonder if changing from label to uneditable text box would be better. I'll investigate on Monday. Thanks for the review!

ejeschke:
I think there should be a formatting option that pads with spaces so that the field is always the same length (in characters), no matter what.  I think that should solve the problem, unless trailing spaces are not treated the same in a label. In that case I would think that if a fixed-width font is used, it should be ok.

pllim:
@ejeschke , thanks for the video; I finally understand.

Looking at Qt widget implementation of `Label`, doesn't seem to be a way to fix the length. In `TextEntry`, there is a `set_length` method but it is a no-op.

pllim:
Maybe instead of making each of the element its own `Label`, I can instead implement one `Label` per line. Maybe that'll do.

ejeschke:
Do Qt labels not respect trailing spaces with a fixed width font?

ejeschke:
Could also go back to displaying them under the array numbers, but in two lines.

pllim:
@ejeschke , tried this (display under array values but in 2 lines) but the problem of label moving around still exists (and appears even worse):
```diff
diff --git a/ginga/rv/plugins/PixTable.py b/ginga/rv/plugins/PixTable.py
index 21a4cde6..af26245c 100644
--- a/ginga/rv/plugins/PixTable.py
+++ b/ginga/rv/plugins/PixTable.py
@@ -136,6 +136,7 @@ class PixTable(GingaPlugin.LocalPlugin):
         # For pixel table
         self.pixtbl_radius = 2
         self.txt_arr = None
+        self.sum_arr = None
         self.sizes = [1, 2, 3, 4]
         self.maxdigits = 9
         self.fmt_cell = f'> {self.maxdigits - 1}.{self.maxdigits // 2}g'
@@ -193,20 +194,6 @@ class PixTable(GingaPlugin.LocalPlugin):
 
         self._rebuild_table()
 
-        vbox2 = Widgets.VBox()
-
-        # Stats report panel
-        captions = (
-            ('Min:', 'label', 'Min', 'llabel',
-             'Mean:', 'label', 'Mean', 'llabel',
-             'Median:', 'label', 'Median', 'llabel'),
-            ('Max:', 'label', 'Max', 'llabel',
-             'RMS:', 'label', 'RMS', 'llabel',
-             'Sum:', 'label', 'Sum', 'llabel'))
-        w, b = Widgets.build_info(captions)
-        self.w.update(b)
-        vbox2.add_widget(w, stretch=0)
-
         btns = Widgets.HBox()
         btns.set_border_width(4)
         btns.set_spacing(4)
@@ -252,6 +239,7 @@ class PixTable(GingaPlugin.LocalPlugin):
         self.w.btn_delete_all = btn2
         btns.add_widget(Widgets.Label(''), stretch=1)
 
+        vbox2 = Widgets.VBox()
         vbox2.add_widget(btns, stretch=0)
 
         btns = Widgets.HBox()
@@ -434,13 +422,10 @@ class PixTable(GingaPlugin.LocalPlugin):
 
         ctr_txt = self.txt_arr[width // 2][height // 2]
 
-        # Report statistics
-        self.w.min.set_text(f'{minval:{fmt_cell}}')
-        self.w.max.set_text(f'{maxval:{fmt_cell}}')
-        self.w.sum.set_text(f'{sumval:{fmt_cell}}')
-        self.w.mean.set_text(f'{avgval:{fmt_cell}}')
-        self.w.rms.set_text(f'{rmsval:{fmt_cell}}')
-        self.w.median.set_text(f'{medianval:{fmt_cell}}')
+        # Append statistics lines
+        fmt_s = '> 10.4g'
+        self.sum_arr[0].text = f'Min: {minval:{fmt_s}}  Mean: {avgval:{fmt_s}}  Median: {medianval:{fmt_s}}'
+        self.sum_arr[1].text = f'Max: {maxval:{fmt_s}}  RMS: {rmsval:{fmt_s}}   Sum: {sumval:{fmt_s}}'
 
         # update the pixtable
         self.pixview.panset_xy(ctr_txt.x, ctr_txt.y)
@@ -541,13 +526,29 @@ class PixTable(GingaPlugin.LocalPlugin):
 
         self.txt_arr = np.array(rows)
 
+        # add summary row(s)
+        x = (font_wd + 2) + 4
+        y += font_ht + 20
+        y2 = y + font_ht + 2
+        dx, dy = crdmap.to_data((x, y))
+        dx2, dy2 = crdmap.to_data((x, y2))
+        s1 = Text(dx, dy, text='', font=self.font,
+                  color=color, fontsize=self.fontsize,
+                  coord='data')
+        s2 = Text(dx2, dy2, text='', font=self.font,
+                  color=color, fontsize=self.fontsize,
+                  coord='data')
+        objs.append(s1)
+        objs.append(s2)
+        self.sum_arr = np.array([s1, s2])
+
         # add all of the text objects to the canvas as one large
         # compound object
         CompoundObject = canvas.get_draw_class('compoundobject')
         canvas.add(CompoundObject(*objs), redraw=False)
 
         # set limits for scrolling
-        self.pixview.set_limits(((0, 0), (max_x, y)), coord='window')
+        self.pixview.set_limits(((0, 0), (max_x, y2)), coord='window')
 
     def set_cutout_size_cb(self, w, val):
         index = w.get_index()
```

As you can see, I also tried changing `> 8.4g` to `> 10.4g` but didn't really improve anything.

At this point, I think if you really want the label not to move, I think the only way is to arrange the stats in a `VBox` (6 rows) instead of 2 rows. Ideas/advice welcome!

ejeschke:
> Ideas/advice welcome!

Let me take a quick stab at it with the current PR.

pllim:
Thanks, @ejeschke ! I also thought about maybe 2 `VBox` side by side in a `HBox` but haven't tried it. The `VBox` might still move around unless there is a way to tell `HBox` to give each `VBox` 50% of total width.

ejeschke:
@pllim, can I push some commits on this PR?  Investigating this took me down a rabbit hole where I found a couple of surprising bugs!

pllim:
Feel free!

ejeschke:
ok, I rebased on top of the dep fix and a couple other fixes, one of which is necessary for getting a proper fixed font.

Give this a whirl, @pllim, and let me know what you think.

ejeschke:
I feel that if we want to transition this plugin to using a widget table, it ought to include both the pixel grid and the summary items.  Lacking a proper table widget, we can probably make do with a `TreeView` or a `GridBox` full of `Label`.  But I think that enhancement (?) should be a separate PR, maybe for release after 3.2.

pllim:
I agree with enhancement being a separate PR. I'll test this next week. Thank you very much for your help!

pllim:
The words don't move around anymore. But I noticed that:

* The text is now not centered, but rather anchored to top left of the display box.
* When I change any of the setting in drop down box (e.g., font size), the text disappears. It does reappear when I mouse over the display again, but that is not how it is currently behaving without this patch.

p.s. Sorry for the delay in my reply.

pllim:
Forgot to attach my screenshot, so here it is.

Screenshot 2020-10-30 160627

ejeschke:
> The text is now not centered, but rather anchored to top left of the display box.

Yes, this is due to a change from data to canvas coordinates for the plotting. I think it actually works better, because it makes efficient use of the space and you can see exactly how wide you will need to enlarge the right panel to see all the numbers, e.g. in a 9x9 table.  We can reduce the vertical space initially assigned to the widget if you think it is excessive. Or we just wait and it will resolve itself when moving to a `TreeView` widget in the next PR.

>  When I change any of the setting in drop down box (e.g., font size), the text disappears. It does reappear when I mouse over the display again, but that is not how it is currently behaving without this patch.

Fixed. Please do a pull to check it out.

> p.s. Sorry for the delay in my reply.

Absolutely no worries, @pllim. :smile:

pllim (commented):
Works great now, @ejeschke . Thank you so much! I cannot approve my own PR. 😹 

Feel free to merge.

pllim:
Actually, might need to update the screenshot. Do you want to do it or do you want me to do it? I might not get to it until next week. If you do it, can use Photoshop or Gimp to scale it down a bit and then squash out the old screenshot from commit history here.

ejeschke:
How big should the screenshot be?

pllim:
> How big should the screenshot be?

I scaled it down to height of 400 pixel, so the one I took was 538x400 and about 60 KB. As long as it appears fine in the rendered doc, should be okay; no need to be super hi-res IMHO.

ejeschke:
That seems awfully small vertically for a right-side plugin.  The [current docs](https://ginga.readthedocs.io/en/stable/_images/pixtable-plugin.png) show a much bigger image, as it includes the main image window.  Am I missing something?

pllim:
@ejeschke , I only included in the screenshot what I thought was necessary. If your opinion differs, feel free to take a different screenshot. My only request is to scale it down appropriately and squash out the older screenshot in the PR. Thanks!

ejeschke:
@pllim, updated screenshot ok?

pllim:
Screenshot looks good to me. Thanks!

I think if you click on the small arrow next to "Merge pull request" and then select "Squash and merge", you might save yourself some manual squashing on the PR. 😄

PR #915, opened 2020-12-29
MNT: Add Python 3.9 to metadata

### LINKED ISSUE #885: Add Python 3.9 support
Python 3.9 should be coming out soon. If everything goes well, all we need to do is add a job for it in CI and update `setup.cfg`. 🤞

pllim:
Fix #885 . Python 3.9 is already tested in CI as "3.x".


ejeschke (approved):
FFTM

PR #916, opened 2020-12-29
MNT: Add auto PR labeler

pllim:
This pull request adds auto PR labeler using https://github.com/actions/labeler in GitHub Actions. I tried to map some existing labels to associated files that seem obvious to me, but feel free to suggest changes if I got things wrong.

The motivation is to have more consistency in how the labels are applied to pull requests. Issues are not affected.

### DIFF
--- a/.github/labeler.yml
+++ b/.github/labeler.yml
@@ -1,0 +1,36 @@
+documentation:
+  - .readthedocs.yml
+  - CITATION
+  - doc/**/*
+  - ginga/doc/*
+  - licenses/*
+ - '*.md'
+ - '*.txt'
+  - any: ['*.rst', '!CHANGES.rst']
+
+maintenance:
+  - .bandit.yaml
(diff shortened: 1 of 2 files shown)


ejeschke:
Do you get the option to remove labels before posting the PR, or does it just automatically auto-label and then you have to remove after-the-fact?  Just curious.

pllim:
Re: https://github.com/ejeschke/ginga/pull/916#issuecomment-753844569 -- You cannot pre-emptively unlabel, though there is a bug where it will remove a label on PR open event when a label is manually added and that label is defined in the YAML but the file changed isn't in the rule for that label. So, when either case happens (add or remove), you can adjust the labels after-the-fact, as this Action only runs once for each PR at the time it is opened.

If this automation is not desired, feel free to close without merge.

Thanks!

ejeschke (approved):
Seems like this would be useful.

PR #917, opened 2020-12-29
MNT: Add issue and PR templates

pllim:
This pull request adds templates for new issues and pull requests. For issues, they are separated into feature request, bug report, and question categories. Also see https://docs.github.com/en/free-pro-team@latest/github/building-a-strong-community/using-templates-to-encourage-useful-issues-and-pull-requests

### DIFF
--- a/.github/ISSUE_TEMPLATE/bug_report.md
+++ b/.github/ISSUE_TEMPLATE/bug_report.md
@@ -1,0 +1,61 @@
+---
+name: Bug report
+about: Create a report describing unexpected or incorrect behavior in ginga.
+labels: 'bug'
+---
+
+<!-- This comments are hidden when you submit the issue,
+so you do not need to remove them! -->
+
+<!-- Please be sure to check out our code of conduct,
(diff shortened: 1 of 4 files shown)


ejeschke:
Is there a way to try it out from the PR?

pllim:
Unfortunately, I don't think so. But I have deployed similar templates elsewhere. So you can look at https://github.com/spacetelescope/synphot_refactor/issues/new/choose for examples of how the issue templates would work. As for PR template, you won't see it in action until you open a PR against that repo, but it basically pre-fill the body of the PR, so instead of blank, you get those words you see on the template.

ejeschke (approved):
@pllim, this looks reasonable to me.  I think my only concern is that for some users this may set enough "friction" that they don't follow through with an issue ("ah, there's all these fields! I won't bother"). 

But then again perhaps this will make it easier for them by the prompts.  In any case, we can adjust it later as needed.

pllim:
Indeed, not everyone reads the template, but from the experience over at `astropy`, you might have maybe 50% success rate. 😉

pllim:
I'll let you merge if you choose to accept the template. 😄

PR #919, opened 2021-01-06
Fixed Pick plugin to autozoom the pick and contour images

ejeschke:
The Pick plugin is supposed to autozoom the pick and contour images in the plugin GUI, at least until the user overrides the zoom level.  This change makes it work correctly.

### DIFF
--- a/doc/WhatsNew.rst
+++ b/doc/WhatsNew.rst
@@ -29,6 +29,7 @@ Ver 3.2.0 (unreleased)
 - Changed splitter widget so that the "thumbs" have a visual indicator
 - Fixed an issue with cursor warp in free panning with Gtk3 backend
 - Fixed an issue where the cursor was not changed from the default
+- Fixed Pick plugin to autozoom the pick and contour images
 
 Ver 3.1.0 (2020-07-20)
 ======================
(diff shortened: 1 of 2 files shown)


pllim (commented):
Before this patch:

Screenshot 2021-01-06 144728

With this patch:

Screenshot 2021-01-06 144833

I think the fixed zoom is better but not sure about the colorbar in the contour tab.

p.s. You can ignore the "triage" job error.

ejeschke:
> not sure about the colorbar in the contour tab

Something not look right?  `Pick` builds a ginga widget for Contours if you have `scikit-image` installed. This allows you to work more interactively with the contour image. It reverts to `matplotlib` (with limited interactivity) if `scikit-image` is not installed.

pllim:
> Something not look right?

The colorbar runs over the contour image. I think I am using the matplotlib visualization.

ejeschke:
> I think I am using the matplotlib visualization.

No, that is the default position for the `colorbar` canvas object in `ginga`. You can tell it is `ginga` widget because it has the usual mouse and keyboard bindings (try setting contrast with `Ctrl+right`).

We can plot it a little lower, or maybe put it in a separate widget like it is in the channel viewer, but that should be a different PR, I think.

ejeschke:
If it is blocking the main contours, try using middle click or `Shift`+ left click to pan the image a little.

ejeschke:
Also, if you make the widget larger, by pulling down the adjustable pane, then the colorbar falls to a lower position in the image.  We could also make the default height of the panel taller.

pllim (approved):
Ah, okay. I agree that colorbar positioning can be future PR, if needed. Thanks for the clarifications. Feel free to merge!

PR #923, opened 2021-01-06
MNT: Compat with astropy 4.3

pllim:
Compatibility with unreleased astropy 4.3 as a result of astropy/astropy#11006

### DIFF
--- a/ginga/doc/download_doc.py
+++ b/ginga/doc/download_doc.py
@@ -5,11 +5,16 @@ import zipfile
 import urllib
 
 from astropy.utils import minversion
-from astropy.utils.data import _find_pkg_data_path
 
 from ginga import toolkit
+from ginga.util.toolbox import ASTROPY_LT_4_3
 
 __all__ = ['get_doc']
+
+if ASTROPY_LT_4_3:
+    from astropy.utils.data import _find_pkg_data_path as get_pkg_data_path
+else:
+    from astropy.utils.data import get_pkg_data_path
 
 
 def _find_rtd_version():
@@ -81,7 +86,7 @@ def _download_rtd_zip(rtd_version=None, **kwargs):
         rtd_version = _find_rtd_version()
 
     data_path = os.path.dirname(
-        _find_pkg_data_path('help.html', package='ginga.doc'))
+        get_pkg_data_path('help.html', package='ginga.doc'))
     index_html = os.path.join(data_path, 'index.html')
 
     # There is a previous download of documentation; Do nothing.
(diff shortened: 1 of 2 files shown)


ejeschke (approved):
Ugh. An API change that is just a name change. I guess `astropy` didn't want to support the old name in deprecation warning for a couple releases?

Guess we'll be able to remove the check once 4.3 is sufficiently far back enough in the rear-view mirror.

ginga/doc/download_doc.py
```diff
@@ -5,12 +5,17 @@
...
 from ginga import toolkit
+from ginga.util.toolbox import ASTROPY_LT_4_3
 
 __all__ = ['get_doc']
 
+if ASTROPY_LT_4_3:
+    from astropy.utils.data import _find_pkg_data_path as get_pkg_data_path
+else:
+    from astropy.utils.data import get_pkg_data_path
+
```
ejeschke, line 19: @pllim, any rationale for this instead of the usual:
```python
try:
    from astropy.utils.data import get_pkg_data_path
except ImportError:
    from astropy.utils.data import _find_pkg_data_path as get_pkg_data_path
```
?

ginga/util/toolbox.py
```diff
@@ -11,8 +11,16 @@
...
+import astropy
 from astropy.utils.exceptions import AstropyUserWarning
+from astropy.utils.introspection import minversion
+
+ASTROPY_LT_4_3 = not minversion(astropy, '4.3')
+
+if ASTROPY_LT_4_3:
+    from astropy.utils.data import _find_pkg_data_path as get_pkg_data_path
+else:
+    from astropy.utils.data import get_pkg_data_path
```
ejeschke, line 24: Ditto.

ejeschke:
I'm good either way. If you prefer the version check, it's certainly more understandable in the code, but could also just put that note in a comment if use the above approach.

ginga/doc/download_doc.py
pllim: `except` is expensive and all usage with `astropy<4.3` (which is basically everyone not using the current dev because 4.3 is unreleased) will hit it. But maybe this is premature optimization. Also, having `if` check also makes finding and removing compatibility easier because the pattern is predictable compared to "human words" in in-line comments.

I'd like to keep the check as is, unless you have strong objection.

Thanks for the review!

pllim:
> I guess `astropy` didn't want to support the old name in deprecation warning for a couple releases?

Actually, the deprecation was discussed in the `astropy` PR and I strongly supported to **not** deprecate it because it was "hidden" function and it is common understanding that "hidden" things can change without notice (`numpy` also follows the same understanding).

ejeschke:
Ah, that makes sense. FFTM.

PR #928, opened 2021-02-01
MNT: Use float instead of np.float, int instead of np.int

pllim:
Numpy 1.20 is released and is causing warnings like this in Ginga:

```
E   DeprecationWarning: `np.float` is a deprecated alias for the builtin `float`.
To silence this warning, use `float` by itself. Doing this will not modify any behavior and is safe.
If you specifically wanted the numpy scalar type, use `np.float64` here.
E   Deprecated in NumPy 1.20; for more details and guidance:
https://numpy.org/devdocs/release/1.20.0-notes.html#deprecations
```

```
E DeprecationWarning: `np.int` is a deprecated alias for the builtin `int`.
To silence this warning, use `int` by itself. Doing this will not modify any behavior and is safe.
When replacing `np.int`, you may wish to use e.g. `np.int64` or `np.int32` to specify the precision.
If you wish to review your current use, check the release note link for additional information.
E Deprecated in NumPy 1.20; for more details and guidance:
https://numpy.org/devdocs/release/1.20.0-notes.html#deprecations
```

### DIFF
--- a/ginga/Bindings.py
+++ b/ginga/Bindings.py
@@ -580,14 +580,14 @@ class ImageViewBindings(object):
 
         arr = np.array([(mnwd, mnht), (mxwd, mnht),
                         (mxwd, mxht), (mnwd, mxht)],
-                       dtype=np.float)
+                       dtype=float)
         x, y = tr.to_(arr).T
 
         rx1, rx2 = np.min(x), np.max(x)
         ry1, ry2 = np.min(y), np.max(y)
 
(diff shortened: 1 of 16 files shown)


ejeschke (approved):
@pllim, looks reasonable to me.  If they were just aliases for `float` and `int`, then I wouldn't expect any change other than removing the deprecation warning.

pllim:
Yes, I would hope so. Merging then. Thanks for the review!

PR #929, opened 2021-02-16
Add orientation setting to all plugins that are orientable

ejeschke:
Adds an "orientation" setting to all plugins that use the "get_oriented_box" function.  It allows the user to fix (override) the auto-orientation.

### DIFF
--- a/doc/WhatsNew.rst
+++ b/doc/WhatsNew.rst
@@ -32,6 +32,7 @@ Ver 3.2.0 (unreleased)
 - Fixed Pick plugin to autozoom the pick and contour images
 - Fixed an issue where Thumbs plugin might not show initial thumb(s)
   when main window is enlarged to certain sizes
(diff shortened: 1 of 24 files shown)


ejeschke:
@pllim, have look when you get the chance.  We have an instance of Ginga in the observation system where the layout is such that the primary orientation for the plugins is horizontal, but due to our (somewhat) recent change of requiring an aspect ratio of 2.0 to force a horizontal layout, they do not lay out as desired.  This PR simply allow one to put e.g. orientation='horizontal' in the plugin settings file and this would override the auto-orientation.

pllim (approved):
I don't see anything wrong with the code but I don't have time to test it as a user right now.

I'll test all the changes as a whole when we have a RC to test. Thanks!

pllim:
p.s. Feel free to merge. 😸

PR #930, opened 2021-02-19
MNT: Replace fromstring with frombuffer, bump numpy minversion

pllim:
I see this warning while working on astropy/astrowidgets#125

```
.../ginga/aggw/CanvasRenderAgg.py:215: DeprecationWarning: The binary mode of fromstring is deprecated,
as it behaves surprisingly on unicode inputs. Use frombuffer instead
    arr8 = np.fromstring(self.surface.tobytes(), dtype=np.uint8)
```

### DIFF
--- a/doc/install.rst
+++ b/doc/install.rst
@@ -21,7 +21,7 @@ REQUIRED
 
 * python (v. 3.7 or higher)
 * setuptools-scm
-* numpy  (v. 1.13 or higher)
+* numpy  (v. 1.14 or higher)
 * astropy
 
 Highly recommended, because some features will not be available without it:
--- a/ginga/AstroImage.py
+++ b/ginga/AstroImage.py
@@ -182,7 +182,7 @@ class AstroImage(BaseImage):
 
     def load_buffer(self, buf, dims, dtype, byteswap=False,
                     naxispath=None, metadata=None):
-        data = np.fromstring(buf, dtype=dtype)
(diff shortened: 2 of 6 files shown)


ejeschke (commented):
@pllim , what version of `numpy` deprecates `fromstring`?  I think there are a few instances where ginga checks for `frombuffer` and falls back to `fromstring` (and similar for `tobuffer`/`tostring` ?).  If we are only supporting numpy versions that have `frombuffer`/`tobuffer` then we can modify those sections too.

pllim:
If I understand https://numpy.org/doc/stable/reference/generated/numpy.fromstring.html correctly, it has been deprecated since 1.14.  I am not familiar with `tostring` but these are all the `fromstring` I found in Ginga.

ejeschke:
Then.., currently `setup.cfg` says that we require:
```python
install_requires = numpy>=1.13
```
maybe we should bump that to 1.14 in this PR?  What do you think?

I can submit a separate PR to tackle the other cases, if I find any.

pllim:
1.14 is pretty old. Even astropy only support `numpy>=1.17` now.

ejeschke:
@plllim, shall I SAM (Squash And Merge)? :smiley: Or are you planning to push more commits?

pllim:
@ejeschke , feel free to SAM. I am done here. Thanks! 😸

PR #932, opened 2021-02-19
TST: Stop ignoring warnings

pllim:
I was wondering how #930 fell through the cracks and then I noticed that we have been ignoring warnings in `pytest`.

~I suspect this will smoke out a lot of warnings, so I still put this on draft status for now until we can figure out how to handle them.~

Well, that was easier than expected. This is ready for review now.

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -1,9 +1,9 @@
 [tool:pytest]
-minversion = 3
+minversion = 5
 testpaths = "ginga" "doc"
 norecursedirs = build doc/_build
 astropy_header = true
-addopts = -p no:warnings --doctest-ignore-import-errors
+addopts = --doctest-ignore-import-errors
 
 [bdist_wheel]
 universal = 1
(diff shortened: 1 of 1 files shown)


pllim:
Huh... Interesting, no other warning other than numpy size (which seems harmless). Maybe we don't test the stuff affected in #930 here in the CI. 🤷

ejeschke:
Is it giving a deprecation warning?

pllim:
@ejeschke , for #930 , yes.

ejeschke (approved):
Looks ok to me, @pllim. 

Just out of curiosity, do you have any insight as to the two warnings we are filtering (size changed)?

pllim:
I think it is from upstream. For example, this one is from `astropy` and it is a known issue -- https://github.com/ejeschke/ginga/runs/1936579281?check_suite_focus=true

pllim:
If you are okay with this, please merge. Thanks!

ejeschke:
> If you are okay with this, please merge. Thanks!

@pllim, ok to squash and merge, or do you want to retain commits?

pllim:
Feel free to squash! 😸

PR #935, opened 2021-03-01
TST: Fix old deps job by bumping install-qt

pllim:
I noticed failure in old deps job that is supposedly fixed already upstream.


ejeschke (approved):
I was wondering about those! Thanks for looking into this, @pllim. Just for my curiosity, are those github hash changes, and/or where do they come from?

pllim:
It is the long hash of the associated release, https://github.com/jurplel/install-qt-action/releases/tag/v2.13.0 .

Also see https://julienrenaux.fr/2019/12/20/github-actions-security-risk/

pllim:
@ejeschke , if you are okay with this, feel free to merge. Thanks!

PR #937, opened 2021-03-04
MNT: Remove duplicate license

pllim:
It is confusing to have 2 copies of the same license and they risk diverging if we forget to update the other one. `LICENSE.txt` is mentioned a lot more, so I pick that one to keep.


ejeschke (approved):
FFTM!

pllim:
Thanks for the review!

PR #941, opened 2021-04-04
Add high performance interactive X/Y plot support

ejeschke:
This PR adds some canvas object types that implement high-performance interactive X/Y plots in a Ginga viewer.
    
As an example, the `Pick` plugin uses one of these plots for the quick mode Cuts plot.


ejeschke:
@pllim, have a look when you get a chance.  Try the two programs added to the ginga/examples/gw folder:
* plot_sine.py
* plot_time_series.py

 and check out the [new chapter I added to the dev manual](https://ginga--941.org.readthedocs.build/en/941/dev_manual/plots.html#).  

Also, try the `Pick` plugin (which uses this plotting), checking the "Quick Mode" option and viewing the "Cuts" tab (after you have established a square area you can just move it around the image without holding down the button; graph will update as you sweep over any object. Press "r" or simply click to have it analyze what's under the box.)

ejeschke:
For the `Pick` plugin, you can go to the "Controls" tab and set warning and alert levels for the peak value shown in the "Cuts" tab.  This can be a quick barometer of values that exceeded some recommended threshold on your detectors.

ejeschke:
Kind of important thing I forgot: please install the package `numpy-indexed` to see the best performance with the time-series plots:
```bash
$ pip install numpy-indexed
```

pllim (changes requested):
I don't think the new Cuts plot in Pick plugin is working for me. I used [jclj01010_drz.fits](https://stsci.box.com/s/k3c0lek2zcon54skqrw01abfh66al4gw) and picked an object at X=1154 and Y=1298. When I enabled Quick Mode, the Cuts plot show up, but when I move cursor way from the object (mouse is over image display), the plot keeps updating as I move my cursor around, which was not what I expect to happen. As a result, the Cuts plot is no longer in sync with the rest of the plugin tabs, which are still centered on the object I picked, and I lost the useful Cuts info.

With this patch, the box moves around with my cursor:

Screenshot 2021-05-14 135756

Current dev version:

Screenshot 2021-05-14 135607

pllim:
Also, personally, I find the "X: NNNN   Y: NNNN" on top of the Cuts plot in Pick plugin confusing, as they unfortunately align with the vertical grid lines. Took me a while to realize they have nothing to do with the grid line placement and are only legends.

doc/quickref.rst
```diff
@@ -427,6 +427,37 @@ Cmap mode
 | Y                    | Restore the color map to default ('gray').        |
 +----------------------+---------------------------------------------------+
 
+.. _plot_mode:
+
+Plot mode
```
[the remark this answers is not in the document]
ejeschke: There is a note in the quick ref about how that mode can only be used with a viewer showing a plot with a `PlotAide`. The aide sets up that mode and puts the viewer in it.  The viewer can only be in one mode at a time (or none) so it will overwrite any mode the user set up before passing it in to `PlotAide`.

ginga/canvas/types/plots.py
```diff
@@ -0,0 +1,842 @@
...
+#
+# This is open-source software licensed under a BSD license.
+# Please see the file LICENSE.txt for details.
+#
+import sys
+import numpy as np
+have_npi = False
+try:
+    # NOTE: $ pip install numpy-indexed
+    import numpy_indexed as npi
```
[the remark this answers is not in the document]
ejeschke: Let me take a look and see if there is another way to accomplish the grouping. I feel like `numpy` itself ought to have some of that functionality, but it was not apparent to me, despite a little bit of experimentation.

ginga/canvas/types/plots.py
```diff
@@ -0,0 +1,842 @@
...
+
+        x_data, y_data = points.T
+
+        # if we can determine the visible region shown on the plot
+        # limit the points to those within the region
+        if np.all(np.isfinite([start_x, stop_x])):
+            idx = np.logical_and(x_data >= start_x, x_data <= stop_x)
+            points = points[idx]
+
+        if have_npi and self.x_func is not None:
```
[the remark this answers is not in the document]
ejeschke: That's right, I just need `group_by`.  I'll look into `pandas` support for something like that, or something using native Python.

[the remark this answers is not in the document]
ejeschke: @pllim, thanks for pushing back on this one! I think this will do nicely and seems to be just as fast:
```python
            u, i = np.unique(cx, return_index=True)
            gr_pts = np.split(points, i)
            x_data = np.array([self.x_func(a.T[0]) for a in gr_pts
                               if len(a) > 0])
            y_data = np.array([self.y_func(a.T[1]) for a in gr_pts
                               if len(a) > 0])
            assert len(x_data) == len(y_data)

            points = np.array((x_data, y_data)).T
```

ejeschke:
@pllim, I think I addressed all of your concerns except for the `Pick` plugin changes. How about if I simply withdraw `Pick` changes from this PR and submit as a new PR to be worked out separately?

ejeschke:
I reverted `Pick`--will resubmit as a separate PR.

pllim:
Sounds good to me. Thanks. I won't be able to follow up till Monday. If you can't wait, feel free to merge.

ejeschke:
No hurry, @pllim. Have a good weekend!

ginga/canvas/types/plots.py
```diff
@@ -0,0 +1,832 @@
...
+            # Reduce each group of Y points that map to a unique X via a
+            # function that reduces to a single value.  The desirable function
+            # will depend on the function of the plot, but mean() would be a
+            # sensible default
+            _, i = np.unique(cx, return_index=True)
+            gr_pts = np.split(points, i)
+            x_data = np.array([self.x_func(a.T[0]) for a in gr_pts
+                               if len(a) > 0])
+            y_data = np.array([self.y_func(a.T[1]) for a in gr_pts
+                               if len(a) > 0])
```
pllim, line 149: Is this too much code simply to avoid tranposing `a` twice?
```suggestion
            x_data = []
            y_data = []
            for a in gr_pts:
                if len(a) <= 0:
                    continue
                a_t = a.T
                x_data.append(self.x_func(a_t[0]))
                y_data.append(self.y_func(a_t[1]))

            x_data = np.array(x_data)
            y_data = np.array(y_data)
```

ginga/canvas/types/plots.py
```diff
@@ -0,0 +1,842 @@
...
+
+    def update_resize(self, viewer, dims):
+        """Called when the viewer is resized."""
+        self.recalc(viewer)
+
+    def update_elements(self, viewer):
+        """This method is called if the plot is set with new points,
+        or is scaled or panned with existing points.
+        """
+        #self.recalc(viewer)
```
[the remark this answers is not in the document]
pllim: Still not sure about this. Can you please clarify?

ginga/canvas/types/plots.py
```diff
@@ -0,0 +1,832 @@
...
+        if flips[2]:
+            x, y = y, x
+        return (x, y)
+
+    def get_title(self):
+        titles_d = self.aide.get_axes_titles()
+        return titles_d[self.kind]
+
+    def add_plot(self, viewer, plot_src):
+        pass
```
pllim, line 275: Why `pass`? Why not `raise NotImplementedError`?

ginga/canvas/types/plots.py
```diff
@@ -0,0 +1,832 @@
...
+            # format according to user's preference
+            lbl.text = self.format_value(x)
+
+    def update_bbox(self, viewer, dims):
+        """This method is called if the viewer's window is resized.
+
+        Update all the XAxis elements to reflect the new dimensions.
+        """
+        title = self.get_title()
+        self._title.text = title if title is not None else '555.55'
```
pllim, line 345: Why is the default `'555.55'`?

ginga/canvas/types/plots.py
```diff
@@ -0,0 +1,832 @@
...
+            x, y = self.get_data_xy(viewer, (lbl.x, lbl.y))
+            lbl.text = self.format_value(y)
+
+    def update_bbox(self, viewer, dims):
+        """This method is called if the viewer's window is resized.
+
+        Update all the YAxis elements to reflect the new dimensions.
+        """
+        title = self.get_title()
+        self._title.text = title if title is not None else '555.55'
```
pllim, line 472: Same question about `'555.55'` here.

ginga/examples/configs/plugin_Pick.cfg
```diff
@@ -26,7 +26,7 @@ quick_mode = False
 quick_from_peak = True
 # for future use
 quick_update_interval = 0.25
-quick_drag_only = True
+quick_drag_only = False
```
pllim, line 29: This should be reverted too.

ginga/examples/gw/plot_time_series.py
```diff
@@ -0,0 +1,364 @@
...
+24 hours of data is kept.  Each plot starts out in "autoaxis X PAN"
+and "autoaxis Y PLOT".
+
+Things you can do in a plot:
+
+1) Scroll the mouse wheel to zoom the X axis
+
+2) Hold CTRL and scroll the mouse wheel to zoom the Y axis.
+
+3) Press 'f' to autozoom the Y axis to the full range of the Y data.
```
pllim, line 20: Is this a typo? f key didn't do anything for me.
```suggestion
3) Press 'v' to autozoom the Y axis to the full range of the Y data.
```

pllim (commented):
Glad you found a solution that doesn't require numpy-indexed. Minor comments below.

General feedback for `plot_time_series.py` example as I run it, though I realize it is just an example, so it is up to you whether you wish to do anything about it. As a user, I feel like it would be nice to have the following instead:

1. When I click, instead of showing Y-position of the cursor, highlight and show the Y-position of the nearest data point at the X-position of the cursor, because that is probably of more interest to me than the mouse cursor position report.
2. Some sort of text status bar to show me why a plot is in red/yellow status. I also thought about hover text when I mouse over the red plot but that might be annoying to some people.

Hope this helps!

ginga/canvas/types/plots.py
ejeschke: I think the list comprehension is shorter and perhaps easier to understand. Probably faster than a Python loop with appends too.

pllim: Depends on how expensive the transpose operation is. Hard to tell without having some realistic data to benchmark on my end. If you think this is premature optimization or even regression, feel free to ignore. Thanks!

ginga/examples/gw/plot_time_series.py
ejeschke: Not in sync with the code, should be 'y' and 'v', not 'f' and 'y'.  Will fix.

ginga/canvas/types/plots.py
ejeschke: Pretty sure that numpy transpose (with .T) does not reorder data in memory, but only makes a new stride/offset for accessing the data. (https://stackoverflow.com/questions/19479384/is-numpy-transpose-reordering-data-in-memory).

pllim: Okay then. I don't know why I remember transposing a cube was eating up all my processing time. Maybe the way I had to do it, it created a copy. 🤷

ginga/canvas/types/plots.py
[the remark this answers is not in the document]
ejeschke: Sorry, missed a few review comments in this file cause it hid the diff. Removing this method. Other plot elements are called with this method when data changes (e.g. axes, etc).  But the `plot*` methods are called on this object to plot new data.

ginga/canvas/types/plots.py
ejeschke: Good point.

ginga/canvas/types/plots.py
```diff
@@ -0,0 +1,842 @@
...
+                # y out of range of plot area, so make it invisible
+                self.ln_alert.alpha = 0.0
+
+            self.ln_alert.y1 = self.ln_alert.y2 = y
+
+        self.check_warning()
+
+    def update_bbox(self, viewer, dims):
+        # this object does not adjust the plot bbox at all
+        pass
```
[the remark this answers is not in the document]
ejeschke: For plot decor objects, this is part of the internal API. The plot aide does a two-pass sequence (``update_resize()`` in ``PlotAide``) when the viewer is resized where it firstly calls `update_bbox()` on every plot decor object. By the end of this first pass they have all together (without knowing it individually) decided the bounding box of the plot area itself. On the second pass, ``update_resize()`` is called on all the plot decor to finalize the layout in each one.

For some plot decor, there may be no contributing change to the plot bbox, so the call is a no-op.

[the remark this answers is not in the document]
pllim: Gotcha. Thanks!

ginga/canvas/types/plots.py
ejeschke: Just something to calculate the pixel size of text height. Not sure why I chose that particular value, other than I thought '5' kind of represents a typical digit size. This value is not shown if the user did not actually set a title, just used for the pixel size calculation.

ginga/canvas/types/plots.py
ejeschke: Hmm, I guess one reason is that ``XAxis`` and ``YAxis`` then have to reimplement the methods with ``pass`` if they don't need to update their state when a plot is added or deleted.  Neither of them need to do anything for this part of the internal plot decor API.  I don't know if it makes it more clear to duplicate these methods in the subclasses.  I changed it so that there is a more detailed comment above the ``pass``.

ejeschke:
Ok, @pllim, I think I have addressed most of these concerns. Take a last look if you want and LMK if LGTM. (LMKILGTM ?? Lol :smile: )

ejeschke:
Thanks, @pllim! Very helpful reviews!

PR #943, opened 2021-04-08
MNT: Update astropy default branch

pllim:
This is an automated update made by the `batchpr` tool :robot: to because `astropy` has changed its default branch! You can report issues to @pllim.


ejeschke:
Or should we do it after 3.2 and all the extant PRs that will be merged, will have been?

PR #944, opened 2021-04-09
MNT: Remove unused astropy config code

pllim:
This is probably leftover from package template. Ginga does not use the `astropy` configuration system at all.

Also see astropy/astropy#11497


ejeschke (approved):
@pllim, Thank you!

I was wondering about this code and whether it was still necessary!

FFTM

PR #945, opened 2021-04-10
Add interactive setting of cut levels within Histogram plugin

### LINKED ISSUE #942: Histogram: Slider to control cut levels
Currently, the cut levels are set by manually typing in numbers into text boxes. I wonder if it would be more convenient if the limits can be set interactively. Given that `matplotlib` does not handle clicks well (unless you have a way to do that already via callbacks), maybe slider is our next best option. But interactive slider might be too sensitive limits that are too far apart but maybe not a deal breaker? This was suggested by @eteq .

screenshot

ejeschke:
Fix #942

You can set cut levels by clicking in the histogram plot:

* left click: set low cut
* middle click: reset (auto cut levels)
* right click: set high cut

In addition, you can dynamically adjust the gap between low and high cuts by scrolling the wheel in the plot (i.e. the "width" of the histogram plot curve).  This has the effect of increasing or decreasing the contrast within the image.  The amount that is changed for each wheel click is set by the plugin configuration file setting ``scroll_pct``.  The default is 10%.

### DIFF
--- a/doc/WhatsNew.rst
+++ b/doc/WhatsNew.rst
@@ -33,6 +33,8 @@ Ver 3.2.0 (unreleased)
 - Fixed an issue where Thumbs plugin might not show initial thumb(s)
   when main window is enlarged to certain sizes
 - Added "orientation" setting to orientable plugins
+- Enhancements to Histogram plugin: ability to click in plot to set low
+  and high cuts levels, scroll plot to expand or contract cuts width
 
 Ver 3.1.0 (2020-07-20)
 ======================
--- a/ginga/Bindings.py
+++ b/ginga/Bindings.py
@@ -829,7 +829,7 @@ class ImageViewBindings(object):
                 loval, hival))
         viewer.cut_levels(loval, hival)
 
-    def _cut_pct(self, viewer, pct, msg=True):
+    def cut_pct(self, viewer, pct, msg=True):
         msg = self.settings.get('msg_cuts', msg)
         loval, hival = viewer.get_cut_levels()
         spread = hival - loval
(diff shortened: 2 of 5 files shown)


pllim (approved):
Wow, this is magical!

Although, I am not sure how useful the scroll functionality is for me, but I guess it doesn't hurt?

ejeschke:
A added a small tweak so that the cuts are visible when "plot by cuts" is chosen.  Previously, this option mapped the cut levels out of the plot so that the lines were not visible.  Now you can see where the cuts are.

ejeschke:
> Although, I am not sure how useful the scroll functionality is for me, but I guess it doesn't hurt?

I suspect if you play with it more you may find out it's uses.  For one thing, it allows you to quickly adjust the cuts if you got them too close together or too far apart.  You can scroll to expand the cuts, say, and then click to reset either end more precisely.

pllim:
I am not sure if "plot by cuts" is working as intended. When I check that box, all I see the is X-axis is reversed.

ejeschke:
> I am not sure if "plot by cuts" is working as intended. When I check that box, all I see the is X-axis is reversed.

That seems quite odd.  Is there a file in the Box directory that I can test with?

pllim:
@ejeschke , can you access https://stsci.box.com/s/nxr28bqo0gcvyk0725ndcsjomd99nzu9 ?

ejeschke:
Thanks, @pllim, that was super helpful. Could you update your copy of the PR and try now? I think I've got the calculation for the limits right.

pllim:
Works fine for me now. Thanks for the quick implementation!

Feel free to merge. 😸

PR #947, opened 2021-04-26
TST: Pin astropy in conda job

### LINKED ISSUE #946: TST: conda job pulling in old astropy
Not sure why it is not using latest released version of astropy.

`astropy-3.2.3              |   py38h516909a_0         7.2 MB  conda-forge`

As a result, I am seeing failures that are already fixed.

pllim:
Fix #946

### DIFF
--- a/.github/workflows/ci_workflows.yml
+++ b/.github/workflows/ci_workflows.yml
@@ -118,7 +118,7 @@ jobs:
         # $CONDA is an environment variable pointing to the root of the miniconda directory
         $CONDA/bin/conda install pip setuptools setuptools_scm
         $CONDA/bin/conda install qtpy pyqt matplotlib
-        $CONDA/bin/conda install attrs astropy pytest-astropy -c conda-forge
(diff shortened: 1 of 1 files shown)


pllim:
Looks like a simple pin fixed the job. Perhaps it was using some cached conda thingies before...

```
astropy-4.2.1              |   py38h6323ea4_1         6.1 MB
```

mwcraig:
This seems fine to me 😃

ejeschke (commented):
Looks ok to me.  I suppose it doesn't make sense to test with `astropy` v3.x.y ?

pllim:
LTS version of astropy is already 4.0.* and will be 5.0.* at the end of 2021, so I don't think we need to worry about astropy 3.0.* unless you have users who are locked into older versions of astropy.

ejeschke:
I think we are using 4 in most places.

pllim:
In that case, I'll go ahead and merge this so CI will be green again. If you still wish to test astropy 3.x in conda, we can address that in a follow-up PR. Thanks!

PR #951, opened 2021-05-15
Modify Pick to use ginga plot for the Cuts plot

ejeschke:
This modifies `Pick` to use the new Ginga `XYPlot` for the Cuts plot used in `Pick`.

This plot has better performance for updating rapidly when moved around the image (matplotlib plot could be quite sluggish).
Also, the plugin is modified a bit to better reflect the way that "Quick mode" is supposed to operate. (explanation TBA)

Depends on merge of #941

### DIFF
--- a/ginga/examples/configs/plugin_Pick.cfg
+++ b/ginga/examples/configs/plugin_Pick.cfg
@@ -26,7 +26,7 @@ quick_mode = False
 quick_from_peak = True
 # for future use
 quick_update_interval = 0.25
-quick_drag_only = True
+quick_drag_only = False
 
 # should the pick shape recenter on the found object center, if any?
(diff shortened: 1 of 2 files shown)


pllim:
Since this is still in draft form, I am not going to review this yet. Let's wait till after #941 is in.

However, you do need to move the `plugin_Pick.cfg` default value change from #941 to this PR. FYI.

ejeschke:
> However, you do need to move the plugin_Pick.cfg default value change from #941 to this PR.

Added and rebased after merge of #941

ejeschke:
@pllim, I propose #952 over this PR.

ejeschke:
Closed via merge of #952

PR #953, opened 2021-05-27
Restore the pipeline renderer as the default renderer

ejeschke:
This was supposed to be the default renderer in release 3.2 but got reverted to some odd rendering that I noticed on one or two occasions.  This restores it to the default.  Need to confirm that it is working properly.


ejeschke:
Rebased

ejeschke:
Whups!

PR #955, opened 2021-05-28
Remove numexpr optimizations

ejeschke:
This removes the use of the `numexpr` package for optimizing rotations.  There are several other (better)  ways to optimize rotations including using `opencv` and `pyopengl` packages, which also add other optimizations.


pllim (approved):
I never install `numexpr`, so I won't miss it. 😹

PR #956, opened 2021-05-28
Remove support for OpenCL

ejeschke:
Remove support for OpenCL.  For the standard renderer, `opencv` is more generally useful, and for GPU optimization, the OpenGL support is better and more widely supported and likely to be used.

This simplifies the code base and documentation.


pllim:
I vaguely remember some GL stuff giving me segfault because my graphics card is too old (#837). Will this bring that problem back?

ejeschke:
No, this is regarding the **OpenCL** support--removing it.  **OpenGL** support is still there, but is not used by default--you have to invoke with --opengl option.

pllim:
But if OpenCL is gone, that means my only other option is OpenGL?

ejeschke:
> But if OpenCL is gone, that means my only other option is OpenGL?

OpenCv provides good support for conventional CPU-based transformations. But OpenGL is now the defacto standard for GPU-based computation. OpenCL was Intel's attempt to forge their own standard, but it never really took off.  Whereas OpenGL is in everything, including the web-browser you use.

ejeschke:
Sorry, didn't quite answer the question. You'll still be fine. Install `python-opencv` and enjoy similar speedup's as you would have gotten with OpenCL, but without the hassle! 😄

ejeschke:
And try `--opengl` if you have a modern display card and a larger monitor...

pllim:
OK, good to know. Thanks!

p.s. Not all my browsers support WebGL, if that is the same thing as OpenGL.

ejeschke:
> Not all my browsers support WebGL, if that is the same thing as OpenGL.

Well, chrome, firefox, safari, edge do--IFF you have a modern graphics card or laptop with a GPU.

pllim:
I am still stuck with RHEL 7 with no possibility to upgrade. 😿

ejeschke:
> I am still stuck with RHEL 7 with no possibility to upgrade.

Sounds like "I am stuck in H*ll". :smiling_imp: 

Well, what about a container or a VM?  Then you could install whatever Linux you want in it!

pllim:
> a container or a VM?

Sigh. Long story... But don't worry too much about my machine. I will work around it. I don't think my machine should block this but it is nice to still have a working Ginga without GL in case other users are in similar situation as I am.

ejeschke:
@pllim, any objection to merge?

ejeschke:
I'll fix the conflict first, of course.

ejeschke:
De-conflicted and rebased.

pllim (approved):
FFTM. I am all for simplifying code. Thanks!

PR #957, opened 2021-05-28
Clarify that opencv acceleration is used by default

ejeschke:
Previously, OpenCv support had to be explicitly enabled via a command line switch or option in a configuration file.  This was due to a problem importing `cv2` that could crash the application on some platforms.  The `OpenCv` project has fixed that bug some time ago and for quite a while now, Ginga has been quietly enabling OpenCv support by default.

This PR removes the `--opencv` switch from example programs and updates the documentation to indicate that it is used by default if installed.  The option is retained on the reference viewer, however it will generate a logger warning that the option has been deprecated.

### DIFF
--- a/ginga/web/pgw/ipg.py
+++ b/ginga/web/pgw/ipg.py
@@ -461,18 +461,14 @@ class WebServer(object):
 
 def make_server(logger=None, basedir='.', numthreads=5,
                 host='localhost', port=9909, viewer_class=None,
-                use_opencv=False):
+                use_opencv=None):
 
     if logger is None:
         logger = log.get_logger("ipg", null=True)
     ev_quit = threading.Event()
 
-    if use_opencv:
-        from ginga import trcalc
-        try:
-            trcalc.use('opencv')
-        except Exception as e:
-            logger.warning("Error using opencv: %s" % str(e))
+    if use_opencv is not None:
+        logger.warning("use_opencv parameter is deprecated")
 
     thread_pool = Task.ThreadPool(numthreads, logger,
                                   ev_quit=ev_quit)
@@ -495,7 +491,7 @@ def main(options, args):
 
     server = make_server(logger=logger, basedir=options.basedir,
                          numthreads=options.numthreads, host=options.host,
-                         port=options.port, use_opencv=options.use_opencv)
+                         port=options.port)
     viewer = server.get_viewer('v1')
 
     logger.info("Starting server with one viewer, connect at %s" % viewer.url)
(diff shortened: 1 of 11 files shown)


ejeschke:
@pllim, I'm going to recommend merging before 3.2 release, since this corrects the documentation for what is already the case.

ginga/web/pgw/ipg.py
```diff
@@ -461,18 +461,14 @@ def get_viewer(self, v_id, viewer_class=None, width=512, height=512,
...
     ev_quit = threading.Event()
 
-    if use_opencv:
-        from ginga import trcalc
-        try:
-            trcalc.use('opencv')
-        except Exception as e:
-            logger.warning("Error using opencv: %s" % str(e))
+    if use_opencv is not None:
+        logger.warning("use_opencv parameter is deprecated")
```
pllim, line 471:
```suggestion
        logger.warning("use_opencv parameter is deprecated, OpenCv will be used if installed")
```

ginga/web/pgw/ipg.py
```diff
@@ -535,9 +531,6 @@ def main(options, args):
     argprs.add_argument("--stderr", dest="logstderr", default=False,
                         action="store_true",
                         help="Copy logging also to stderr")
-    argprs.add_argument("--opencv", dest="use_opencv", default=False,
```
pllim, line 538: Do you want to also issue a warning here instead of outright removing it? Will user call this via CLI directly?

pllim (commented):
I don't think I have opencv installed either. How will this affect me as a user?

pllim:
I think we also have to modify the `astrowidgets` API to match depending on the Ginga version it pulls. cc @mwcraig 

https://github.com/astropy/astrowidgets/blob/ace5abb820c030122add475ecf106114aaa68e1b/astrowidgets/core.py#L63

ejeschke:
> I don't think I have opencv installed either. How will this affect me as a user?

Depending on how intensively you use Ginga, probably wont affect you much.  If you do a fair bit of rotation or maybe mosaicing, it might speed things up if you install it.

ejeschke:
> I think we also have to modify the `astrowidgets` API to match depending on the Ginga version it pulls. cc @mwcraig
> 
> https://github.com/astropy/astrowidgets/blob/ace5abb820c030122add475ecf106114aaa68e1b/astrowidgets/core.py#L63

Hmm, except I haven't changed the API at all here.

pllim:
Re: astrowidgets -- If I understood this PR correctly, we won't need `use_opencv=True` option in `astrowidgets` anymore, right?

ejeschke:
> Re: astrowidgets -- If I understood this PR correctly, we won't need use_opencv=True option in astrowidgets anymore, right?

Correct. But I didn't remove it from the API.  It will just ignore it for now.

ejeschke:
> It will just ignore it for now.

It was already using OpenCv it you had it installed, no matter what you passed in.

ejeschke:
@pllim, OK2M?

pllim (approved):
Sure! Have a good weekend.

Issue #959, opened 2021-06-02
Merge wcs_astropy and wcs_astropy_ape14


pllim:
Particularly, `wcs_astropy_ape14` should just be renamed to `wcs_astropy`. At the same time, we deprecate `wcs_astropy_ape14` and eventually remove it. So there should be only one `astropy` WCS interface and it is the one that follows [Astropy APE 14](https://github.com/astropy/astropy-APEs/blob/main/APE14.rst).

As suggested by @ejeschke in https://github.com/ejeschke/ginga/pull/823#issuecomment-850755829.

ejeschke:
I think that as long as the one we end up with is working as well as the other (plus the extra functionality), then we are good.  This is a little bit tricky issue because the WCS is used for sooo many different files!

pllim:
I don't understand why the hack is there and why it needs such a big offset (1 degree in the sky). When there is distortion, such a big offset outside of the detector is not going to give you anything good.

pllim:
Oh ops... I replied to the wrong issue.

PR #967, opened 2021-07-12
wcs_astropy_ape14: Fall back to old way when no convergence

### LINKED ISSUE #960: Pan "hack" causing compass failure for HST/ACS FLC image
I am not sure why it is forcing an error here for a big (1 degree) offset.

https://github.com/ejeschke/ginga/blob/833517ab092a525b2eac63b736515a21951dacd2/ginga/rv/plugins/Pan.py#L310-L311

This causes the following failure for HST/ACS FLC image, [jbt7a3k7q_flc.fits](https://stsci.box.com/s/gitjnk0xxb0q0g5hbygy7wtu7td5yhvd). The detector has 202x202 arcsec^2 FOV and the image WCS is not corrected for distortion at this stage.

```
WCSError: 'WCS.all_world2pix' failed to converge to the requested accuracy.
After 2 iterations, the solution is diverging at least for one input point.
```

pllim:
Fix #960 . I don't understand why the "old way" works but not the APE 14 interface. I wonder if @mcara can advise here. The image is [jbt7a3k7q_flc.fits](https://stsci.box.com/s/gitjnk0xxb0q0g5hbygy7wtu7td5yhvd).

```
Filename: jbt7a3k7q_flc.fits
No.    Name      Ver    Type      Cards   Dimensions   Format
  0  PRIMARY       1 PrimaryHDU     281   ()      
  1  SCI           1 ImageHDU       249   (4096, 2048)   float32   
  2  ERR           1 ImageHDU        56   (4096, 2048)   float32   
  3  DQ            1 ImageHDU        48   (4096, 2048)   int16   
  4  SCI           2 ImageHDU       245   (4096, 2048)   float32   
  5  ERR           2 ImageHDU        56   (4096, 2048)   float32   
  6  DQ            2 ImageHDU        48   (4096, 2048)   int16   
  7  HDRLET        1 NonstandardExtHDU     22   (60480,)      
  8  HDRLET        2 NonstandardExtHDU     22   (60480,)      
  9  HDRLET        3 NonstandardExtHDU     26   (112320,)      
 10  HDRLET        4 NonstandardExtHDU     26   (112320,)      
 11  HDRLET        5 NonstandardExtHDU     26   (112320,)      
 12  WCSCORR       1 BinTableHDU     59   14R x 24C   [40A, I, A, 24A, 24A, 24A, 24A, D, D, D, D, D, D, D, D, 24A, 24A, D, D, D, D, J, 40A, 128A]   
 13  WCSDVARR      1 ImageHDU        15   (64, 32)   float32   
 14  WCSDVARR      2 ImageHDU        15   (64, 32)   float32   
 15  D2IMARR       1 ImageHDU        15   (64, 32)   float32   
 16  D2IMARR       2 ImageHDU        15   (64, 32)   float32   
 17  WCSDVARR      3 ImageHDU        15   (64, 32)   float32   
 18  WCSDVARR      4 ImageHDU        15   (64, 32)   float32   
 19  D2IMARR       3 ImageHDU        15   (64, 32)   float32   
 20  D2IMARR       4 ImageHDU        15   (64, 32)   float32   
 21  HDRLET        6 NonstandardExtHDU     26   (112320,)      
 22  HDRLET        7 NonstandardExtHDU     26   (112320,)      
 23  HDRLET        8 NonstandardExtHDU     26   (112320,)  
```
```
WCS Keywords

Number of WCS axes: 2
CTYPE : 'RA---TAN-SIP'  'DEC--TAN-SIP'  
CRVAL : 168.9840199589  1.4973412658182  
CRPIX : 2048.0  1024.0  
CD1_1 CD1_2  : -2.1416145107558e-06  -1.3918446407705e-05  
CD2_1 CD2_2  : -1.3680995986142e-05  1.2042976985148e-06  
NAXIS : 4096  2048
```

This PR is a pre-requisite for #959 .

### DIFF
--- a/ginga/util/wcsmod/wcs_astropy_ape14.py
+++ b/ginga/util/wcsmod/wcs_astropy_ape14.py
@@ -7,6 +7,7 @@ from astropy import coordinates
 from astropy import units as u
 from astropy import wcs
 from astropy.io import fits
+from astropy.wcs import NoConvergence
 
 from ginga.util.wcsmod import common
 
@@ -127,6 +128,18 @@ class AstropyWCS(common.BaseWCS):
 
         try:
             xy = self.wcs.world_to_pixel_values(*args)[:2]
+        except NoConvergence:  # Fall back to pre-APE 14 calculations
+            if coords == 'data':
+                origin = 0
+            else:
+                origin = 1
+            try:
+                pix = self.wcs.wcs_world2pix(np.array([args], np.float_), origin)
+            except Exception as e:
+                self.logger.error(
+                    "Error calculating radectopix: {}".format(str(e)))
+                raise common.WCSError(e)
+            xy = float(pix[0, 0]), float(pix[0, 1])
         except Exception as e:
             self.logger.error(
                 "Error calculating radectopix: {}".format(str(e)))


ejeschke (approved):
Seems like a reasonable approach in the case of no convergence, if the API for the old method will be retained.  Is it in the works for astropy.WCS to keep supporting that?

mcara:
> Fix #960 . I don't understand why the "old way" works but not the APE 14 interface. I wonder if @mcara can advise here. 

On line 137 you call `pix = self.wcs.wcs_world2pix(np.array([args], np.float_), origin)` - notice **`wcs_`**`world2pix` - which **does NOT** use iteration to revert coordinates since it ignores polynomial (SIP) distortions and inversion can be performed analytically. APE14 calls **`all_`**`world2pix` which **will** fail if you are away by 1 degree from the center of an ACS image.

This may be an unnecessary complication, but instead of this complicated pirouette with adding 1 degree and catching exceptions, could you create a bounding box around image(s) (multiple images for multiple ACS chips) and then create a spherical polygon and test if a point is inside the polygon before trying to invert coordinates? Just a suggestion.

pllim:
> adding +1 degree is in a different module

I did try to make the "hack" smaller but it didn't fix the problem. Somewhere else in the code, it tries to do something that also triggers `NoConvergence`. Between `Compass` and the lower level `wcs.py` in the code base, I got lost where this is really happening. Maybe @ejeschke knows.

> create spherical polygon

I hope you are not asking me to add `spherical-geometry` as a dependency here.

mcara:
> I hope you are not asking me to add spherical-geometry as a dependency here.

He-he. Yes, that's what I meant.

> I did try to make the "hack" smaller but it didn't fix the problem.

What is the problem you are trying to solve by adding +1 degree?

mcara:
Instead of `spherical_geometry`, you could take corners of the image and compute maximum "distance" (in degrees) from CRVAL and consider that to be the "radius of convergence" for coordinate inversion. If a point is inside the circle - call APE14, if not - call `wcs_world2pix`.

PR #970, opened 2021-08-25
Change Pick plot of FWHM to match rest of UI report

### LINKED ISSUE #969: Pick function showing different FWHM values between the FWHM figure and the Readout tab
Sometimes when using Pick, the FWHM value at the top of the FWHM figure doesn't match the value in the Readout tab. In the attached example, the FWHM Y at the top of the figure is 7.70 px but in the Readout tab is 5.625 px. What is the difference between these two values? 

Note that that I am trying to fit a double peaked feature and it is the direction of the double peak that shows the FWHM discrepancy. 

Thanks! 
Screen Shot 2021-08-17 at 8 10 48 PM
Screen Shot 2021-08-17 at 8 08 10 PM

ejeschke:
Changes the center point of FWHM calculation in `evaluate_peaks` from the local peak to the center-of-mass centroid of the local peak, if it can be calculated.

This may result in a slightly more accurate FWHM in the `evaluate_peaks` pass that can be used to solve the problem outlined in #969

Fix #969

### DIFF
--- a/ginga/rv/plugins/Pick.py
+++ b/ginga/rv/plugins/Pick.py
@@ -1235,13 +1235,13 @@ class Pick(GingaPlugin.LocalPlugin):
 
     def plot_fwhm(self, qs, image):
         # Make a FWHM plot
-        x, y, radius = qs.x, qs.y, qs.fwhm_radius
+        x, y = qs.x - self.pick_x1, qs.y - self.pick_y1
+        radius = qs.fwhm_radius
 
         try:
-            self.fwhm_plot.plot_fwhm(x, y, radius, image,
-                                     cutout_data=self.pick_data,
-                                     iqcalc=self.iqcalc,
-                                     fwhm_method=self.fwhm_alg)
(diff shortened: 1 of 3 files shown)


ejeschke:
@pllim, if you think this approach has merit, I can push an additional commit to fix `Pick` to (potentially) solve the problem outlined in #969.  We would need some careful testing (mostly with "ill-behaved" objects) to make sure it is working correctly.

pllim:
I think it is good to have consistent values displayed across plugin and don't see anything wrong with using center of mass. 👍

obi-wan76:
Screen Shot 2021-08-31 at 11 54 09 AM
Screen Shot 2021-08-31 at 11 54 01 AM

Yes, the values are consistent between the figure and Pick Readout. Thanks!

pllim:
> That may alarm some users.

@ejeschke , which one is the lesser of two evils?

@obi-wan76 is happy with this patch, so I am okay with merging as-is unless you still have concerns.

pllim:
Thanks for the fix! 😄

PR #971, opened 2021-09-23
Change Contents so that "genthumb" setting of channel applies

ejeschke:
This makes the `Contents` plugin in the reference viewer behave so that if the `genthumb` setting in the channel is set to False that
no entry is made in Contents. Additionally, if an individual image has `nothumb`  == True then no entry is added.

This makes Contents behave similarly to Thumbs for channel and image settings.

### DIFF
--- a/ginga/rv/plugins/Contents.py
+++ b/ginga/rv/plugins/Contents.py
@@ -277,7 +277,9 @@ class Contents(GingaPlugin.GlobalPlugin):
         self.logger.debug("name=%s" % (name))
 
         if image is not None:
-            nothumb = image.get('nothumb', False)
+            channel = self.fv.get_channel(chname)
+            nothumb = (image.get('nothumb', False) or
+                       not channel.settings.get('genthumb', True))
             if nothumb:
                 return
 
@@ -313,6 +315,9 @@ class Contents(GingaPlugin.GlobalPlugin):
         """Almost the same as add_image_cb(), except that the image
         may not be loaded in memory.
         """
+        if not channel.settings.get('genthumb', True):
+            return
+
         chname = channel.name
         name = image_info.name
         self.logger.debug("name=%s" % (name))
@@ -365,6 +370,9 @@ class Contents(GingaPlugin.GlobalPlugin):
     def add_channel_cb(self, viewer, channel):
         """Called when a channel is added from the main interface.
         Parameter is a channel (a Channel object)."""
+        if not channel.settings.get('genthumb', True):
+            return
+
         chname = channel.name
 
         # add old highlight set to channel external data
@@ -386,6 +394,9 @@ class Contents(GingaPlugin.GlobalPlugin):
     def delete_channel_cb(self, viewer, channel):
         """Called when a channel is deleted from the main interface.
         Parameter is a channel (a Channel object)."""
+        if not channel.settings.get('genthumb', True):
+            return
+
         chname = channel.name
         del self.name_dict[chname]
 
@@ -447,6 +458,8 @@ class Contents(GingaPlugin.GlobalPlugin):
 
     def redo(self, channel, image):
         """This method is called when an image is set in a channel."""
+        if not channel.settings.get('genthumb', True):
+            return
 
         imname = image.get('name', 'none')
         chname = channel.name
@@ -504,8 +517,7 @@ class Contents(GingaPlugin.GlobalPlugin):
             self.fv.show_error("Please select some images first")
             return
 
-        l_img = list(map(lambda tup: "%s/%s" % (tup[0], tup[1].imname),
-                         images))
+        l_img = ["%s/%s" % (tup[0], tup[1].imname) for tup in images]
 
         verb = action.capitalize()
         l_img.insert(0, "%s images\n" % (verb))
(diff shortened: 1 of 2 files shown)


pllim:
I can see why someone might not want thumbnails, but I don't see why Contents also needs to hide them. How does one remove data from channel or move it to another channel if it is not visible in Contents? Can you please clarify? Thanks!

ejeschke:
@pllim, I agree for typical use.  However, there are cases where you don't want `Contents` to be cluttered up with hundreds of entries for particular channels.  A good example is receiving guider images every few seconds throughout a night.  It just doesn't make sense to add a Contents entry for every one of these.  In our case we don't even save most of these guider images on disk and they just disappear after exiting the channel buffer. 

It was for this purpose that the "nothumb" flag in images and the "genthumb" setting in a channel were implemented.  At the time, the Contents plugin did not exist (thus the naming scheme for these flags).  But currently the rationale for using these in the case of the example above is the same for Contents as it is for Thumbs.

In any case, the default is always to create Thumbs and Contents entries. It would be worth thinking about a future PR in which there could be a separate setting for whether Contents entries are created.

pllim:
Thanks for the explanation! In that case, perhaps the docstring of `Contents.py` needs to be updated to explain the usage of this option.

ejeschke:
@pllim, good suggestion.  I added some info to the docstrings of both ``Contents`` and ``Thumbs``.

ginga/rv/plugins/Contents.py
```diff
@@ -39,6 +39,22 @@
 make it so by setting the "closeable" setting to True in the configuration
 file--then Close and Help buttons will be added to the bottom of the UI.
 
+** Excluding images from Contents **
```
pllim, line 42: Remove whitespace to correctly render it as bold font.
```suggestion
**Excluding images from Contents**
```

ginga/rv/plugins/Contents.py
```diff
@@ -39,6 +39,22 @@
...
+
+Although the default behavior is for every image that is loaded into the
+reference viewer to show up in ``Contents``, there may be cases where this
+is undesirable (e.g. when there are many images being loaded at a
```
pllim, line 46:
```suggestion
is undesirable (e.g., when there are many images being loaded at a
```

ginga/rv/plugins/Contents.py
```diff
@@ -39,6 +39,22 @@
...
+periodic rate by some automated process).  In such cases there are two
+mechanisms for suppressing certain images from showing up in ``Contents``:
+
+* Assigning the "genthumb" setting to False in a channel's settings
+  (for example from the ``Preferences`` plugin, under the "General"
+  settings) will exclude the channel itself and any of its images.
+* Setting the "nothumb" keyword in the metadata of an image wrapper
+  (not the FITS header, but by e.g. ``image.set(nothumb=True)``)
```
pllim, line 54:
```suggestion
  (not the FITS header, but by e.g., ``image.set(nothumb=True)``)
```

ginga/rv/plugins/Thumbs.py
```diff
@@ -27,6 +27,22 @@
 make it so by setting the "closeable" setting to True in the configuration
 file--then Close and Help buttons will be added to the bottom of the UI.
 
+** Excluding images from Thumbs **
```
pllim, line 30:
```suggestion
**Excluding images from Thumbs**
```

ginga/rv/plugins/Thumbs.py
```diff
@@ -27,6 +27,22 @@
...
+
+Although the default behavior is for every image that is loaded into the
+reference viewer to show up in ``Thumbs``, there may be cases where this
+is undesirable (e.g. when there are many images being loaded at a
```
pllim, line 34:
```suggestion
is undesirable (e.g., when there are many images being loaded at a
```

ginga/rv/plugins/Thumbs.py
```diff
@@ -27,6 +27,22 @@
...
+periodic rate by some automated process).  In such cases there are two
+mechanisms for suppressing certain images from showing up in ``Thumbs``:
+
+* Assigning the "genthumb" setting to False in a channel's settings
+  (for example from the ``Preferences`` plugin, under the "General"
+  settings) will exclude the channel itself and any of its images.
+* Setting the "nothumb" keyword in the metadata of an image wrapper
+  (not the FITS header, but by e.g. ``image.set(nothumb=True)``)
```
pllim, line 42:
```suggestion
  (not the FITS header, but by e.g., ``image.set(nothumb=True)``)
```

ginga/rv/plugins/Thumbs.py
```diff
@@ -27,6 +27,22 @@
 make it so by setting the "closeable" setting to True in the configuration
 file--then Close and Help buttons will be added to the bottom of the UI.
 
+** Excluding images from Thumbs **
+
```
pllim, line 31:
```suggestion

.. note:: This also controls the behavior of ``Contents``.

```

ginga/rv/plugins/Contents.py
```diff
@@ -39,6 +39,22 @@
 make it so by setting the "closeable" setting to True in the configuration
 file--then Close and Help buttons will be added to the bottom of the UI.
 
+** Excluding images from Contents **
+
```
pllim, line 43:
```suggestion

.. note:: This also controls the behavior of ``Thumbs``.

```

pllim (commented):
Thanks! Some minor comments about the doc.

ejeschke:
@pllim, OKTM?

pllim:
FFTM 😸

PR #972, opened 2021-10-05
This modifies the reference viewer to stop all plugins before exit

ejeschke:
This PR modifies the reference viewer to stop all plugins that are running before exiting.

Mostly to make things terminate more cleanly if a plugin has some particular state, now ``stop()`` will be called on the plugin if it was activated at the time of the viewer closing.


pllim (commented):
Seems fine code wise. Do you want me to manually test this out? Will I even notice anything?

ejeschke:
> Do you want me to manually test this out? Will I even notice anything?

You shouldn't notice anything.  In any case, nothing about this is tested until you close the program.  If you want, you might try opening a few of your plugins and then close the viewer, checking for any error messages from the plugin in the log.

This is more for specialty plugins that do something like open a socket or start a thread while they are open. Ginga might not exit cleanly if the `stop()` method is not called on the plugins to clean up/shut down this state. We have a couple of specialty ones like that and this is helpful to make sure things shut down cleanly.

ejeschke:
Thank you @pllim, merging.

PR #974, opened 2021-10-30
Add NoConvergence workaround for radectopix in wcs_astropy.py

ejeschke:
Fixes the routines ``radectopix`` and ``wcspt_to_datapt`` in the WCS wrapper for astropy so that if the convergence fails it falls back to the best solution (based on the information in [this issue](https://github.com/astropy/astropy/issues/3203)).

The rationale for this is that these routines are mainly used for routines like drawing a compass, etc and do not need extreme accuracy of solution.

### DIFF
--- a/ginga/util/wcsmod/wcs_astropy.py
+++ b/ginga/util/wcsmod/wcs_astropy.py
@@ -121,7 +121,11 @@ class AstropyWCS(common.BaseWCS):
         skycrd = np.array([args], np.float_)
 
         try:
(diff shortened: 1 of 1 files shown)


ejeschke:
@pllim, can #967 use this technique? What do you think?

pllim:
Thanks! But I might not have time to review for a few weeks. Is this urgent?

ejeschke:
No worries, @pllim. Not sure a review is even necessary for this.

PR #978, opened 2021-12-03
Refactor handling of inherit_primary_header in AstroImage

ejeschke:
This PR refactors AstroImage, AstroTable and ginga.util.io_fits in the following ways:
- the primary header is saved as a metadata item called 'primary_header' (the header is already saved similarly as 'header'); this makes the handling of these two entities similar.
- the astropy/fitsio specific details for loading an HDU has been moved from AstroImage and AstroTable to ginga.util.io_fits. This encapsulates the details of the specific support package in io_fits
- added loading of FITS tables via the `fitsio` package in io_fits
- modified the TableView class to be able to view tables loaded with either astropy or fitsio

### DIFF
--- a/doc/WhatsNew.rst
+++ b/doc/WhatsNew.rst
@@ -29,7 +29,8 @@ Ver 3.3.0 (unreleased)
   - for previous parameter of usecrop=True, use sample=crop
 - Moved loading of FITS HDUs from AstroImage to io_fits module,
   encapsulating the details of this file format into the module
-  responsible for loading those files
+  responsible for loading those files:
+
   - added loading of FITS tables via the fitsio package in io_fits
   - TableView can now view tables loaded with either astropy or fitsio
   - inherit_primary_header in general.cfg now defaults to True and
--- a/doc/WhatsNew.rst
+++ b/doc/WhatsNew.rst
@@ -27,6 +27,14 @@ Ver 3.3.0 (unreleased)
 - Changes to 'histogram' and 'stddev' autocuts algorithms:
   - choice of sampling by grid now; useful for mosaics and collages
   - for previous parameter of usecrop=True, use sample=crop
+- Moved loading of FITS HDUs from AstroImage to io_fits module,
+  encapsulating the details of this file format into the module
+  responsible for loading those files
+  - added loading of FITS tables via the fitsio package in io_fits
+  - TableView can now view tables loaded with either astropy or fitsio
+  - inherit_primary_header in general.cfg now defaults to True and
+    actually controls the behavior of always saving the primary header
+    if set to False
 
 Ver 3.2.0 (2021-06-07)
 ======================
--- a/doc/WhatsNew.rst
+++ b/doc/WhatsNew.rst
@@ -25,8 +25,10 @@ Ver 3.3.0 (unreleased)
   workspace type that was selected in the drop down menu
 - Updated pg widgets web backend due to changes to Tornado asyncio handling
 - Changes to 'histogram' and 'stddev' autocuts algorithms:
+  
   - choice of sampling by grid now; useful for mosaics and collages
   - for previous parameter of usecrop=True, use sample=crop
+    
 - Moved loading of FITS HDUs from AstroImage to io_fits module,
   encapsulating the details of this file format into the module
   responsible for loading those files:
--- a/ginga/AstroImage.py
+++ b/ginga/AstroImage.py
@@ -102,7 +102,8 @@ class AstroImage(BaseImage):
         if self.io is None:
             raise ImageError("No IO loader defined")
 
-        self.io.load_hdu(hdu, fobj=fobj, naxispath=naxispath)
+        self.io.load_hdu(hdu, dstobj=self, fobj=fobj, naxispath=naxispath,
+                         save_primary_header=False)
 
     def load_nddata(self, ndd, naxispath=None):
         """Load from an astropy.nddata.NDData object.
--- a/ginga/AstroImage.py
+++ b/ginga/AstroImage.py
@@ -277,7 +277,7 @@ class AstroImage(BaseImage):
 
     def has_primary_header(self):
         primary_hdr = self.metadata.get('primary_header', None)
-        return primary_hdr is not None
+        return isinstance(primary_hdr, AstroHeader)
 
     def clear_all(self):
         # clear metadata and data
--- a/ginga/AstroImage.py
+++ b/ginga/AstroImage.py
@@ -38,7 +38,8 @@ class AstroImage(BaseImage):
         cls.ioClass = klass
 
     def __init__(self, data_np=None, metadata=None, logger=None,
-                 name=None, wcsclass=None, ioclass=None):
+                 name=None, wcsclass=None, ioclass=None,
+                 inherit_primary_header=False, save_primary_header=True):
 
         BaseImage.__init__(self, data_np=data_np, metadata=metadata,
                            logger=logger, name=name)
@@ -60,6 +61,9 @@ class AstroImage(BaseImage):
                 ioclass = self.ioClass
 
         self.io = ioclass(self.logger)
+
+        self.inherit_primary_header = inherit_primary_header
+        self.save_primary_header = inherit_primary_header or save_primary_header
 
         if metadata is not None:
             header = self.get_header()
@@ -96,14 +100,23 @@ class AstroImage(BaseImage):
 
         self.set_naxispath(naxispath)
 
-    def load_hdu(self, hdu, fobj=None, naxispath=None):
+    def load_hdu(self, hdu, fobj=None, naxispath=None,
+                 inherit_primary_header=None):
         """NOTE: this is for astropy.io.fits HDUs only."""
 
         if self.io is None:
             raise ImageError("No IO loader defined")
 
+        if inherit_primary_header is None:
+            inherit_primary_header = self.inherit_primary_header
+        else:  # This ensures get_header() is consistent
+            self.inherit_primary_header = inherit_primary_header
+
+        save_primary_header = (self.save_primary_header or
+                               inherit_primary_header)
+
         self.io.load_hdu(hdu, dstobj=self, fobj=fobj, naxispath=naxispath,
-                         save_primary_header=False)
+                         save_primary_header=save_primary_header)
 
     def load_nddata(self, ndd, naxispath=None):
         """Load from an astropy.nddata.NDData object.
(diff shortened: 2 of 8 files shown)


pllim (commented):
I don't know how I feel about this. This is definitely a breaking change for `stginga`. What prompted this refactor?

ejeschke:
@pllim, I certainly don't want to break things for `stginga`.  I'm just trying to clean up the "include primary header" implementation because it felt inconsistent with the way the regular headers are stored and handled. Could you let me know what breaks with this PR?  I'm am happy to modify it.

With this PR, the primary header is always saved, and is available from the image by `image.get('primary_header')`. The primary header is visible in the `Header` plugin by clicking the checkbox.

pllim:
I'll have to double check. At the very least, there is a setting in `general.cfg`. I will also have to see if I call that affected API directly or not. I'll get back to you in a week or two. Thanks for your patience!

https://github.com/ejeschke/ginga/blob/04973bfe4e74d4c804abea7124d309b720b0a7a0/ginga/examples/configs/general.cfg#L51-L52

ejeschke:
> Thanks for your patience!

No hurry on this one, @pllim!

ginga/AstroImage.py
```diff
@@ -38,8 +38,7 @@ def set_ioClass(cls, klass):
         cls.ioClass = klass
 
     def __init__(self, data_np=None, metadata=None, logger=None,
-                 name=None, wcsclass=None, ioclass=None,
-                 inherit_primary_header=False, save_primary_header=True):
```
pllim, line 42: I don't see `stginga` explicitly calling this in the constructor. Looks like `stginga` only sets it via `general.cfg` and you did not change the config name there, so that's good.

However, there is no guarantee that all your downstream packages/users also only set this via `general.cfg`. This is a public API change without a deprecation period. I am not sure if it is a good practice.

pllim: Recommendation: Instead of removing the keywords, deprecate them but preserve a backward-compatible functionality with deprecation warning, change log, and API documentation informing users and downstream devs. Wait for maybe two releases (your call on exactly how many) before removing them altogether.

Pro: No surprise for anyone.
Con: More work for you.

ginga/AstroImage.py
```diff
@@ -62,14 +61,6 @@ def __init__(self, data_np=None, metadata=None, logger=None,
 
         self.io = ioclass(self.logger)
 
-        self.inherit_primary_header = inherit_primary_header
-        self.save_primary_header = inherit_primary_header or save_primary_header
```
pllim, line 66: Same comment as https://github.com/ejeschke/ginga/pull/978/files#r776394464

ginga/AstroImage.py
```diff
@@ -106,41 +97,13 @@ def setup_data(self, data, naxispath=None):
...
-        # collect HDU header
-        ahdr = self.get_header()
-        self._copy_hdu_header(hdu, ahdr)
-
-        # Set PRIMARY header
-        if inherit_primary_header is None:
-            inherit_primary_header = self.inherit_primary_header
-        else:  # This ensures get_header() is consistent
-            self.inherit_primary_header = inherit_primary_header
+                 save_primary_header=False):
```
pllim, line 100: Same comment as https://github.com/ejeschke/ginga/pull/978/files#r776394464 , except the warning here will be a bit different (renaming instead of removing).

`astropy` supports both cases: See https://github.com/astropy/astropy/blob/main/astropy/utils/decorators.py

ginga/AstroImage.py
```diff
@@ -221,27 +184,19 @@ def set_io(self, io):
     def get_data_size(self):
         return self.get_size()
 
-    def get_header(self, create=True, include_primary_header=None):
+    def get_header(self, create=True, include_primary_header=False):
```
pllim, line 187: Is this a behavior change, so would warrant mention in change log too but no need to undo the change.

ginga/AstroImage.py
```diff
@@ -221,27 +184,19 @@ def set_io(self, io):
...
-                    card = hdr.get_card(key)
-                    bnch = displayhdr.__setitem__(card.key, card.value)
-                    bnch.comment = card.comment
-                for key in self._primary_hdr.keyorder:
-                    if key not in hdr:
-                        card = self._primary_hdr.get_card(key)
-                        bnch = displayhdr.__setitem__(card.key, card.value)
-                        bnch.comment = card.comment
+                displayhdr.merge(hdr)
+                displayhdr.merge(primary_hdr)
```
pllim, line 199: Better be explicit here instead of relying on default setting, which might change in the future. One should never overwrite keyword in data extension from primary header.

```suggestion
                displayhdr.merge(hdr)
                displayhdr.merge(primary_hdr, override_keywords=False)
```

ginga/rv/Control.py
```diff
@@ -96,8 +96,8 @@ def __init__(self, logger, thread_pool, module_manager, preferences,
                               numImages=10,
                               # Offset to add to numpy-based coords
                               pixel_coords_offset=1.0,
-                              # inherit from primary header
-                              inherit_primary_header=False,
+                              # save primary header when loading files
+                              inherit_primary_header=True,
```
pllim, line 100: This behavior change also would need to be mentioned in change log.

ginga/table/AstroTable.py
```diff
@@ -45,7 +44,10 @@ def __init__(self, data_ap=None, metadata=None, logger=None, name=None,
 
         self._data = data_ap
         self.naxispath = []
+        self.colnames = []
 
+        # TODO: How to handle table with WCS data? For example, spectrum
+        #       table may store dispersion solution as WCS.
```
pllim, line 50: Maybe use [specutils](https://github.com/astropy/specutils)?

ginga/util/io_fits.py
```diff
@@ -246,18 +250,48 @@ def get_hdu_type(self, hdu):
 
         return None
 
-    def load_hdu(self, hdu, dstobj=None, **kwargs):
+    def load_hdu(self, hdu, dstobj=None, fobj=None, naxispath=None,
+                 save_primary_header=False, **kwargs):
```
pllim, line 254: Might also want to mention this in the change log. Addition might still break API if someone has been calling the keywords as arguments.

ginga/util/io_fits.py
```diff
@@ -288,14 +341,13 @@ def load_hdu(self, hdu, dstobj=None, **kwargs):
         return dstobj
 
     def load_file(self, filespec, numhdu=None, dstobj=None, memmap=None,
-                  **kwargs):
-        inherit_primary_header = kwargs.pop('inherit_primary_header', False)
+                  save_primary_header=False, **kwargs):
```
pllim, line 344: Same comment as https://github.com/ejeschke/ginga/pull/978/files#r776422501

ginga/util/io_fits.py
```diff
@@ -491,7 +543,7 @@ def __init__(self, logger):
         self.hdu_info = []
         self.hdu_db = {}
 
-    def fromHDU(self, hdu, ahdr):
```
pllim, line 494: Might want to deprecate this and have it return `copy_header(...)` instead of outright removal.

ginga/util/io_fits.py
```diff
@@ -509,54 +561,74 @@ def get_hdu_type(self, hdu):
 
         return None
 
-    def load_hdu(self, hdu, dstobj=None, **kwargs):
+    def load_hdu(self, hdu, dstobj=None, fobj=None, naxispath=None,
+                 save_primary_header=True, **kwargs):
```
pllim, line 565: Same comment as https://github.com/ejeschke/ginga/pull/978/files#r776422501

ginga/util/io_fits.py
```diff
@@ -509,54 +561,74 @@ def get_hdu_type(self, hdu):
...
+            dstobj.set_data(data)
 
         dstobj.io = self
 
         return dstobj
 
     def load_file(self, filespec, numhdu=None, dstobj=None, memmap=None,
-                  **kwargs):
-        inherit_primary_header = kwargs.pop('inherit_primary_header', False)
+                  save_primary_header=True, **kwargs):
```
pllim, line 625: Same comment as https://github.com/ejeschke/ginga/pull/978/files#r776422501

pllim (commented):
I have some concerns about breaking change with no deprecation period (see comments above), but at a glance, I don't see this outright breaking `stginga`.

Therefore, it is okay to proceed and I will do an in-depth review once this is out of draft form. Would be nice to have deprecation but ultimately, it is your call.

If you decide to proceed with no deprecation period, you should bump the major version the release this patch is in.

ejeschke:
Hi @pllim, can you have another look at this?  I have restored the non-private attributes and API of `AstroImage` and their former behavior.  I will submit a separate PR later for the deprecation of the items and we can discuss the issue further in that PR.

It is true that inherit_primary_header now defaults to `True` in `Control.py`, but that is a good thing, because in the previous version it did not matter at all whether it was True or False, the primary header was always saved.

ejeschke:
Rebased

ginga/AstroImage.py
```diff
@@ -223,35 +198,29 @@ def get_data_size(self):
...
-                displayhdr = hdr
-
-        except KeyError as e:
+        # By convention, the fits header is stored in a dictionary
+        # under the metadata keyword 'header'
+        if 'header' not in self.metadata:
             if not create:
-                raise e
+                # TODO: change to ValueError("No header found")
+                raise KeyError('header')
```
pllim, line 206: Do you intent to address the TODO as part of this PR?

ginga/AstroImage.py
```diff
@@ -307,7 +276,8 @@ def update_metadata(self, key_dict):
             self.wcs.load_header(header)
 
     def has_primary_header(self):
-        return self._primary_hdr is not None
+        primary_hdr = self.metadata.get('primary_header', None)
+        return primary_hdr is not None
```
pllim, line 280: Can this be simplified without actually trying to access the data that you never use?
```suggestion
        return 'primary_header' in self.metadata
```

pllim (commented):
Still needs a change log.

I don't want to be a blocker of this getting merged. But I would want to have some time to test the release candidate with this change. Thanks for your patience!

ejeschke:
> Still needs a change log.
> 
> I don't want to be a blocker of this getting merged. But I would want to have some time to test the release candidate with this change. Thanks for your patience!

No problem, @pllim.

ginga/AstroImage.py
ejeschke: No, that would change the exception type.  Will look at that for the follow-up PR.

ginga/AstroImage.py
ejeschke: I think I'd like to keep the type check on the value of the primary_header key. At least checking that it is not `None`.  We could check that it is a subclass of `Header`...

ejeschke:
I will add some comments to the WhatsNew.

ejeschke:
Rebased on master.

ginga/AstroImage.py
pllim: If you prefer to keep type check, I agree checking for `Header` might be more explicit. 😸

doc/WhatsNew.rst
```diff
@@ -27,6 +27,14 @@ Ver 3.3.0 (unreleased)
 - Changes to 'histogram' and 'stddev' autocuts algorithms:
   - choice of sampling by grid now; useful for mosaics and collages
   - for previous parameter of usecrop=True, use sample=crop
+- Moved loading of FITS HDUs from AstroImage to io_fits module,
+  encapsulating the details of this file format into the module
+  responsible for loading those files
```
pllim, line 32: Maybe this will fix `doc/WhatsNew.rst:36: ERROR: Unexpected indentation`?
```suggestion
  responsible for loading those files:
```

ejeschke: Looked at an example further down in the file.  Looks like I needed the colon and perhaps a blank line for RST to recognize the sublist.

doc/WhatsNew.rst
```diff
@@ -27,6 +27,15 @@ Ver 3.3.0 (unreleased)
 - Changes to 'histogram' and 'stddev' autocuts algorithms:
   - choice of sampling by grid now; useful for mosaics and collages
   - for previous parameter of usecrop=True, use sample=crop
```
pllim, line 29: Fixing this is out of scope, perhaps, but it needs fixing too.
```suggestion

  - choice of sampling by grid now; useful for mosaics and collages
  - for previous parameter of usecrop=True, use sample=crop
```

ejeschke: Good catch!

pllim (approved):
Can't keep this open forever... Feel free to merge. Thanks! 😸

ejeschke:
> Can't keep this open forever... Feel free to merge. Thanks! smile_cat

Thanks for the great review, @pllim. If you notice anything amiss we can fix it before the next release.

Issue #560, opened 2017-10-06
Incorporate astropy.visualization in Ginga's image scaling


pllim:
Partly address #555

ejeschke:
I'm not sure that I am ready to do that.  I'd prefer that the core widget does not depend on `astropy`. It's a general toolkit for scientific visualization and there is no particular need for that part to depend on astropy.  I'm totally fine with ginga using astropy for astronomy plugins or other parts of ginga that will be used for astronomy.

I believe the glue viewer does not even use the astropy-helpers toolkit, and it is considered an astropy affiliated project.

pllim:
Re: Glue -- Hmm, good point considering its integration is "green" over at glue-viz/glue#1407 . @astrofrog and @eteq , could you please clarify?

pllim:
FWIW `astropy.visualization` comes "free" as `astropy` is a required dependency, so it is always there anyway.

astrofrog:
Using astropy-helpers is not a requirement for affiliated packages. Glue does use astropy.visualization for the image viewer: https://github.com/glue-viz/glue/blob/master/glue/viewers/image/composite_array.py#L9

pllim:
If we want to revisit this, needs some sort of translation layer to https://github.com/ejeschke/ginga/blob/master/ginga/ColorDist.py , maybe.

Also see https://docs.astropy.org/en/stable/visualization/normalization.html

pllim:
Maybe https://github.com/ejeschke/ginga/blob/master/ginga/AutoCuts.py too.

eteq:
I'm also curious about

>  I'd prefer that the core widget does not depend on astropy

in the context of an *optional* dependency, would this be OK? I.e., if the astropy machinery is used if its present and the user passes in this machinery.

ejeschke:
> in the context of an optional dependency, would this be OK? I.e., if the astropy machinery is used if its present and the user passes in this machinery.

That is fine.

I think it would be straightforward to make a `ColorDist` or `AutoCuts` class that uses the astropy visualization stuff.  These are just objects that are called from the viewer, so they are like black boxes in a sense--you could replace them with a different instance as long as the class presents the same API.

pllim:
[...]

PR #979, opened 2021-12-05
Update autocuts algorithms 

ejeschke:
This PR updates the auto cuts algorithms in the following ways:
- auto cuts features in the viewer can now work properly with images plotted on the canvas (not just a single loaded image); for example, the output of the `Collage` plugin
- the auto cuts algorithms that defaulted to taking a small crop sample from the center of the image ('histogram' and 'stddev') are expanded to include an option to sample the data in a grid pattern across the image. This is useful for large images (collages or large data frames such as mosaics), where a small crop from the center is not representative of the data across the image and the image is much too large to apply the algorithm to the full data.
- adds a `full_px_limit` parameter to 'histogram' and 'stddev' that specifies the number of pixels limit (in image size), beyond which the full image data calculation falls back to a crop image calculation. This prevents the viewer from accidentally attempting to calculate cut levels on a really large image, which can lock up the viewer.

Here is an example of the PR improvement.  First, a collage using the 'zscale' auto cuts (before PR)
hsc_mosaic_before

and after PR:
hsc_mosaic_after

The documentation has been greatly enhanced in the `AutoCuts` module and the user manual will be expanded to include it.

### DIFF
--- a/doc/WhatsNew.rst
+++ b/doc/WhatsNew.rst
@@ -26,6 +26,7 @@ Ver 3.3.0 (unreleased)
 - Updated pg widgets web backend due to changes to Tornado asyncio handling
 - Changes to 'histogram' and 'stddev' autocuts algorithms:
   - choice of sampling by grid now; useful for mosaics and collages
+  - for previous parameter of usecrop=True, use sample=crop
 
 Ver 3.2.0 (2021-06-07)
 ======================
--- a/doc/WhatsNew.rst
+++ b/doc/WhatsNew.rst
@@ -24,6 +24,8 @@ Ver 3.3.0 (unreleased)
 - Fixed a bug where creating a new workspace did not set the correct
   workspace type that was selected in the drop down menu
 - Updated pg widgets web backend due to changes to Tornado asyncio handling
+- Changes to 'histogram' and 'stddev' autocuts algorithms:
+  - choice of sampling by grid now; useful for mosaics and collages
 
 Ver 3.2.0 (2021-06-07)
 ======================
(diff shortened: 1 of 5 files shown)


ejeschke:
Hi @pllim, have a look when you get a chance. No hurry.

ejeschke:
Rebased

doc/WhatsNew.rst
```diff
@@ -24,6 +24,8 @@ Ver 3.3.0 (unreleased)
 - Fixed a bug where creating a new workspace did not set the correct
   workspace type that was selected in the drop down menu
 - Updated pg widgets web backend due to changes to Tornado asyncio handling
+- Changes to 'histogram' and 'stddev' autocuts algorithms:
+  - choice of sampling by grid now; useful for mosaics and collages
```
pllim, line 28: Will users that only load single images see any difference? Maybe clarify that a little here?

pllim (approved):
I don't have strong opinions on how autocuts are calculated, so if your users are happy with this, it is fine by me. At a glance, I don't see any breaking change in terms of API that end-users would care about nor any configuration changes.

Out of scope here but I would still like #560 to happen someday... 😸 

👍 on documentation additions!

ejeschke:
> Out of scope here but I would still like #560 to happen someday...

We'll have to put together a sprint for that one, I think.

ejeschke:
Thanks for the review, @pllim!  I added a comment to the change log that will hopefully clarify the change for stddev and histogram algorithms (the only ones affected).

PR #985, opened 2022-01-29
Remove save_primary_header, inherit_primary_header attributes

ejeschke:
This begins the removal of the `save_primary_header` and `inherit_primary_header` attributes and kwargs from AstroImage.
The saving of the primary header is done when loading an image through the `io_fits` module and it is merged when `include_primary_header==True` in the `AstroImage` `get_header`() method.

The two kwargs issue a deprecation warning if used when creating an AstroImage.

### DIFF
--- a/doc/WhatsNew.rst
+++ b/doc/WhatsNew.rst
@@ -46,8 +46,10 @@ Ver 3.3.0 (unreleased)
 
 - Fixed an issue where the Catalogs plugin would not start correctly if
   astroquery was not installed
-- Begin deprecation of kwargs save_primary_header and
-  inherit_primary_header in the AstroImage class
+- The keywords ``save_primary_header`` and ``inherit_primary_header`` in the ``AstroImage``
+  constructor are deprecated. Use ``save_primary_header`` keyword in ``AstroImage.load_hdu()``
+  or ``AstroImage.load_file()`` methods instead. Several other methods in ``AstroImage`` are
+  deprecated as well; they were previously pending deprecation.
 
 Ver 3.2.0 (2021-06-07)
 ======================
--- a/doc/WhatsNew.rst
+++ b/doc/WhatsNew.rst
@@ -46,6 +46,8 @@ Ver 3.3.0 (unreleased)
 
 - Fixed an issue where the Catalogs plugin would not start correctly if
   astroquery was not installed
+- Begin deprecation of kwargs save_primary_header and
+  inherit_primary_header in the AstroImage class
 
 Ver 3.2.0 (2021-06-07)
 ======================
--- a/ginga/AstroImage.py
+++ b/ginga/AstroImage.py
@@ -65,7 +65,7 @@ class AstroImage(BaseImage):
         inherit_primary_header = False
         if 'inherit_primary_header' in kwargs:
             warnings.warn("inherit_primary_header kwarg has been deprecated--"
-                          "use save_primary_header when loading image",
+                          "use save_primary_header in load_hdu() or load_file()",
                           PendingDeprecationWarning)
             inherit_primary_header = kwargs['inherit_primary_header']
         self.inherit_primary_header = inherit_primary_header
@@ -73,7 +73,7 @@ class AstroImage(BaseImage):
         save_primary_header = True
         if 'save_primary_header' in kwargs:
             warnings.warn("save_primary_header kwarg has been deprecated--"
-                          "use save_primary_header when loading image",
+                          "use save_primary_header in load_hdu() or load_file()",
                           PendingDeprecationWarning)
             save_primary_header = kwargs['save_primary_header']
         self.save_primary_header = inherit_primary_header or save_primary_header
--- a/ginga/AstroImage.py
+++ b/ginga/AstroImage.py
@@ -38,8 +38,7 @@ class AstroImage(BaseImage):
         cls.ioClass = klass
 
     def __init__(self, data_np=None, metadata=None, logger=None,
-                 name=None, wcsclass=None, ioclass=None,
-                 inherit_primary_header=False, save_primary_header=True):
+                 name=None, wcsclass=None, ioclass=None, **kwargs):
 
         BaseImage.__init__(self, data_np=data_np, metadata=metadata,
                            logger=logger, name=name)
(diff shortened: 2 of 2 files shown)


ejeschke:
@pllim, in line with your comments for deprecation of API on #978.

ginga/AstroImage.py
```diff
@@ -38,8 +38,7 @@ def set_ioClass(cls, klass):
         cls.ioClass = klass
 
     def __init__(self, data_np=None, metadata=None, logger=None,
-                 name=None, wcsclass=None, ioclass=None,
-                 inherit_primary_header=False, save_primary_header=True):
```
pllim, line 42: Should keep the old signature during deprecation phase. Same comment for all similar occurrences.

ginga/AstroImage.py
```diff
@@ -62,7 +61,21 @@ def __init__(self, data_np=None, metadata=None, logger=None,
 
         self.io = ioclass(self.logger)
 
+        # these attributes to be removed--DO NOT USE!!!
+        inherit_primary_header = False
+        if 'inherit_primary_header' in kwargs:
+            warnings.warn("inherit_primary_header kwarg has been deprecated--"
+                          "use save_primary_header when loading image",
+                          PendingDeprecationWarning)
```
pllim, line 69: Not sure if Pending makes sense here. Why not just deprecate for real? Same comment for all similar occurences.

```suggestion
                          DeprecationWarning)
```

ginga/AstroImage.py
```diff
@@ -62,7 +61,21 @@ def __init__(self, data_np=None, metadata=None, logger=None,
...
+            inherit_primary_header = kwargs['inherit_primary_header']
         self.inherit_primary_header = inherit_primary_header
+
+        save_primary_header = True
+        if 'save_primary_header' in kwargs:
+            warnings.warn("save_primary_header kwarg has been deprecated--"
+                          "use save_primary_header when loading image",
```
pllim, line 76: I don't understand this warning...

ginga/AstroImage.py
```diff
@@ -368,86 +387,6 @@ def radectopix(self, ra_deg, dec_deg, format='deg', coords='data'):
         return self.wcs.radectopix(ra_deg, dec_deg, coords=coords,
                                    naxispath=self.revnaxis)
 
-    # -----> TODO:
-    #   This section has been merged into ginga.util.wcs or
-    #   ginga.util.mosaic .  Deprecate it here.
-    #
-    def get_starsep_XY(self, x1, y1, x2, y2):
```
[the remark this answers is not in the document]
ejeschke: Ok, added these back in with PendingDeprecationWarning changed to DeprecationWarning.

ginga/AstroImage.py
ejeschke: The problem is that if we leave the kwargs in we don't know whether someone passed the parameter or not.  By using the **kwargs notation we can issue the warning only if they actually passed the parameter.

I'm assuming we don't want to raise a warning just by instantiating the class, if the parameter was never passed.

ginga/AstroImage.py
ejeschke: Hmm, well, I'm ok with that.  I did a little web searching on the correct way to use the deprecation warnings, and it seemed to me that the proscribed wisdom seemed to be that first you should issue a PendingDeprecationWarning to give a heads up that something *will* be deprecated in the release after next, and then (in the next release) change it to a DeprecationWarning, which is sort of a last warning that the feature will be gone in the following release.

ginga/AstroImage.py
ejeschke: Probably needs a better message.  The idea is that when loading a file you can specify `save_primary_header=True` and it will additionally save the primary header in the AstroImage. Because that is what is actually happening--the primary header is not being actually inherited--it's just saved separately.  When you ask for the header (via `get_header()` with `include_primary_header=True` you get a merged header.

ejeschke: See what you think of the updated warning messages.

ginga/AstroImage.py
```diff
@@ -62,7 +61,21 @@ def __init__(self, data_np=None, metadata=None, logger=None,
 
         self.io = ioclass(self.logger)
 
+        # these attributes to be removed--DO NOT USE!!!
+        inherit_primary_header = False
```
ejeschke, line 65: BTW, as shown here, the `inherit_primary_header` and `save_primary_header` get the same default values (False and True, respectively) as they had previously.

ginga/AstroImage.py
pllim: Problem is I don't think PendingDeprecationWarning even shows up in the logs unless you specifically enable it.

https://docs.python.org/3/library/exceptions.html#PendingDeprecationWarning

ejeschke: Changed all `PendingDeprecationWarning` to `DeprecationWarning`

ginga/AstroImage.py
pllim: I wonder if one of the "deprecated" decorators would work for you: https://docs.astropy.org/en/latest/utils/index.html#module-astropy.utils.decorators

ginga/AstroImage.py
```diff
@@ -62,7 +61,21 @@ def __init__(self, data_np=None, metadata=None, logger=None,
...
+            warnings.warn("inherit_primary_header kwarg has been deprecated--"
+                          "use save_primary_header in load_hdu() or load_file()",
+                          DeprecationWarning)
+            inherit_primary_header = kwargs['inherit_primary_header']
         self.inherit_primary_header = inherit_primary_header
+
+        save_primary_header = True
+        if 'save_primary_header' in kwargs:
+            warnings.warn("save_primary_header kwarg has been deprecated--"
+                          "use save_primary_header in load_hdu() or load_file()",
```
pllim, line 76: OK, clearer now. Thanks! 😸

ginga/AstroImage.py
ejeschke: Hmm, documentation says "The decorator assumes that the argument with the old_name was removed from the function signature and the new_name replaced it at the same position in the signature."

But both kwargs are being removed completely and not replaced.

I think maybe for this particular PR the deprecation decorators are not necessary. But thanks for pointing them out, I will definitely use them if renaming a class method or attribute in the future!

pllim: I think if `new_name` is `None`, that would satisfy your situation but here you are deprecating two, so maybe applying that decorator twice is overkill, I agree.

ejeschke:
Rebased

ejeschke:
@pllim, ok to merge?

doc/WhatsNew.rst
```diff
@@ -46,6 +46,8 @@ Ver 3.3.0 (unreleased)
 
 - Fixed an issue where the Catalogs plugin would not start correctly if
   astroquery was not installed
+- Begin deprecation of kwargs save_primary_header and
+  inherit_primary_header in the AstroImage class
```
pllim, line 50:
```suggestion
- The keywords ``save_primary_header`` and ``inherit_primary_header`` in the ``AstroImage``
  constructor are deprecated. Use ``save_primary_header`` keyword in ``AstroImage.load_hdu()``
  or ``AstroImage.load_file()`` methods instead. Several other methods in ``AstroImage`` are
  deprecated as well; they were previously pending deprecation.
```

pllim (approved):
Just nitpick on the change log, otherwise LGTM. Thanks!

ejeschke:
Thanks, @pllim!

PR #995, opened 2022-02-23
Remove the mock backend

ejeschke:
This PR removes the mock backend.  It is rarely used and kind of a pain to keep current.

It was mostly used for a couple of tests and as a template for making a new backend. I think one of the common non-widget backends like pilw (pillow) is quite sufficient for the tests. For making a new backend, one should follow the example of the closest backend that already exists.


pllim:
No objection from me. I don't think I ever used the mock stuff. 👍

PR #990, opened 2022-02-14
Remove deprecated attributes and methods from AstroImage

ejeschke:
To be merged after the deprecation period, probably in a major release (4.0?)

PR #996, opened 2022-02-26
Remove camel-case deprecated calls

### LINKED ISSUE #898: evaluate_peaks takes bright_radius but does not use it
https://github.com/ejeschke/ginga/blob/965d1ec23fdf02b7a48c86b428153346c6ce5cad/ginga/util/iqcalc.py#L466

`ginga.util.iqcalc.evaluate_peaks()` takes an optional keyword called `bright_radius`. Calling function even passes it in, but it is not used anywhere inside `evaluate_peaks`. Did I miss something, @ejeschke ?

xref #896

ejeschke:
This removes a number of camel-case deprecated methods.  It is targeted for release 4.0.

There are a few other deprecated items also removed.

supercedes #990
fix #898

### DIFF
--- a/doc/WhatsNew.rst
+++ b/doc/WhatsNew.rst
@@ -15,6 +15,8 @@ Ver 4.0.0 (unreleased)
   active canvas bindings.
 - Removed the "Quick Mode" and "From Peak" options in the Pick plugin
   to simplify operation.
+- Many deprecated camelcase (non-PEP8) methods were removed. Use the
+  "snake-case" names instead.
 
 Ver 3.4.0 (2022-06-28)
 ======================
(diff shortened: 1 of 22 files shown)


ejeschke:
@pllim, this one should be pretty uncontroversial.  It just removes deprecated camelCase names of methods (and a few other already deprecated items).  LMK if you have any concerns.

pllim:
I don't remember if stginga or wss_tools still using any of these calls, but I don't think so? Still, I would have to double check when I get the chance. I don't think I should be blocking merge in any case, because it is trivial for downstream to change the calls. So feel free to merge if you are in a rush. Thanks!

ejeschke:
Thanks, @pllim. Trying to keep things moving along for an October release of 4.0.

PR #998, opened 2022-03-04
Make pillow a required package, change piexif to exifread

ejeschke:
Since we are planning to deprecate the mock backend, and `pillow` is used in many places, as well as being a very common package, this adds `pillow` as a requirement.  It also changes the optional `piexif` package to `exifread`, which is a little less code verbose to use, a bit more popular and well maintained.

This also restores a couple of lines needed in the OpenCv RGB file handler to make auto-orientation of RGB images work correctly.

### DIFF
--- a/ginga/trcalc.py
+++ b/ginga/trcalc.py
@@ -29,8 +29,7 @@ pil_resize = dict(nearest=Image.NEAREST,
                   bicubic=Image.BICUBIC,
                   lanczos=Image.LANCZOS)
 
-interpolation_methods = list(set(['basic'] + list(pil_resize.keys())))
-interpolation_methods.sort()
+interpolation_methods = sorted(set(['basic'] + list(pil_resize.keys())))
 
 have_opencv = False
 try:
@@ -44,8 +43,7 @@ try:
                       lanczos=cv2.INTER_LANCZOS4)
     have_opencv = True
 
-    interpolation_methods = list(set(['basic'] + list(cv2_resize.keys())))
-    interpolation_methods.sort()
+    interpolation_methods = sorted(set(['basic'] + list(cv2_resize.keys())))
 
 except ImportError:
     pass
--- a/ginga/trcalc.py
+++ b/ginga/trcalc.py
@@ -8,7 +8,6 @@ import sys
 import math
 import numpy as np
 
-interpolation_methods = ['basic']
 _use = None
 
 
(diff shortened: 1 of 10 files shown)


ejeschke:
When you have a chance, @pllim.

ginga/trcalc.py
```diff
@@ -22,52 +21,37 @@ def use(pkgname):
...
+# Python Imaging Library
+from PIL import Image
+pil_resize = dict(nearest=Image.NEAREST,
+                  linear=Image.BILINEAR,
+                  area=Image.HAMMING,
+                  bicubic=Image.BICUBIC,
+                  lanczos=Image.LANCZOS)
+
+interpolation_methods = list(set(['basic'] + list(pil_resize.keys())))
+interpolation_methods.sort()
```
pllim, line 33:
```suggestion
interpolation_methods = sorted(set(['basic'] + list(pil_resize.keys())))
```

ginga/trcalc.py
```diff
@@ -22,52 +21,37 @@ def use(pkgname):
...
-        'lanczos': PILimage.LANCZOS,
-    }
-    have_pillow = True
-
-    if 'nearest' not in interpolation_methods:
-        interpolation_methods = list(set(['basic'] +
-                                         list(pil_resize.keys())))
-        interpolation_methods.sort()
+    interpolation_methods = list(set(['basic'] + list(cv2_resize.keys())))
+    interpolation_methods.sort()
```
pllim, line 48:
```suggestion
    interpolation_methods = sorted(set(['basic'] + list(cv2_resize.keys())))
```

ginga/util/rgb_cms.py
```diff
@@ -15,10 +15,10 @@
 
 from . import paths
 
+from PIL import Image
 # How about color management (ICC profile) support?
 try:
     import PIL.ImageCms as ImageCms
```
pllim, line 21: So there is still a chance PIL has no CMS or something?

pllim (commented):
I think this is ok. Just minor comments. Thanks!

ginga/util/rgb_cms.py
ejeschke: Possibly. For a long time, it was always included when installing via `pip`, but was missing when installing via `conda` (just the CMS part). Lately, it has been included even when installing via conda, but I'm not yet assuming that is fully resolved yet.  For most uses of Ginga, CMS is not that important, so we can optionally handle its availability this way.

pllim (approved):
LGTM. Thanks! Feel free to merge.

ejeschke:
Thank you, @pllim!

PR #1002, opened 2022-04-07
Add support for PyQt6 and PySide6

ejeschke:
This adds support for PyQt6 and PySide6 widget toolkits.  It removes support for PyQt4 and PySide, since those are not supported any more by `qtpy`.

### DIFF
--- a/doc/install.rst
+++ b/doc/install.rst
@@ -57,8 +57,8 @@ do), you will need at least one of the following:
 
 .. TODO: This can be broken down in a clearer way.
 
-* QtPy (PyQt4 or PyQt5)
-* PySide (Qt4/Qt5 alternative)
+* QtPy (PyQt5 or PyQt6)
+* PySide (pyside2 or pyside6)
 * pygobject (gi) **AND** pycairo (GTK 3)
 * `tkinter <https://docs.python.org/3/library/tk.html>`_
 * matplotlib
@@ -87,7 +87,6 @@ Helpful, but not necessary (may optimize or speed up certain operations):
 * pyopengl + pycairo (for using OpenGL features; very useful for 4K or larger
   monitors)
 * filemagic (aids in identifying files when opening them)
-* Pillow (useful for various RGB file manipulations)
 
 ==============================
 Notes on Supported Widget Sets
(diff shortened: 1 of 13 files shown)


ejeschke:
Text drawn in certain widgets and icons can look fuzzy, at least on hi-res (e.g. 4K) displays. Still looking for a fix for this.

ejeschke:
I had to create test environments installing pyqt and qtpy from PyPI, siince conda does not yet have Qt6.

ejeschke:
Fixed the "fuzziness issue", at least as far as I have been able to test it.

doc/install.rst
```diff
@@ -57,8 +57,8 @@ do), you will need at least one of the following:
 
 .. TODO: This can be broken down in a clearer way.
 
-* QtPy (PyQt4 or PyQt5)
-* PySide (Qt4/Qt5 alternative)
+* QtPy (PyQt5 or PyQt6)
+* PySide (pyside2 or pyside6)
```
pllim, line 61: What's pyside2? That is not mentioned in the change log.

pllim (commented):
In principle, I am okay with removing support for old packages. Do we need to add `qt6` under `options.extra_requires` in `setup.cfg`? There is already an entry for `qt5`.

doc/install.rst
ejeschke: It's their name for the Qt5 equivalent. I guess they jumped from pyside2 to pyside6 to follow Qt's versioning.

ejeschke:
> In principle, I am okay with removing support for old packages. Do we need to add `qt6` under `options.extra_requires` in `setup.cfg`? There is already an entry for `qt5`.

What do you think? In this case either one is fine. Should we list every possible version of every back end there?

ejeschke:
> > Should we list every possible version of every back end there?
> 
> We never listed Qt4 but going forward, Qt6 is probably going to be more and more common. If we list it, it would be clear to user what they need to install for Qt6, but then again it increases maintenance burden, so I'll leave it to your good judgment.

Maybe change it to qt6 when that becomes the norm in conda? I think that's kind of the benchmark for our user-base.

ejeschke:
@pllim, do you want more time to test or are you ok to merge?

pllim:
Feel free to merge. Thanks!

PR #1007, opened 2022-05-06
TST: Update workflow versions

pllim:
Update GitHub Actions workflow versions, where applicable.


ejeschke (approved):
I'll leave it to your judgement on this, @pllim. I don't know too much about the new workflow config (I had a reasonable handle on the previous one).

I do like the better integration into github though with the "Actions" tab.

pllim:
Thanks for your vote of confidence, @ejeschke ! I think this patch is pretty safe as we are using them already in `astropy` and things seem to work as expected. 😸

PR #1009, opened 2022-05-11
DOC: Update RTD env

pllim:
RTD has changed their way to specify Python version. We are using the same thing in `astropy`, so should be uncontroversial.


ejeschke (approved):
LGTM, @pllim !

pllim:
Thanks for the review!

PR #1013, opened 2022-08-02
Fixed a longstanding issue for event handling on a ginga canvas

ejeschke:
Fixed a longstanding issue where events registered on a canvas could be masked by default key/cursor bindings (not associated with a mode). This meant, for example, that only certain keystrokes could be captured by an event handler registered on a ginga canvas, because any keystrokes that had a default binding would take precedence. Now such bindings are only executed if the event is not handled by any active canvas bindings.

### DIFF
--- a/doc/WhatsNew.rst
+++ b/doc/WhatsNew.rst
@@ -6,6 +6,13 @@ Ver 4.0.0 (unreleased)
 ======================
 - fixed getattr functionality in Bunch
 - removed the "mock" backend; use "pil" backend for similar purposes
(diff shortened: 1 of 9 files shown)


ejeschke:
Looks like our version of flake8 was updated on the CI.  I fixed all the new flake8 errors flagged by the CI.

ejeschke:
@pllim, don't think this should affect anything at all in `stginga`.

pllim:
LGTM at a glance. Alas, I don't have time to test this. Feel free to merge. Thanks!

PR #1016, opened 2022-08-17
Simplify the Pick plugin

ejeschke:
This removes two confusing checkbox options from the Pick plugin.
The "quick" mode is not actually quicker for most use cases and just makes the plugin more complex and confusing to users.

### DIFF
--- a/doc/WhatsNew.rst
+++ b/doc/WhatsNew.rst
@@ -13,6 +13,8 @@ Ver 4.0.0 (unreleased)
   keystrokes that had a default binding would take precedence.
   Now such bindings are only executed if the event is not handled by any
   active canvas bindings.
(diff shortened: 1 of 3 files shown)


pllim (commented):
Hopefully people not calling the API directly downstream. Probably doesn't need deprecation, but should at least add a note that it was removed in v4.0 in the docstring.

ejeschke:
docstring for the plugin?

pllim:
> docstring for the plugin?

Yes, the one that gets rendered on RTD. 😸

ejeschke:
@pllim, any objection to merge?

pllim:
Feel free to merge. Thanks!

ejeschke:
Rebased

Issue #1011, opened 2022-06-17
Rename ORIGIN branch name


ejeschke:
Following the lead of many other projects, change the name of "master" branch to "main".

ejeschke:
@pllim, you are more in touch with the CI integration than I am. I will need your help on this. Shall we take a stab at this for release 4.0 or wait for 4.1?

pllim:
Should not be hard to do for v4.0. Two main things:

1. `grep` for mentions of `master` in this repo and change to `main`.
2. If you mention `master` in your `github.io` stuff, you need to change that too.
3. Go into RTD admin and change the default branch from `master` to `main` but only do this after GitHub change has happened.

PR #1017, opened 2022-08-23
Make loaders discoverable via entry points

ejeschke:
This makes loaders for a Ginga viewer discoverable via entry points.  This will make it possible to write 3rd-party loaders that are not bundled with Ginga.

Loaders are consolidated in `ginga.util.io`.  Legacy stubs with warning on import are left in `ginga.util.io_rgb` and `ginga.util.io_fits`.


pllim:
I like the concept but `+1,557 −1,541` 😱 

Also, how backward compatible is this?

ejeschke:
The large line change is because I left stubs in place for backward compatibility for anyone that might be importing `io_fits` or `io_rgb` from the previous location.  The actual operation of the opener hasn't really changed as far as API.

Not quite ready for testing or review yet...

ejeschke:
@pllim, do you have time to take a look and/or test? This is the last PR to be merged for 4.0, unless we want to try to tackle #1011.  I pushed the due date for 4.0 to 2022-12-15.

Don't be daunted by LOC change, almost all is due to moving `io_fits.py` and `io_rgb.py` from `ginga.util` to `ginga.util.io`, but leaving placeholders in the old location--thus it looks like two new files and big changes to the old two files. There is no significant change to the loaders.  I will probably add some documentation still.

pllim:
> do you have time

Not much. Maybe I can look at the documentation when you add them and go from there. It is unlikely that STScI will be adding their own loaders at this point.

ejeschke:
@pllim, I updated the documentation in doc/dev_manual/internals.rst.  Since I haven't created a "ginga-loader-template" repo, I don't have new instructions on that. For this PR, just updated the code example that loads HDF5 images and a bit of wording. It recommends installing a new loader using the old `ginga_config.py` technique (which still works).

Have a look if you have time. Would like to put out a release by the 20th.

doc/WhatsNew.rst
```diff
@@ -38,6 +38,9 @@ Ver 4.0.0 (unreleased)
   the Thumbs icon
 - ScreenShot now correctly captures the background color of the viewer
 - Fixed a DeprecationWarning related to use of entry points
+- File loaders are now discoverable under the "ginga_loaders" entry point.
+  Loaders can be registered for MIME types and a mime.types file can be
```
pllim, line 42:
```suggestion
  Loaders can be registered for MIME types and a ``mime.types`` file can be
```

doc/WhatsNew.rst
```diff
@@ -38,6 +38,9 @@ Ver 4.0.0 (unreleased)
...
+  added to your $HOME/.ginga to identify types by file extension.
```
pllim, line 43:
```suggestion
  added to your ``$HOME/.ginga`` to identify types by file extension.
```

doc/dev_manual/code/io_hdf5.py
```diff
@@ -75,11 +79,11 @@ def load_idx(self, idx, **kwargs):
             from ginga import AstroImage, RGBImage
 
             header = self.get_header(idx)
-            data_np = np.copy(self._f[idx].value)
+            data_np = np.copy(self._f[idx][()])
```
pllim, line 82: I don't understand passing in empty parenthesis as a subscript. Is this HDF5 specific usage?

ejeschke: Yes, I got a deprecation warning with the earlier syntax.  This was suggested on the h5py site.  I tested it and it works fine.

PR #1030, opened 2022-11-08
Add AutoLoad plugin

ejeschke:
- a plugin that can monitor a folder for new files appearing and automatically load them into a channel
- includes updated doc and screenshot


ejeschke:
@pllim, you might find this plugin useful in certain cases.  I know you are busy, feel free to review if you have time.

ginga/rv/plugins/AutoLoad.py
```diff
@@ -0,0 +1,210 @@
...
+        self._watch = None
+        self.data_dir = self.settings.get('watch_folder', None)
+        self.regex = self.settings.get('filename_regex', None)
+        self.regex_c = None
+        self.pause_flag = self.settings.get('start_paused', False)
+        self.gui_up = False
+
+    def build_gui(self, container):
+        if not have_watchdog:
+            raise ValueError("Install 'watchdog' to use this plugin")
```
pllim, line 75:
```suggestion
            raise ImportError("Install 'watchdog' to use this plugin")
```

pllim (commented):
Thanks for the ping. This looks very powerful but I am a little concerned that most users will not be familiar with regex and will accidentally enter something that crashes it. I wonder if it needs to be stricter... 💭

ejeschke:
> I am a little concerned that most users will not be familiar with regex and will accidentally enter something that crashes it. I wonder if it needs to be stricter...

It catches any failure to compile the regex and shows the error message in the "Errors" plugin. I think if you could crash Python by passing a bad regexp to `re.compile` then Python would have a very serious weakness. I entered a number of bad regexps during my debugging of this and it never crashed. If you pass a regexp that compiles, but doesn't match your filename then simply nothing gets loaded and there is a message in the log saying that the file didn't match the regex.

pllim:
I was thinking about the case where it matches something that looks like a plugin but the file is actually corrupt.

ejeschke:
Like FITS or PNG, etc.

pllim:
Ah, sorry! I misunderstood. OK, in that case, please ignore me. Sorry for the noise. 😅 

p.s. Kinda dangerous when `numImages = 0` though. One can run out of memory.

https://github.com/ejeschke/ginga/blob/de4baf921d65afa46b48bc3ce2dd0bdecbcdbf15/ginga/examples/configs/general.cfg#L47-L49

ejeschke:
> Kinda dangerous when numImages = 0 though. One can run out of memory.

True.  I thought it was you that asked for 0 to be unlimited though, IIRC--back in the day?  These days 10 seems like a lot with huge data files.  I've thought about adding a feature that would be based on memory use per channel rather than number of images.  But that's another PR.

ejeschke:
The idea behind this plugin is for when you have a pipeline say, that is processing and spitting finished images into a folder, Ginga can load them automatically into a channel using this plugin.

pllim:
> I thought it was you that asked for 0 to be unlimited though

Yes, I did. We had a use case where user wants to load hundreds of small cutout images but they didn't know how many would come in.

I wonder if autoload should be disabled if unlimited images setting is detected...

ejeschke:
> I wonder if autoload should be disabled if unlimited images setting is detected...

I could add a warning message.

I think since the default is non-zero and the user has to explicitly set it to zero to get unlimited, and it's pretty understandable what "unlimited" means (:smile_cat:) then they could run into trouble just regularly loading images in the normal way. In that case, if they use this plugin that is their decision.

pllim:
Well... the user could unwittingly inherited a project-wide setting or set it a long time ago and forgot about it.

ejeschke:
In that case should we issue a warning the first time they load an image into an unlimited channel? What if they set the limit to 1,000,000? It's effectively the same, they could easily run out of memory anytime.

I would suggest maybe adding a separate PR that checks the total size of all images loaded and issues a pop-up warning when that is exceeded, or when memory seems to be running low...

pllim:
Separate PR is fine. Thanks!

ejeschke:
Thanks for the review, @pllim!

PR #1031, opened 2022-11-15
Enhancements to the Pipeline plugin (not yet enabled by default)

ejeschke:
- Add undo/redo capability to various stages
- Add stage docstrings
- Refinements to output stage, fixes for crop and rotate stages
  - so crop and rotate stages can be inserted more than once
- Add load/save capability to all stages
- Misc updates to stages


ejeschke:
Incremental updates/improvements to pipeline stages for RGB image processing. The `Pipeline` plugin is not yet enabled by default.

PR #1035, opened 2022-12-15
Change ColorMapPicker to be a hybrid local/global plugin

ejeschke:
This changes ColorMapPicker so that it can be opened as a local or global plugin.  If opened as a global plugin, it operates exactly as before. If opened as a local plugin, then it only effects the associated channel's viewer's color map.

We have a use case where we want to be able to restrict the change of color map for a specific channel only.


ejeschke:
@pllim, maybe not too controversial?

ginga/rv/plugins/ColorMapPicker.py
```diff
@@ -36,11 +38,19 @@
 __all__ = ['ColorMapPicker']
 
 
-class ColorMapPicker(GingaPlugin.GlobalPlugin):
+class ColorMapPicker(GingaPlugin.LocalPlugin):
```
pllim, line 41: Interesting, so subclassing LocalPlugin can still act like global plugin?

ejeschke: Yes, there are a couple of existing examples. Check out `FBrowser`.

pllim:
Feel free to merge, I won't have time to review but I don't see anything controversial at a glance and in principle, I like this new feature. Thanks!

PR #1039, opened 2022-12-29
Change implementation of splash ("About") dialog

ejeschke:
- from channel viewer to a pop-up modal dialog
- shows version in title bar

### DIFF
--- a/ginga/rv/Control.py
+++ b/ginga/rv/Control.py
@@ -1543,25 +1543,38 @@ class GingaShell(GwMain.GwMain, Widgets.Application):
             text = '1/%.2fx' % (1.0 / scalefactor)
         return text
 
-    def banner(self, raiseTab=False):
+    def banner(self, raiseTab=False):  # noqa
+        # NOTE: raiseTab kwarg ignored and will be deprecated
+        # load banner image
         banner_file = os.path.join(self.iconpath, 'ginga-splash.ppm')
-        chname = 'Ginga'
-        channel = self.get_channel_on_demand(chname)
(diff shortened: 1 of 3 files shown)


ejeschke:
@pllim, absolutely no hurry on this one. Just a small and cosmetic change.

ginga/rv/Control.py
```diff
@@ -1543,25 +1543,38 @@ def scale2text(self, scalefactor):
...
-        header = image.get_header()
-        header['VERSION'] = __version__
+        w = Viewers.GingaViewerWidget(viewer=viewer)
+        w.resize(wd, ht)
+        vbox.add_widget(w, stretch=1)
 
-        if raiseTab:
-            self.change_channel(chname)
+        viewer.set_image(image)
+        top.show()
```
pllim, line 1577: Is there a way to suppress the pop up?

pllim:
TBH, I never understood why splash banner was needed. In other commercial software, it only pops up when you go to Help and click About. Is that what this is doing?

ejeschke:
It only pops up the first time you run the program and thereafter only when you select the "About" item from the Help menu.  It can (and is) suppressed by putting a "noBanner = True" in your $HOME/.ginga/general.cfg. It does this automatically the first time you run the program.

The only difference from the past is that in the past implementation it popped up as a "Ginga" channel tab, but with no mention of the version.

pllim:
In that case, feel free to merge. Thanks!

ejeschke:
Thanks, @pllim!

PR #1044, opened 2023-03-02
Fixed a bug with the implementation of Bresenham's line algorithm

ayshih:
The implementation of Bresenham's line algorithm used for `get_pixels_on_line()` has a bug because it uses `>` instead of `>=` and `<` instead of `<=`.  Compare against the final code example in this section on the Wikipedia article: https://en.wikipedia.org/wiki/Bresenham%27s_line_algorithm#All_cases.  ~This bug means that the pixels other than the closest ones are occasionally chosen.~

Edit: After further investigation, it appears that the difference in comparison operator simply means a different choice is made when two pixels are equally close to the line.  After I confirm, I'll close this PR as an unnecessary change.

### DIFF
--- a/ginga/BaseImage.py
+++ b/ginga/BaseImage.py
@@ -539,10 +539,10 @@ class BaseImage(ViewerObjectBase):
             if (x == x2) and (y == y2):
                 break
             e2 = 2 * err
-            if e2 > -dy:
+            if e2 >= -dy:
                 err = err - dy
                 x += sx
-            if e2 < dx:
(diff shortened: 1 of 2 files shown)


ayshih:
Sorry for the noise!  It turns out this "bug" is simply a difference in choice when there are two pixels equally close to the line.  This change is not necessary.

ejeschke:
@ayshih , thanks for the interest and the thoughtful analysis!

PR #1047, opened 2023-04-18
Add contrast and brightness controls to Preferences

ejeschke:
This adds contrast and brightness (bias) sliders to the channel preferences, offering another way to set these values that may be more intuitive for some users.

This also changes the `ginga.RGBMap` object to remove the `shift_array` and `color_array` preferences. These are now replaced by `color_map_invert`, `color_map_rot_pct`, `contrast` and `brightness`. The former were array values and the latter are scalars; this makes the channel settings file a little easier to read and parse.  This is backward incompatible from an API standpoint to the `RGBMap` object with regard to these arrays. However, it removes some problematic issues with managing color preferences.

For end users, the effect is simply that an old channel settings file will not load any saved contrast, inverted or rotated color settings. Once they save any channel settings again the new keywords will be saved and take effect on the next restart.


ejeschke:
@pllim, when you get a chance, try out this PR and checkout the differences in the channel settings (`Preferences` plugin).
Let me know what you think.

ejeschke:
@pllim, would you have a chance to check this before SciPy? I'd like to include this merge this into dev for a v5.0.0 release this fall.  Perhaps we can discuss that release at the conference.

pllim:
Sorry, it's on my queue but things keep popping up. I'll bump this up for next week.

doc/manual/plugins_local/figures/contrast-prefs.png
pllim, line 1: I like these sliders. Could never do it properly before with a sensitive mouse.

doc/manual/plugins_local/figures/cmap-prefs.png
pllim, line 1: I don't understand what "rotate" is supposed to mean here. I thought you mean image rotation (like N-up E-left).

ginga/rv/plugins/Preferences.py
```diff
@@ -60,15 +60,44 @@
...
 
-Ginga comes with a good selection of color maps, but should you want
-more, you can add custom ones or, if ``matplotlib`` is installed, you
-can load all the ones that it has.
-See "Customizing Ginga" for details.
+The "Invert CMap" checkbox can be used to invert the selected color map
+(note that a number of colormaps are also selectable from the "Colormap"
+controlin inverted form).
+
+The "Rotate" control can be used to rotate the colormap, while the
```
pllim, line 77: What does rotating a colormap mean?

ginga/rv/plugins/Preferences.py
```diff
@@ -60,15 +60,44 @@
...
 The "Intensity" control selects which intensity map should be used
 with the color map.  The intensity map is applied just before the color
 map, and can be used to change the standard linear scale of values into
 an inverted scale, logarithmic, etc.
 
-Ginga comes with a good selection of color maps, but should you want
-more, you can add custom ones or, if ``matplotlib`` is installed, you
-can load all the ones that it has.
-See "Customizing Ginga" for details.
+The "Invert CMap" checkbox can be used to invert the selected color map
```
pllim, line 73: Invert checkbox is really nice. 👍

ginga/rv/plugins/Preferences.py
```diff
@@ -60,15 +60,44 @@
...
+(note that a number of colormaps are also selectable from the "Colormap"
+controlin inverted form).
+
+The "Rotate" control can be used to rotate the colormap, while the
+"Unrotate CMap" button will restore the rotation to its default, unrotated
+state.
+
+The "Color Defaults" button will reset all the color mapping controls to
```
pllim, line 81: Personally, I don't see much use in "Color Defaults" because I can already "uninvert" and "unrotate" separately. But I guess does not hurt. 🤷

pllim (approved):
It works though I am not sure what "rotate colormap" means. Not a blocker though. Thanks!

doc/manual/plugins_local/figures/cmap-prefs.png
ejeschke: Rotate in this context means to rotate the color map.  Probably a little used feature for most people, but I've seen it in some fits viewers.

ginga/rv/plugins/Preferences.py
ejeschke: Imagine the color map wrapping around, so that if you rotate it right, you would see the colors in the color bar rotate right and wrap around to the left. And vice-versa if you rotate it left.  If you watch the color bar you can see it happen as you move the control.

ejeschke:
Rebased on top of latest release

doc/manual/plugins_local/figures/contrast-prefs.png
ejeschke: Yes, I agree.  Some people will prefer separate controls.  It also shows you in the controls what you are setting if you adjust it the old way. Plus there is now a nifty API for setting these things in 0-1 range for the viewer, where 0.5 is the default for both brightness and contrast.  This will work for the web widgets as well.

ejeschke:
Thank you, @pllim!

PR #1049, opened 2023-05-06
TST: Update pin on conda job

### LINKED ISSUE #1048: TST: conda_linux_tests failed
Example log: https://github.com/ejeschke/ginga/actions/runs/4848985893/jobs/8640572843

It appears to be using astropy 5.1. Hopefully just bumping versions on that job would fix things up but I don't have time to investigate right away.

pllim:
Fix #1048 

Not sure why conda is pulling astropy 5.1 when astropy 5.2.2 is already available on conda-forge. Maybe forcing a pin update would do it?


ejeschke (approved):
@pllim, is it possible that the python version mismatch is in any way related?  What is the minimum python version for astropy 5? A bit of googling yields "3.8".

In any case, this looks fine to me. FFTM!

pllim:
Yeah, I think we bumped the Python version.

PR #1050, opened 2023-05-06
TST: Update jurplel/install-qt-action to v3.2.0

pllim:
This hopefully will fix the nodejs warning that I see in Action log.


ejeschke (approved):
@pllim, I defer to your expertise here. Not familiar with `jurplel`.

PR #1052, opened 2023-05-25
Remove deprecated dump data types

### LINKED ISSUE #1051: Remove deprecated use of numpy.bool
The data type `np.bool` is used in a few places, for example `canvas/types/mixins.py`.

This was deprecated in NumPy 1.20 and now causes an exception in 1.24.x.

With this change Ginga can no longer be imported.

StuartLittlefair:
Fixes #1051 

NumPy v1.24.0 [removed deprecated data types](https://numpy.org/doc/stable/release/1.24.0-notes.html) `np.object`, `np.bool`, `np.float`, `np.complex`, `np.str`, and `np.int`, meaning Ginga could no longer be used with the latest NumPy versions.

This PR removes these data types from Ginga in favour of the Python data types `float`, `int`, `bool` etc. According to the raised error in NumPy:

```
`np.bool` was a deprecated alias for the builtin `bool`. To avoid this error in existing code, use `bool` by itself. Doing this will not modify any behavior and is safe. If you specifically wanted the numpy scalar type, use `np.bool_` here.
The aliases was originally deprecated in NumPy 1.20; for more details and guidance see the original release note at:
    https://numpy.org/devdocs/release/1.20.0-notes.html#deprecations
```

### DIFF
--- a/ginga/rv/plugins/TVMask.py
+++ b/ginga/rv/plugins/TVMask.py
@@ -269,7 +269,7 @@ class TVMask(LocalPlugin):
 
         try:
             # 0=False, everything else True
-            dat = fits.getdata(filename).astype(np.bool)
+            dat = fits.getdata(filename).astype('bool')
         except Exception as e:
             self.logger.error('{0}: {1}'.format(e.__class__.__name__, str(e)))
             return
@@ -318,7 +318,7 @@ class TVMask(LocalPlugin):
     def _rgbtomask(self, obj):
         """Convert RGB arrays from mask canvas object back to boolean mask."""
         dat = obj.get_image().get_data()  # RGB arrays
-        return dat.sum(axis=2).astype(np.bool)  # Convert to 2D mask
+        return dat.sum(axis=2).astype('bool')  # Convert to 2D mask
 
     def hl_table2canvas(self, w, res_dict):
         """Highlight mask on canvas when user click on table."""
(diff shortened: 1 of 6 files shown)


ejeschke (commented):
@StuartLittlefair , thank you for this.  One question before we merge. In other cases of this kind of numpy deprecation we have simply replaced `np.float` with `float`, not 'float'. It appears the same could be done here: `bool` instead of 'bool', etc.  Is there a reason for your choice?

StuartLittlefair:
None whatsoever, I'm happy with either.

ejeschke:
> None whatsoever, I'm happy with either.

Then, would you mind changing it so they are consistent? Looks like not so many changes. Then I will merge it.

pllim (approved):
LGTM but I'll give @ejeschke a chance to review. Thanks!

ginga/rv/plugins/TVMask.py
```diff
@@ -318,7 +318,7 @@ def recreate_toc(self):
     def _rgbtomask(self, obj):
         """Convert RGB arrays from mask canvas object back to boolean mask."""
         dat = obj.get_image().get_data()  # RGB arrays
-        return dat.sum(axis=2).astype(np.bool)  # Convert to 2D mask
+        return dat.sum(axis=2).astype('bool')  # Convert to 2D mask
```
pllim, line 321: I think you missed one.
```suggestion
        return dat.sum(axis=2).astype(bool)  # Convert to 2D mask
```

ejeschke:
Looks good. Thanks, @StuartLittlefair !

PR #1058, opened 2023-07-25
Fix for change in API in PIL v10.0

### LINKED ISSUE #1057: Pillow v10.0 has deprecated many text size methods, including ImageDraw.textsize
See https://pillow.readthedocs.io/en/stable/deprecations.html#font-size-and-offset-methods

This means that  the following lines need changing to use an alternative method

https://github.com/ejeschke/ginga/blob/804ca99f8bde9f3bc75220b26b3020e9367bc5f9/ginga/pilw/PilHelp.py#L39

https://github.com/ejeschke/ginga/blob/804ca99f8bde9f3bc75220b26b3020e9367bc5f9/ginga/pilw/PilHelp.py#L134

ejeschke:
- fix #1057

### DIFF
--- a/ginga/pilw/PilHelp.py
+++ b/ginga/pilw/PilHelp.py
@@ -34,13 +34,12 @@ def load_font(font_name, font_file):
 
 def text_size(text, font):
     f = get_cached_font(font.fontname, font.fontsize)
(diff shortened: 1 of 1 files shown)


ejeschke:
Seems to work fine for older versions as well.

PR #1066, opened 2023-09-08
Add a confirmation dialog when the reference viewer is closed

ejeschke:
- pops up a confirmation dialog that confirms you intend to quit
- adds support across backends
- can be overridden by adding a line to your general.cfg
- example general.cfg updated

### DIFF
--- a/ginga/gtk3w/Widgets.py
+++ b/ginga/gtk3w/Widgets.py
@@ -2375,7 +2375,7 @@ class Application(Callback.Callbacks):
         GObject.threads_init()
         # self._time_save = time.time()
 
-        for name in ['close', 'shutdown']:
+        for name in ('close', 'shutdown'):
             self.enable_callback(name)
 
         # Set up Gtk style
--- a/ginga/gtk3w/Widgets.py
+++ b/ginga/gtk3w/Widgets.py
@@ -2375,7 +2375,7 @@ class Application(Callback.Callbacks):
         GObject.threads_init()
         # self._time_save = time.time()
 
-        for name in ('shutdown', ):
+        for name in ['close', 'shutdown']:
             self.enable_callback(name)
 
         # Set up Gtk style
@@ -2435,7 +2435,20 @@ class Application(Callback.Callbacks):
     def mainloop(self):
         Gtk.main()
 
+    def close(self):
+        """Called when someone is asking the application to close.
+        Can register for this callback if you want an application-wide
+        event to confirm closure.
+        """
+        self.make_callback('close')
+
     def quit(self):
+        """Called when someone is forcibly quitting the application.
+        Can register for this callback if you want an application-wide
+        event to clean up before shutdown.
+        """
+        self.make_callback('shutdown')
+
         Gtk.main_quit()
 
 
--- a/ginga/qtw/Widgets.py
+++ b/ginga/qtw/Widgets.py
@@ -2117,7 +2117,7 @@ class Application(Callback.Callbacks):
         ydpi = screen.physicalDotsPerInchY()
         self.screen_res = max(xdpi, ydpi)
 
-        for name in ('shutdown', ):
+        for name in ['close', 'shutdown']:
             self.enable_callback(name)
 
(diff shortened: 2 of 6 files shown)


ejeschke:
@pllim, if/when you have time for a quick review. Just adds a pop-up confirmation dialog when you close the reference viewer.

Not sure why we are getting the RTD build failure, but looks like #1060 is the cause.  Did I see something go by in my email about a fix for this in `stginga`?

pllim:
For the docs, hopefully #1067 will fix it. 🤞 

Thanks for the ping. I will review when I get the chance.

ejeschke:
Rebased to pick up the RTD fix

ginga/gtk3w/Widgets.py
```diff
@@ -2375,7 +2375,7 @@ def __init__(self, logger=None, settings=None):
         GObject.threads_init()
         # self._time_save = time.time()
 
-        for name in ('shutdown', ):
+        for name in ['close', 'shutdown']:
```
pllim, line 2378: No reason to not use tuple.
```suggestion
        for name in ('close', 'shutdown'):
```

ginga/qtw/Widgets.py
```diff
@@ -2117,7 +2117,7 @@ def __init__(self, logger=None, settings=None):
         ydpi = screen.physicalDotsPerInchY()
         self.screen_res = max(xdpi, ydpi)
 
-        for name in ('shutdown', ):
+        for name in ['close', 'shutdown']:
```
pllim, line 2120:
```suggestion
        for name in ('close', 'shutdown'):
```

ginga/web/pgw/Widgets.py
```diff
@@ -3312,7 +3312,7 @@ def __init__(self, logger=None, base_url=None,
         self.caller_id = 0
         self.callers = {}
 
-        for name in ('shutdown', ):
+        for name in ['close', 'shutdown']:
```
pllim, line 3315:
```suggestion
        for name in ('close', 'shutdown'):
```

pllim (commented):
As a user, I cannot tell what is the difference between "close" and "shutdown" in Ginga. Can you please clarify?

ginga/gtk3w/Widgets.py
ejeschke: I read recently that tuples and lists are both stored as arrays (with associated Pyobject) in Python. Why do you prefer a tuple in this case? Just curious.

pllim: Tuple is immutable, so I feel safer. 😆

ejeschke:
> As a user, I cannot tell what is the difference between "close" and "shutdown" in Ginga. Can you please clarify?

It just provides for a 2-step quit sequence.  When someplace calls `fv.close()` the "close" callback is called. Ginga has a function registered for that callback to pop up a confirmation dialog box. If the user confirms, then `fv.quit()` is called which initiates the "shutdown" of the program. Any code that as registered for the "shutdown" callback will then get called to properly close any resources and terminate the program.

pllim:
Re: https://github.com/ejeschke/ginga/pull/1066#issuecomment-1712146057

Maybe I am not fully understanding but won't it be simpler to have `fv.quit(prompt_dialog=True)` instead of a separate `fv.close()`?

ejeschke:
The "close" and "shutdown" callbacks allow a flexible way to handle the closure of the program. For example, a plugin might want to register for "shutdown" in order to close a running thread they started or close a database connection, or they could register for "close" to make a checkpoint or pop up their own dialog to say that there are unsaved changes, etc.  The reference viewer is basically this core around which there are a large collection of plugins.

Plus, we use the framework for many other programs that are not the reference viewer, but benefit from this style.

pllim:
Thanks for the clarification! The diff looks reasonable to me and I am okay with this feature's addition. Feel free to merge. 😸

ejeschke:
Thanks, @pllim!

PR #1068, opened 2023-09-11
Only set Vizier row limit if imported

tbowers7:
Move the Vizier row limit setting to inside the import try..except to avoid an error if `astroquery` is not imported.


pllim (approved):
Good catch. Thanks!

ejeschke (approved):
Thanks, @tbowers7!

PR #1072, opened 2023-09-13
Fixes for numpy 2.0

### LINKED ISSUE #1054: TST: Test against numpy 2.0.dev
numpy 2.0.dev is available from their nightly wheel and it has breaking changes. Since the backend here calls a lot of numpy, better make sure they are compatible sooner than later.

ejeschke:
To make ginga compatible with `numpy` 2.0.

This will continue to be updated as more of the third party packages become available (scipy, matplotlib, etc).

- fix #1054
-


ejeschke:
There seems to be very little needed to be changed, so far.

pllim:
Actually both scipy and matplotlib dev are available from https://anaconda.org/scientific-python-nightly-wheels/repo too.

ejeschke:
Haven't yet installed `scipy` or `matplotlib`. This is tested so far with numpy 2.0 nightly and astropy nightly and pillow 10.

pllim (approved):
The diff looks reasonable though we really cannot tell for sure on the CI until it is set up. But this is a good start. Thanks!

pllim:
Great news. Thanks for checking!

ejeschke:
@pllim, this is passing all tests when I run them locally.  After testing a number of plugins pretty thoroughly, I am thinking that we could merge this.  If new issues arise with np2 we can fix them when they pop up.  What do you think?

pllim (approved):
Sounds good. Feel free to merge. Thanks!

ejeschke:
Thanks, @pllim!

PR #1073, opened 2023-09-15
TST: Use tox for CI

### LINKED ISSUE #1054: TST: Test against numpy 2.0.dev

pllim:
Use tox for CI except for the conda job.

Fix #1054

### DIFF
--- a/tox.ini
+++ b/tox.ini
@@ -4,9 +4,6 @@ envlist =
     codestyle
     twine
     bandit
-requires =
-    setuptools >= 30.3.0
-    pip >= 19.3.1
 
 [testenv]
 setenv =
--- a/tox.ini
+++ b/tox.ini
@@ -1,0 +1,93 @@
+[tox]
+envlist =
+    py{38,39,310,311}-test{,-oldestdeps,-devdeps,-predeps}
+    codestyle
+    twine
+    bandit
+requires =
+    setuptools >= 30.3.0
+    pip >= 19.3.1
+
+[testenv]
+setenv =
+    devdeps: PIP_EXTRA_INDEX_URL = https://pypi.anaconda.org/astropy/simple https://pypi.anaconda.org/scientific-python-nightly-wheels/simple
+
+# Pass through the following environemnt variables which are needed for the CI
+passenv = HOME,WINDIR,CC,CI
+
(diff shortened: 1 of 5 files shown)


pllim:
@ejeschke , this PR switches the test machinery to tox. This makes it easier to mix and match test matrix and also pulling in different upstream dev wheels (e.g., numpy-dev).

tox.ini
```diff
@@ -0,0 +1,93 @@
+[tox]
+envlist =
+    py{38,39,310,311}-test{,-oldestdeps,-devdeps,-predeps}
+    codestyle
+    twine
+    bandit
+requires =
+    setuptools >= 30.3.0
```
ejeschke, line 8: No need for `setuptools_scm` in `tox.ini`?

ejeschke (approved):
@pllim, I'm new to `tox` (but learning!) so will trust to your judgement on this one.  It sounds like it will give some added flexibility. FFTM

tox.ini
pllim: `astropy` does not list it, and it is already declared in pyproject.toml . But if you want it listed, doesn't hurt. 🤷

ejeschke: Only if an accurate version number is needed in the testing?

pllim: Maybe. I actually don't fully understand this section. I was just copy pasting from a setup I know will work. Do you want me to remove it and see?

ejeschke: I guess if tox does not bypass the `setup.cfg` then it shouldn't be necessary ...?

ejeschke: > Do you want me to remove it and see?

I think we need it in `setup.cfg` (and likely `pyproject.toml`). I just don't know how `tox` interacts with those, but presumably uses pip to install things so I suppose that is good enough.  So no need to change anything here, I guess.

pllim: I removed it. I am curious now.

pllim:
CI still passed. I guess we don't need that section after all, good catch!

ejeschke:
Thanks for investigating. FFTM

pllim:
Thanks for the review!

PR #1074, opened 2023-09-26
TST: Use Python 3.12 with devdeps

### LINKED ISSUE #1071: TST: Test against Python 3.12
Python 3.12 RC is out. Though might have to wait a little because upstream is a little slow in adapting unless you want to install a bunch of dev versions (numpy, scipy, etc). And also blocked by:

* https://github.com/astropy/astropy/pull/14784
* https://github.com/ejeschke/ginga/pull/1073

pllim:
Fix #1071

### DIFF
--- a/.github/workflows/ci_workflows.yml
+++ b/.github/workflows/ci_workflows.yml
@@ -65,8 +65,8 @@ jobs:
 
           - name: Test with dev dependencies
             os: ubuntu-latest
-            python: '3.11'
-            toxenv: py311-test-devdeps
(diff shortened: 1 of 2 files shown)


ejeschke:
It's a good sign, but probably needs an interactive test.

ejeschke (approved):
Looks good. FFTM!

pllim:
> probably needs an interactive test

For sure. Though it will be a while before it is available via conda. Unless you want to build it from source; Possible but probably not trivial.

PR #1077, opened 2023-12-02
Fix for importing mpl colormaps with recent versions of matplotlib

ejeschke:
Fixes an issue with the API in recent versions of matplotlib.

Ginga can continue to import and use matplotlib colormaps.


ginga/cmap.py
```diff
@@ -13309,8 +13309,7 @@ def add_matplotlib_cmap(cm, name=None):
 def add_matplotlib_cmaps(fail_on_import_error=True):
     """Add all matplotlib colormaps."""
     try:
-        import matplotlib.pyplot as plt
-        from matplotlib import cm as _cm
+        from matplotlib import colormaps as _mpl_cm
         if MPL_LT_3_4:
             from matplotlib.cbook import mplDeprecation as MatplotlibDeprecationWarning
         else:
```
ejeschke, line 13315: @pllim, should we drop support for mpl < 3.4 and git rid of this workaround needed for suppressing `MatplotlibDeprecationWarning` ?

ejeschke:
Here is a simple one if you have a minute or two only, @pllim.

ginga/cmap.py
pllim: Matplotlib is at 3.8 now.  Matplotlib 3.3 has not been touched since early 2021. Do you still have any users stuck with matplotlib < 3.4? If not, probably less work to drop it.

ejeschke: Sounds good. I'll remove the support from this module for that.

pllim: Should also bump this? 

https://github.com/ejeschke/ginga/blob/f0eecdf92eb1b3955482240aa6fd957aa7f807e4/setup.cfg#L60

ejeschke: good catch

ginga/cmap.py
```diff
@@ -13309,27 +13305,23 @@ def add_matplotlib_cmap(cm, name=None):
...
 
     # NOTE: Update if matplotlib has new public API for this.
-    for name in plt.colormaps():
+    for name in _mpl_cm:
         if not isinstance(name, str):
             continue
         try:
             # Do not load deprecated colormaps
             with warnings.catch_warnings():
                 warnings.simplefilter('error', MatplotlibDeprecationWarning)
```
pllim, line 13323: I think the deprecation warning catching should be removed. It should no longer happen.

ejeschke: @pllim, because those particular color maps have been removed some time ago? I guess we can add the code back if they deprecate more color maps in the future. Wondering why they deprecated color maps...

pllim: @ejeschke , I don't follow matplotlib development closely, so not sure. But I do not see any warning now using the matplotlib I have locally. Do you?

I guess does not hurt to keep it as well.

pllim (approved):
LGTM FFTM TYVM 😸

ejeschke:
TY@PLLIM!

PR #1083, opened 2024-01-01
Use svg icons

ejeschke:
- Change the PNG (bitmap) icons to SVG (vector, scalable) icons
- Improve cursor handling and make scalable cursors
- Replace some icons with (hopefully) more intuitive ones


ejeschke:
@pllim, when you get a chance can you simply review the look of the new icons and give me any feedback on them?  I think they should be a bit more intuitive than the old ones.

ginga/cursors/pencil.svg
pllim, line 1: What is this one for? Looks almost transparent.

ginga/icons/auto_cuts.svg
pllim, line 1: I don't see this one in the default layout. The A looks a bit off-center.

ginga/icons/center_image.svg
pllim, line 1: This one reminds me of Instagram but I guess it is okay if there is no better icon to use.

pllim (approved):
I am not UX/UI designer so I don't know how legit my review is. Overall I think they look good. Thanks!

If you have "dark mode" in your roadmap, you might want to see if they can be easily inverted for dark mode. Otherwise, you can ignore this paragraph.

ginga/icons/auto_cuts.svg
ejeschke: It's in the toolbar, right after reset rotation icon. Probably a lot of users use the button in the Synopsis (Info) plugin for auto cuts usually--but it is there in the toolbar.

ginga/icons/center_image.svg
ejeschke: Let me see if I can find something better.

ejeschke:
> If you have "dark mode" in your roadmap, you might want to see if they can be easily inverted for dark mode. Otherwise, you can ignore this paragraph.

Hmm, sounds like support for "themes" or "skins". Will consider for the future, but probably pretty low priority.

ginga/icons/center_image.svg
ejeschke: How about this?

center_image

It would show up much smaller, of course.

ginga/cursors/pencil.svg
ejeschke: That one is a cursor, activated in 'draw' mode, which is not used by most users of the reference viewer (activated by space, then 'b'), since they can simply right-drag to draw or the plugin will offer a toggle button for move/draw/edit, etc.

But I am thinking of using it for indicating drawing in the future, and would like to have it in the cursors.  I've tweaked it a bit to make it more visible in the normal range of contrast.

ejeschke:
Thanks, @pllim! We can further tweak the icons as needed down the road. I will update the documentation in a separate commit, since there are some other doc changes I want to do.  Several screenshots showing the toolbar need to be updated.

PR #1089, opened 2024-02-22
Unleash dependabot

pllim:
There is no need to run CI here and you won't see the benefit until after merge. This would allow Dependabot to submit PRs to upgrade things like GitHub Actions versions as needed.

See https://docs.github.com/en/code-security/getting-started/dependabot-quickstart-guide

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,26 @@
+version: 2
+updates:
+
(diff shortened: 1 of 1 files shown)


pllim:
It automatically bumps versions. I seldom see the pip one in action but it bumps Actions from time to time. Not necessarily just security fixes. Ok to merge?

ejeschke:
Let's give it a whirl.

PR #1096, opened 2024-03-18
Remove imports of distutils

### LINKED ISSUE #1095: distutils is completely deprecated in python 3.12
The `distutils` module is [to be removed in python 3.12](https://docs.python.org/3.10/whatsnew/3.10.html#distutils-deprecated). 

Ginga currently imports the module in:
- `ginga.rv.plugins.MultiDim` (using `spawn()`)
- `ginga.opengl.CanvasRenderGL` (using `LooseVersion`)

ejeschke:
- fix #1095

for rationale, [see this](https://docs.python.org/3.10/whatsnew/3.10.html#distutils-deprecated)

This will be applied also to the 5.0.1 release.

### DIFF
--- a/doc/WhatsNew.rst
+++ b/doc/WhatsNew.rst
@@ -7,6 +7,7 @@ Ver 5.1.0 (unreleased)
 - Fixed an issue where --modules option did not start a global plugin
   automatically
 - Improved error checking on contains_pts() method for canvas items
+- Removed references to distutils module, deprecated in python 3.12
 
 Ver 5.0.0 (2024-02-24)
 ======================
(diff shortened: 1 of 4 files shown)


ejeschke:
@pllim, I'm not sure whether `packaging` is in the standard library or not. I suppose I might have to add it as a dependency if it is not. Still googling...

ejeschke:
Looks like it might be not.  I'll add it to `setup.cfg`.

ejeschke:
@pllim, I'll squash merge if you have no objection.  I've tested the changes.

pllim (approved):
LGTM. I think `packaging` is pretty standard now for version check. I think astropy already pulls that in for you but does not hurt to be explicit. Thanks!

PR #1106, opened 2024-08-19
TST: Adjust Lorentz1D FWHM result for devdeps

pllim:
Adjust Lorentz1D FWHM result to be compatible with astropy v7.0 probably due to astropy/astropy#16794 bug fix upstream. Without this patch, the devdeps job fails. Example failure log: https://github.com/ejeschke/ginga/actions/runs/10448173088/job/28928224243

Would be nice if @larrybradley can confirm that such a change downstream is actually expected from the Lorentz1D `fit_deriv` fix. 🙏

### DIFF
--- a/ginga/tests/test_iqcalc.py
+++ b/ginga/tests/test_iqcalc.py
@@ -2,12 +2,15 @@ import logging
 
 import numpy as np
 import pytest
+from astropy.utils import minversion
 from numpy.testing import assert_allclose, assert_array_equal
 
 from ginga.AstroImage import AstroImage
 from ginga.util import iqcalc, iqcalc_astropy
 from ginga.util.iqcalc import have_scipy  # noqa
 from ginga.util.iqcalc_astropy import have_photutils  # noqa
+
+ASTROPY_LT_7_0 = not minversion("astropy", "7.0.dev")
 
 
 @pytest.mark.parametrize(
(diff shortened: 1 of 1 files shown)


larrybradley:
@pllim Yes, having incorrect derivatives can result in the fit not fully converging to the correct values.  One way to test this is to set `estimate_jacobian=True` when calling the fitter object.  That option will *not* use the model `fit_deriv`, and then you can compare results with and without `estimate_jacobian` (that's how the astropy model `fit_deriv` tests are performed)

ejeschke (approved):
Looks reasonable to me, @pllim.

pllim:
Thanks, all!

ginga/tests/test_iqcalc.py
```diff
@@ -2,13 +2,16 @@
...
 import pytest
+from astropy.utils import minversion
 from numpy.testing import assert_allclose, assert_array_equal
 
 from ginga.AstroImage import AstroImage
 from ginga.util import iqcalc, iqcalc_astropy
 from ginga.util.iqcalc import have_scipy  # noqa
 from ginga.util.iqcalc_astropy import have_photutils  # noqa
 
+ASTROPY_LT_7_0 = not minversion("astropy", "7.0.dev")
```
olebole, line 13: I know, this comes late, but this should check for Astropy 6.1.3, as the Astropy patch was backported and 6.1.3 was just released.

pllim: Ooops sorry for the oversight! Feel free to PR. I won't be able to get to this until Tue the earliest. It's USA holiday weekend. I apologize for any inconvenience caused.

PR #1121, opened 2025-02-21
MNT: Add Python 3.13 to test matrix, bump pillow minversion

pllim:
### Description

This pull request is to 

* Only address the Python 3.13 portion of #1112
* Fix #1119

PR #1124, opened 2025-02-22
TST: Sync pillow pin in oldestdeps

pllim:
Sync pillow pin in oldestdeps with #1121 . Sorry, should have been part of that PR.


ejeschke (approved):
FFTM

